PR #177, opened 2013-07-27
Dispatch jobs as soon as possible

fagu:
Currently, the evaluation service checks for jobs to dispatch every 2 seconds. Hence, there is an idle interval of some random length (0-2 seconds) between two jobs. This obviously doesn't have a big impact on long running jobs, but can make a difference if the jobs are all rather easy (e.g. compilation), especially when debugging (tasks or the server).

This commit pushes the `dispatch_jobs` method to the event queue for being called immediately whenever a job is added to the job queue or a job finished. This is done via a call to `add_timeout` with `seconds=0`. The event library has been adapted so that the function gets called only once if `seconds=0`.

As far as I see, `dispatch_jobs` often runs in parallel to `invalidate_submission`. Is that OK?

### DIFF
--- a/cms/io/GeventLibrary.py
+++ b/cms/io/GeventLibrary.py
@@ -229,6 +229,8 @@ class Service:
         plus (object): additional data to pass to the function.
         seconds (float): the function will be called every seconds
                          seconds.
+                         If seconds is 0, the function will be called
+                         only once.
         immediately (bool): if True, func will be called also at the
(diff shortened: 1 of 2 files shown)


stefano-maggiolo:
This became obsolete with the triggeredservice refactoring. Jobs now are dispatched as soon as they arrive in the queue - sorry for that.

PR #179, opened 2013-07-30
User groups for timing

fagu:
I wanted to propose the introduction of "user groups". This means that the `start` and `stop` time and the `per_user_time` is not specified for a contest but for groups of users. This might be useful for two reasons:

1. In the German selection camps, we sometimes had some contestants which were unable to attend them. They were admitted to take part online, often after the exam at the camp.
2. It would be nice to have test users that are able to submit before the contest starts and after it ends.

As an exercise in understanding the CMS code, I already implemented this feature. It basically seems to work but it is missing documentation and the test suite has not been adapted. Before doing this, I wanted to ask for your opinions.

One problem that arises when introducing user groups is that there are some places in the interface, where user-independent contest start and stop times are needed, for example the "Time left: ..." message in the admin interface and on the rankings page. To this end, I added a `main_group_id` column to the contest table, which specifies which group's timing to use in such cases. This is of course not optimal. For example, the submission time shown on the rankings page is incorrect; that problem already arised with USACO timing, however.

### DIFF
--- a/cms/db/__init__.py
+++ b/cms/db/__init__.py
@@ -46,7 +46,7 @@ __all__ = [
     # contest
     "Contest", "Announcement",
     # user
-    "User", "Message", "Question",
+    "User", "Message", "Question", "Group",
     # task
     "Task", "Statement", "Attachment", "SubmissionFormatElement", "Dataset",
     "Manager", "Testcase",
@@ -69,7 +69,7 @@ __all__ = [
 
 # Instantiate or import these objects.
 
-version = 5
+version = 6
 
 
 engine = create_engine(config.database, echo=config.database_debug,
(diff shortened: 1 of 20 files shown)


stefano-maggiolo:
Apologies to @fagu, this is obviously not going to pick up anytime soon :(
Closing it.

wil93:
Maybe it's possible to re-think this idea with the newly added `Team` entity in mind (e.g. adding a `Team.per_user_time` nullable field that, when != null, will override the `Contest.per_user_time` for users in that team). Just throwing this out there :smile:

stefano-maggiolo:
I thought the same, but then I thought that assuming that Team == container for other type of shared properties is probably too limiting most of the time.

PR #224, opened 2013-12-07
add registration form.

qnighy:
Someone may want to let their contestants create their account (without using external login providers.)


stefano-maggiolo:
This is way too old to be useful :(
Registration is probably still not a priority, but I'm curious, @qnighy (or, I assume @wafrelka), what are you using now? A version of this patch, login via oauth, ...?

wafrelka:
Currently we are using a modified version of this patch to hold an online contest. (https://github.com/wafrelka/cms/commit/fccdd34e0a5a2ff71ed9810f13dc5b209f3e073a)

PR #238, opened 2014-01-31
One testcase for multiple subtasks

qnighy:
GroupMin, GroupMul and GroupThreshold are not able to use a result of one testcase for multiple subtasks. If one wanted to do that, only way was to duplicate the testcase. This can become source of confusion and resource wasting.

I have attached the new scoretype NamedGroup. Using this scoretype, one testcase can appear in multiple subtasks. Sample implementation for YamlLoader is also available, but it may be unuseful.

### DIFF
--- a/cms/grading/scoretypes/NamedGroup.py
+++ b/cms/grading/scoretypes/NamedGroup.py
@@ -1,0 +1,298 @@
+#!/usr/bin/env python2
+# -*- coding: utf-8 -*-
+
+# Programming contest management system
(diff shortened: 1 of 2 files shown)


stefano-maggiolo:
Closing this, though given that we included the change to scoretypegroup allowing regexp-based subgroups, there is no reason not to have another one accepting a list of codenames as second parameters, if somebody wants (but I guess the regex-based is probably easier to use)

PR #346, opened 2014-08-14
Decoupling users and contests, allow tasks to be unassigned to any contest

wil93:
I'm creating this PR to keep track of what it's needed to separate users from contests and to allow task to be unassigned to any contest. Once all this is merged onto master, task cloning support should be drafted.

To do list:
- [x] Change loader interfaces in order to import users/tasks/contests separately.
- [x] Fix polygon importer.
- [x] Do changes needed to separate users from contests (add UserContest).
- [x] Revert changes done to separate tasks from contests (remove TaskContest).
- [x] Do changes needed to allow tasks to be unassigned to any contest.
- [x] Change the cmsAddUser script, making it use a UserLoader.
- [x] Create a cmsAddTask script, which uses a TaskLoader.
- [x] Create a cmsAddContest script, which uses a ContestLoader.
- [x] Change the cmsAddUser script, allowing to import multiple users.
- [x] Change AWS as needed to handle users/tasks *before* selecting a contest, that is:
  - Make user creation and task creation contest-independent.
  - Change the homepage (currently it forces to choose a contest or to create one), by allowing to access a list of all existing users (same for tasks).
  - Allow to tie an existing user (or existing, but unassigned yet, task) to an existing contest.
- [ ] Decide how to handle dump exporting/importing.
  - Currently, a dump is a "contest dump". A possible solution is to redefine a dump as an entire "database dump".
- [x] Fix the test suite.
- [x] Update documentation.


wil93:
I'm preparing this branch for merging ([here](https://github.com/wil93/cms/commits/decouple_user_contest_rebased) a rebased and cleaned up version can be found).

The only thing left to do is to decide how to do database dumps.

wil93:
I thought about the dump thing and I realized that it doesn't make much sense to delegate the dump operation to CMS anymore because, if we decide to define "dump" as "database dump", then there is already [a more robust way](http://www.postgresql.org/docs/9.1/static/backup-dump.html) to do it.

Regarding dump updates, we could just drop them and switch to using [alembic](http://alembic.readthedocs.org/en/latest/tutorial.html) that basically means updating the db schema directly (which makes a lot of sense, since it's a pretty common situation to need to update CMS having a non-empty database, while currently one should: dump, cmsDrop, update cms, cmsInit, and finally restore the dump).

giomasce:
Hi William.

In the past we used to update the contest description directly in the database (and, actually, to update a dump you had to reload it in a database using a suitably old version of CMS, update it and dump it again). We realized some time ago that this was cumbersome for most of the users: for most users usually CMS is deployed for one week at most; then the contest dump is retained, but everything else is thrown away and forgotten. So it is more sensible to provide ways to update the dump then the database representation. I know that your use case is the opposite one, but this is the one we are targeting.

For the same reason, we will not use alembic or anything like that. The decision of updating the dump was made consciously.

This, of course, spawns a couple of other issues.

 * We would like to have some reasonable way to support what you do with the Italian testing CMS instance. One possibility would be to support both dump and database upgrading, but it seems a bit too complicated to me. Another way that I see is to streamline a bit the dump-upgrade-restore procedure, so that it is fast enough (for instance, you do not dump and restore files, which do not change anyway), objects ids are retained and everything is done in the same (possibly EXCLUSIVE) transaction. Maybe refining this a bit one can obtain something reasonably transparent to the users.

 * We have to decide what to put in a dump. My personal idea is that a dump should not necessarily be a contest dump, but a generic dump of CMS objects (provided that dependencies are respected). So, for example, one could request a dump of two contests with all the depending objects, including of course the objects that are dependencies of both (the common users). This should not be too difficult to implement and should cover more or less all the use cases. As an alternative that can be handy in more corner cases, we could attach some sort of UUID to users, that are written in the dumps and retained just for collapsing users coming from different dumps.

I will try to have a look at your patches as soon as I can. In the meantime, consider that missing documentation update will be most probably considered a required improvement before the PR can be accepted. If you can, a couple of lines that could be included in the next release notes would be appreciated as well.

Thank you (and all the others) for the contribution, BTW.

wil93:
> to update a dump you had to reload it in a database using a suitably old version of CMS, update it and dump it again

If we delegate the export/import operation to postgres then there will be no need to have a suitably old version of CMS (let's hope that `pg_dump` is not prone to breaking changes, though).

Note that alembic creates a table named `alembic_version` with a single `version_num` field, so the dumps ("database dumps") would be stored *along with the schema version* and, if they are ever restored (to a new database, obviously) all one must do to update it is:

    $ alembic upgrade head

wil93:
Did I already mention that alembic is explicitly built for SQLAlchemy and [can automatically detect changes](http://alembic.readthedocs.org/en/latest/autogenerate.html) to the DB schema? I mean this is pretty cool isn't it:

    $ alembic revision --autogenerate -m "Added account table"

giomasce:
Autogenerating migration scripts is not a particularly useful feature: usually they are either trivial to write or so complicated that no automated process can hope to produce anything meaningful. So I see no real point in using alembic to have the migration autogeneration. On the other hand, in the (hopefully rare) cases when you have to write a migration by hand, having to write SQL makes you much more error prone than having to manipulate reasonable Python structures.

Also consider that storing a `pg_dump` output is not optimal for other reasons as well: the dump can contain a lot of cruft that have nothing to do with the contest you are exporting, that fill up disk space and potentially cause security concerns. Moreover, a `pg_dump` output contains information relevant to a specific configuration (such that PostgreSQL usernames, schema information, ID choices, ...) which have no place in a dump meant to be interoperable. For instance, you cannot import a `pg_dump` output in a database that already exists (ID would collide) and you cannot feed such a dump to some contest analysis script or importer to other systems (I am not aware of anything like that, but I like that if someone want to build something interoperable with CMS they have an easy format to target).

For all of these reasons, we (the other developers and I) think that `pg_dump` is not a good replacement for the current dump format.

wil93:
> the dump can contain a lot of cruft that have nothing to do with the contest you are exporting, that fill up disk space and potentially cause security concerns

The "cruft" part doesn't seem to be a big concern (at least on a sample contest  I used, with 7 tasks):

    $ cmsContestExporter
    $ ls -l dump_problemi.tar.gz
    -rw-r--r-- 1 wil93 wil93 3302394 23 mar 16.03 dump_problemi.tar.gz

With `pg_dump` we are wasting about 3 KB.

    $ sudo -u postgres pg_dump database --format=tar --file=/tmp/pgdump.tar
    $ sudo -u postgres gzip /tmp/pgdump.tar
    $ ls -l /tmp/pgdump.tar.gz
    -rw-r--r-- 1 postgres postgres 3305628 23 mar 16.09 pgdump.tar.gz

Also, I read from http://www.postgresql.org/docs/9.4/static/app-pgdump.html that:

> The archive file formats are designed to be portable across architectures.

I don't know exactly *how* portable it is, but the pg_dump archive format sure "sounds good".

> For instance, you cannot import a pg_dump output in a database that already exists (ID would collide) and you cannot feed such a dump to some contest analysis script or importer to other systems

Uhm, I don't know but maybe it would be better to solve that on a higher level (e.g. a system in CMS that merges two databases, assuming that they are at "the same alembic version", maybe taking care of merging the participations, scores, submissions of the same users, when they collide) instead of inventing (and maintaining) a transition format and the needed scripts (to dump, restore and update).

wil93:
Anyway:

> Autogenerating migration scripts is not a particularly useful feature: usually they are either trivial to write or so complicated that no automated process can hope to produce anything meaningful

Indeed that is not enough reason to use it, but it's neat :smile:. Note, however, that the produced script is meant to be edited by hand (because some things cannot be detected) as noted in the link I posted earlier (index are autodetected starting from alembic 0.6.1, foreign key constraints are detected starting from alembic 0.7.1).

> On the other hand, in the (hopefully rare) cases when you have to write a migration by hand, having to write SQL makes you much more error prone than having to manipulate reasonable Python structures

I'm not sure if you're referring to alembic, but note that if we use alembic we wouldn't have to use raw SQL commands, instead we would just use SQLAlchemy (as we do today).

P.S. one more thing that alembic does better: the migration scripts are named with a hash instead of a integer index (this is *awesome* when you have a pull request like this one, where we created  `updater_14.py` and then we had to rename it to `updater_15.py` because in the meantime the schema changed in the master branch).

giomasce:
> The "cruft" part doesn't seem to be a big concern (at least on a sample contest I used, with 7 tasks):

Yours is not a good example. If you have a database with more than one contest, maybe a lot of files coming from various tests and re-evaluations, maybe files from other contests that you would like to keep private, then you have a lot of unwanted cruft, which you are completely unable to manage.

> I don't know exactly how portable it is, but the pg_dump archive format sure "sounds good".

The problem is not about architectures, but about different databases and systems. One may want to feed a contest dump to NoSQL database in order to perform some analysis there. You cannot (reasonably) do that with a SQL dump.

Again, making an SQL dump of a database is a good thing if you want to have a backup. If you want to archive data or share them, than an SQL dump is not a good thing.

> Uhm, I don't know but maybe it would be better to solve that on a higher level.

Indeed, that is what we do. We provide an application-specific high level format that makes anyone able to do all the reasonable operations with data for which CMS has been designed.

While I understand that upgrading an SQL database in-place would be desirable in some cases, it simply lies outside of the CMS common practices for most of the others (with "common practices" being: setup a contest, run it, have and store a dump and perhaps feed it to a later setup of CMS, operating on another database).

giomasce:
That said, I do not think anyone would stop you from maintaining a set of scripts for performing in-place database upgrades. But we core developers are pretty decided in not considering this the preferred form for database dump and upgrading.

BTW, in the past we used to manage data migrations in a way like the one you say, and we felt a big improvement when we developed the new way.

wil93:
> One may want to feed a contest dump to NoSQL database in order to perform some analysis there. You cannot (reasonably) do that with a SQL dump.

That's one use case where the cms-dump would have an advantage over the pg-dump (even if you'd have to write some code anyway, in order to make the NoSQL db understand the cms-dump).

However (not that this is automatically a bad thing) note that this means that the dump is not really a dump anymore, because it has a structure, and we use it to query for statistics, rankings, task statements, and so on. When will we be able to run contests from dumps? :stuck_out_tongue:

What I mean is: there is already an interface to the data; if something wants to use the data then it can use SQLAlchemy like CMS does. Maybe it's just me, but I think that if I'm in a situation where I need to calculate rankings/statistics about a contest *and the database is empty* then I've made a mistake somewhere. I feel that the cms-dump system is basically a way to avoid the pain of importing an old contest in order to compute statistics for it :confused: (all this at the cost of having to write the code that computes these statistics in an ugly way, because of not using SQLAlchemy).

(Am I wrong in saying that everything one would want to query from the cms-dump would be easier, or equally difficult at most, if the dump was restored first on the db? I mean, you would have SQLAlchemy... I don't know...)

At the end of the day, for me (and probably also for the UNIX principles) the dump should just be a dump. It should not be used for elaboration (there is SQLAlchemy for that), and importing a dump to a clean db in order to elaborate its data is not *that* painful after all (and it's balanced by the advantage of having SQLAlchemy easing the elaboration phase).

> BTW, in the past we used to manage data migrations in a way like the one you say

Well, not really :stuck_out_tongue:. The one I say doesn't require to grab an old version of CMS :smile:

stefano-maggiolo:
William, the technical discussion is just one side of the problem. The other sides are not jumping around between different ways of doing things, and whether this is a way to save time in the long run or not.

I don't believe that we would save time using alembic (given this discussion, the ramping up, and the fact that difficult migrations would need to be spelled out anyway). About avoid jumping around, I'm sorry if the original discussion didn't involve you, but we consider this decision settled, and the advantages you described (that we at least partially agree on) not enough to reconsider it.

Again, we appreciate your contributions (and sorry again for the unfortunate "disappearance" of all the three of us in the past couple of months), but I would prefer if we could avoid circling around too much, and keep them constructive (like the good work Lorenzo, Luca and you did in this PR). Thanks.

wil93:
> The other sides are not jumping around between different ways of doing things, and whether this is a way to save time in the long run or not.

Yes, in fact, I'm cool with the current cms-dump system (it's already written, so why bother?).

This discussion spawned from the fact that, with this PR, CMS is becoming less contest-centric and the cms-dump system *has* to be changed in some way. My suggestion for "solving" this was to ditch the dump system and to use one that's more generic and that's already done (thus removing code, not even having to write more of it) and that likely won't require further changes in the future.

Of course I don't believe we should change the dump system just for the sake of changing it. I'm sorry if I sounded like that :smile:

giomasce:
It did not sound like that to me. But the point is that we like it as it is now. The required changes are trivial.

wil93:
I pushed some fixes needed after #402 was merged (mostly s/user/participation/). I updated the documentation (although I realized that it would be sensible to make `cmsAddTask` able to import "to a new dataset" instead of failing).

Question: is "reimporting" needed anymore? If you want to add users/tasks to an existing contest you just `cmsAdd*` them and then use AWS to "tie" them. Maybe we want to have a `--force` flag to `cmsAdd*` in order to overwrite a user (keeping submissions) but I'm not sure...

@giomasce if you want to fix the cms-dump I gave you the authorization to push on wil93/cms, in case you'd want it :stuck_out_tongue:

giomasce:
I will try to have at look at this this evening.

cms/server/AdminWebServer.py
```diff
@@ -759,17 +1087,20 @@ def post(self, task_id):
...
-    def get(self, statement_id):
+    def get(self, task_id, statement_id):
         statement = self.safe_get_item(Statement, statement_id)
-        task = statement.task
-        self.contest = task.contest
+        task = self.safe_get_item(Task, task_id)
+
+        # Additional check.
+        if task is not statement.task:
+            raise tornado.web.HTTPError(404)
```
stefano-maggiolo: This seems weird enough to require logging an error

cms/server/AdminWebServer.py
```diff
@@ -523,6 +525,330 @@ def get(self, contest_id):
...
+
+        self.contest = self.safe_get_item(Contest, contest_id)
+
+        try:
+            task_id = self.get_argument("task_id")
+            operation = self.get_argument("operation")
+            assert operation in (
+                "Remove from contest",
+                "Move up",
+                "Move down"
```
stefano-maggiolo: Would it be possible to have these as class constants? (Same for EditUserContest above.)

cms/server/AdminWebServer.py
```diff
@@ -822,17 +1152,20 @@ def post(self, task_id):
...
-    def get(self, attachment_id):
+    def get(self, task_id, attachment_id):
         attachment = self.safe_get_item(Attachment, attachment_id)
-        task = attachment.task
-        self.contest = task.contest
+        task = self.safe_get_item(Task, task_id)
+
+        # Additional check.
+        if attachment.task is not task:
+            raise tornado.web.HTTPError(404)
```
stefano-maggiolo: Same here.

cms/server/AdminWebServer.py
```diff
@@ -886,17 +1218,22 @@ def post(self, dataset_id):
...
 
     """
-    def get(self, manager_id):
+    def get(self, dataset_id, manager_id):
         manager = self.safe_get_item(Manager, manager_id)
+        dataset = self.safe_get_item(Dataset, dataset_id)
+
+        # Additional check.
+        if manager.dataset is not dataset:
+            raise tornado.web.HTTPError(404)
```
stefano-maggiolo: Same (you can also save the comments if the log messages are good).

cms/server/AdminWebServer.py
```diff
@@ -905,48 +1242,109 @@ def get(self, manager_id):
 
 
 class AddDatasetHandler(BaseHandler):
-    """Add a dataset to a task.
+    """Add a new, clean dataset to a task.
```
stefano-maggiolo: What does "clean" mean?

cms/server/AdminWebServer.py
```diff
@@ -1393,10 +1781,15 @@ class DeleteTestcaseHandler(BaseHandler):
...
 
     """
-    def get(self, testcase_id):
+    def get(self, dataset_id, testcase_id):
         testcase = self.safe_get_item(Testcase, testcase_id)
+        dataset = self.safe_get_item(Dataset, dataset_id)
+
+        # Additional check.
+        if dataset is not testcase.dataset:
+            raise tornado.web.HTTPError(404)
```
stefano-maggiolo: Same.

cms/server/AdminWebServer.py
```diff
@@ -1675,81 +2061,93 @@ def post(self, contest_id):
...
     """
-    def get(self, ann_id):
+    def get(self, contest_id, ann_id):
         ann = self.safe_get_item(Announcement, ann_id)
-        contest_id = ann.contest.id
+        self.contest = self.safe_get_item(Contest, contest_id)
+
+        # Additional check.
+        if self.contest is not ann.contest:
+            raise tornado.web.HTTPError(404)
```
stefano-maggiolo: Same.

cms/server/AdminWebServer.py
```diff
@@ -523,6 +525,330 @@ def get(self, contest_id):
...
+
+        if try_commit(self.sql_session, self):
+            # Create the user on RWS.
+            self.application.service.proxy_service.reinitialize()
+
+        # Maybe they'll want to do this again (for another task)
+        self.redirect(fallback_page)
+
+
+class TasklistHandler(BaseHandler):
```
stefano-maggiolo: If I'm not mistaken, this and the next can be a SimpleContestHandler(something).

cms/server/AdminWebServer.py
```diff
@@ -1675,81 +2061,93 @@ def post(self, contest_id):
...
+                            .first()
 
-            self.get_string(attrs, "first_name")
-            self.get_string(attrs, "last_name")
-            self.get_string(attrs, "username", empty=None)
-            self.get_string(attrs, "password")
-            self.get_string(attrs, "email")
+        # Check that the participation is valid.
+        if participation is None:
+            raise tornado.web.HTTPError(404)
```
stefano-maggiolo: Maybe you can save a few lines factoring out these in an _init method for this class.

cms/server/AdminWebServer.py
```diff
@@ -1917,20 +2314,25 @@ class QuestionIgnoreHandler(BaseHandler):
...
         ref = self.get_argument("ref", "/")
-
-        # Fetch form data.
         question = self.safe_get_item(Question, question_id)
-        self.contest = question.user.contest
+        self.contest = self.safe_get_item(Contest, contest_id)
+
+        # Additional check.
+        if self.contest is not question.participation.contest:
+            raise tornado.web.HTTPError(404)
```
stefano-maggiolo: Same.

cms/server/AdminWebServer.py
wil93: Probably.. however, is it better to do these kind of refactoring here or in the new AWS? I'm asking because I suspect that the new AWS could need a lot of this old code changed

cms/server/AdminWebServer.py
wil93: Uhm if I recall correctly that check was more like an assert. The only case I can think of that happening is when the user visits an URL that "purposely" mixes up a statement_id and a task_id that are not related (and that doesn't happen when "browsing" the admin panel), so I thought that a 404 response was the most semantically accurate thing to do.

cms/server/AdminWebServer.py
wil93: SimpleContestHandler provides contest_id, which is not needed by these

cms/server/AdminWebServer.py
wil93: Is the equivalent to the current behavior when the dataset_id_to_copy given is equal to the string: `-`. Now AddDatasetHandler adds a new (always empty) dataset, while the new CloneDatasetHandler creates a new dataset by cloning an existing one.

cms/server/AdminWebServer.py
stefano-maggiolo: As I understand the new AWS will have none of these things as it will be mainly a client side app. I'm not asking to share the constant between Python and JS, just give a name to these string in the class so you can use it both here and later.

cms/server/AdminWebServer.py
stefano-maggiolo: Might be worth to add a SimpleHandler than.

cms/server/AdminWebServer.py
stefano-maggiolo: Yes, what I meant was "please use a few more words to explain it" :)

cms/server/AdminWebServer.py
stefano-maggiolo: Ah, right, I mistakenly thought this check was for possible database corruption. Yes, the 404 is appropriate, the comment is... perfectible :) Can you change it everywhere to  something like "Protect against URLs providing incompatible parameters"?

cms/server/AdminWebServer.py
wil93: Done

wil93:
I tried updating the polygon loaders but I get this error when using the new PolygonTaskLoader:

    wil93@wil93-arch /tmp/prova
    (venv)> $ cmsAddTask .
    2015/05/08 22:51:26 - INFO [None] Using configuration file /usr/local/etc/cms.conf.
    2015/05/08 22:51:26 - INFO [None] Loading parameters for task prova.
    {u'primary_statements': u'["it"]', u'submission_format': [<cms.db.task.SubmissionFormatElement object at 0x7effc9f17350>], u'statements': [<cms.db.task.Statement object at 0x7effca3bd950>], u'name': 'prova', u'title': 'Somma'}
      File "/home/wil93/venv/bin/cmsAddTask", line 9, in <module>
        load_entry_point('cms==1.3.0rc0', 'console_scripts', 'cmsAddTask')()
      File "/home/wil93/venv/lib/python2.7/site-packages/cms-1.3.0rc0-py2.7.egg/cmscontrib/AddTask.py", line 114, in main
        loader_class=loader_class
      File "/home/wil93/venv/lib/python2.7/site-packages/cms-1.3.0rc0-py2.7.egg/cmscontrib/AddTask.py", line 69, in do_import
        task = self.loader.get_task()
      File "/home/wil93/venv/lib/python2.7/site-packages/cms-1.3.0rc0-py2.7.egg/cmscontrib/loaders/polygon.py", line 160, in get_task
        task = Task(**args)
      File "<string>", line 4, in __init__
      File "/home/wil93/venv/lib/python2.7/site-packages/sqlalchemy/orm/state.py", line 306, in _initialize_instance
        manager.dispatch.init_failure(self, args, kwargs)
      File "/home/wil93/venv/lib/python2.7/site-packages/sqlalchemy/util/langhelpers.py", line 60, in __exit__
        compat.reraise(exc_type, exc_value, exc_tb)
      File "/home/wil93/venv/lib/python2.7/site-packages/sqlalchemy/orm/state.py", line 303, in _initialize_instance
        return manager.original_init(*mixed[1:], **kwargs)
      File "/home/wil93/venv/lib/python2.7/site-packages/cms-1.3.0rc0-py2.7.egg/cms/db/base.py", line 224, in __init__
        setattr(self, prp.key, val)
      File "/home/wil93/venv/lib/python2.7/site-packages/sqlalchemy/orm/attributes.py", line 224, in __set__
        instance_dict(instance), value, None)
      File "/home/wil93/venv/lib/python2.7/site-packages/sqlalchemy/orm/attributes.py", line 1022, in set
        lambda adapter, i: adapter.adapt_like_to_iterable(i))
      File "/home/wil93/venv/lib/python2.7/site-packages/sqlalchemy/orm/attributes.py", line 1057, in _set_iterable
        collections.bulk_replace(new_values, old_collection, new_collection)
      File "/home/wil93/venv/lib/python2.7/site-packages/sqlalchemy/orm/collections.py", line 790, in bulk_replace
        new_adapter.append_with_event(member)
      File "/home/wil93/venv/lib/python2.7/site-packages/sqlalchemy/orm/collections.py", line 653, in append_with_event
        self._data()._sa_appender(item, _sa_initiator=initiator)
      File "/home/wil93/venv/lib/python2.7/site-packages/sqlalchemy/orm/collections.py", line 1529, in set
        self.__setitem__(key, value, _sa_initiator)
      File "/home/wil93/venv/lib/python2.7/site-packages/cms-1.3.0rc0-py2.7.egg/cms/db/smartmappedcollection.py", line 186, in __setitem__
        assert self._linked
    AssertionError

What could that mean? To test it, I created this sample task: https://dl.dropboxusercontent.com/u/45305908/somma.zip

PolygonUserLoader, instead, looks fine.

wil93:
My bad, I was (again) using a too new SQLAlchemy :smile:

cms/server/ContestWebServer.py
```diff
@@ -160,20 +166,47 @@ def get_current_user(self):
...
+            .first()
+
+        # Check if participaton exists.
+        if participation is None:
             self.clear_cookie("login")
             return None
-        if config.block_hidden_users and user.hidden:
+
+        # If a contest-specific password is defined, use that. If it's
+        # not, use the user's main password.
```
stefano-maggiolo: Do we need this level of detail? Why a single password is not good enough?

cms/server/ContestWebServer.py
```diff
@@ -135,10 +136,15 @@ def prepare(self):
         self.r_params = self.render_params()
 
     def get_current_user(self):
-        """Gets the current user logged in from the cookies
+        """Gets the current user and participation from cookies.
 
-        If a valid cookie is retrieved, return a User object with the
-        username specified in the cookie. Otherwise, return None.
+        If a valid cookie is retrieved, return a (User, Participation)
```
stefano-maggiolo: Isn't it easier to just return the participation and let the caller use participation.user, which is already loaded since we are in the same session?

cms/server/ContestWebServer.py
wil93: Recentyl I thought about that and my stance now is this: we should keep only `Participation.password` *and* we should move `User.username` to `Participation.username` (so a user can be authenticated as `f.lastname` in one contest and as `f.lastname2` in another contest where there is a participating user with the same first name and last name).

Furthermore, since that would mean that username/password is *only* an authentication mechanism (and a rather "old" one in the OAuth2 era...) my proposal is to ditch: `User.username`, `User.password`, `Participation.password`, and add: `Participation.credentials` as a JSON string which depends on the authentication mechanism decided by the admins of that particular contest.

cmscontrib/updaters/update_15.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def split_user(self, user_id, user_data):
+        participation_id = self.get_id()
+        participation_data = split_dict(
+            user_data,
+            "contest", "delay_time",
+            "extra_time", "ip",
+            "messages", "usertests",
+            "questions", "starting_time",
+            "submissions")
```
wil93: Todo: add "hidden"

giomasce:
This was merged at last.

However, be warned: next time you try again to submit such a huge patch without properly separating different logical changes something terribly bad will happen to you! :-P

wil93:
Hooray!  https://www.youtube.com/watch?v=QUAItQmq-LU

> next time you try again to submit such a huge patch without properly separating different logical changes something terribly bad will happen to you

:innocent:

Issue #39, opened 2012-10-11
Use the new sandbox by Martin and Bernard


giomasce:
Martin Mares and Bernard Blackham wrote a new sandbox, that uses more advanced Linux features instead of ptrace to control the running program. This results into lower overhead and less painful configuration (particularly, no need to maintain a syscall list that varies between CPU and distribution).

The interface of the new sandbox isn't much different from the old one, although some changes are required anyway both in Sandbox and in the TaskTypes. These changes should actually be mostly simplifications, but they can be somewhat invasive and probably need a few care.

CMS should move to this new sandbox.

PR #363, opened 2014-10-11
Docker support

veluca93:
This commit should add support for running cms inside a docker container.


lerks:
I eventually had time to look into this PR (I had to learn what exactly docker is just for this!).

I think we may want to include in the repo something like your dockerfile. But I didn't properly understand what its purpose is, i.e. what use case you had in mind when you wrote it. Could you tell me? (That's probably due to the fact that you didn't write any documentation explaining what this is and how it's supposed to be used: if you want this to be merged you'll have to provide it)

wil93:
I'd suggest to split docker-related files from CMS. Take [discourse](https://github.com/discourse/discourse) as an example, they created a [discourse_docker](https://github.com/discourse/discourse_docker) repository that contains what's needed to create and update the docker image.

lerks:
It depends. I don't know what discourse is nor if their approach to docker is the "recommended" one. But as far as I can see their repo has a very different style than this PR.

I'm not sure if docker is a good solution to run CMS in a production environment, but you could try to convince me of it (for this you'll need more than the one-liner that is the description of this PR; and, again, you'll have to provide documentation). In this case (docker for production) an approach like the discourse one would probably be the way to go.

What I thought when looking at your PR was to provide an simple and almost fail-proof way for people to try out and play with CMS in a clean and constrained to learn and evaluate it and see if it fits their needs. That is: if one doesn't want to (or cannot) change their system by installing dependencies, packages, giving SUID bits, etc. one can download a docker container already designed to host CMS and run it from there. In this case the dockerfile and the rest of the setup would be so simple that an approach like the one of this PR would probably be enough.

stefano-maggiolo:
+1 to Luca, I don't know Docker but I'd like to know if it would ease the deployment of the environment for a real contest (not only for trying it out). It could be a good idea to add this even if it was just the latter, but I'd like to know :)

veluca93:
The intended use-case is to easily run CMS on a single machine, without polluting the environment/having root access; the installed packages are the ones we usually need to import the tasks. I have not tried yet to use a different container/different machines for different services, but I suppose it could be done with little changes.
Docker could also be useful to have available different versions of CMS.

wil93:
The main argument in favor of Docker, I think, is that it's an effective solution to dependency hell. You no longer have to worry about which version of Python, SQLAlchemy (which, by the way, was recently updated to 1.0 and broke something), you have installed.

Note also that Docker containers are often used to easily deploy microservices (e.g. a dumb worker that offers an evaluation/compilation service via RESTfult API), though CMS is currently quite monolithic (in the sense that you have to completely install CMS on every server involved).

lerks:
If I understand it correctly, you're basically using docker to set up a lightweight virtual machine, so that the little nuisances that CMS requires for its dependencies don't affect your main system. (I believe "dependency hell" is exaggerated, but maybe you can explain me why you think of it that way.) I see roughly two types of issues with our dependencies, and I think docker isn't the optimal solution for any of them:
- __CMS requires specific versions of its dependencies, that are either too new, too old or too heterogeneous to be found all at the same time in the repository of a certain distribution.__ The proper solution to this is, first of all, to have CMS loosen its requirements, by not using features that are too recent (or by having a proper fallback), by using a switch if the API of a library changed in a non-backwards-compatible manner (that's what we did a few times with psutil), by abandoning obsolete APIs, etc. Then the version requirements should be documented properly and, finally, a system should be used to enforce them in a safe and easy way: such a system could either be to build and provide a DEB package for CMS (and similarly for other distributions) so that the package manager can automatically verify and provide dependencies, or to use virtualenv and install dependencies via PIP. Docker, in this context, seems a lazy and ineffective solution: lazy because instead of tackling the problem one just tries to hide it under the carpet (by moving it to the virtual machine), ineffective because since you'll have to install the right dependencies inside the docker container you'll be facing again exactly the problem you were trying to solve.
- __isolate requires root access to be installed (to set the SUID bit).__ This has the more general effect of not allowing CMS to be installed if one doesn't have root access. But if one comes to think of it, isolate and docker are basically doing the same job with the same tools: using namespaces and control groups to provide a constrained environment to run isolated code. So, why would using docker solve the issues we're having because of isolate? I believe for one or both of the following reasons:
  - _because docker (as before) doesn't actually solve these issues but just moves them_: it would solve the root-access requirement of isolate (by providing an environment in which isolate can act like root without actually being it) but it would then re-introduce this same set of issues for itself: docker needs a daemon to be running in background and for this daemon to be installed one needs root access! So, on a system on which neither isolate nor docker are provided one would still need root access to install at least one of them. The only point in favor of docker here is that, being more popular, it's more likely to find it already installed, especially in some cloud systems.
  - _because isolate isn't taking full advantage of the kernel features._ When isolate was developed the namespaces and control groups were still a work in progress (that is, they were gradually being merged into the Linux kernel, a few features at a time). In particular the possibility for an unprivileged user to create containers was expected but still unimplemented. Thus isolate had to require privileged access (SUID bit). [see https://github.com/cms-dev/cms/issues/39 for the full story] I have not kept up-to-date with the news but I think this feature has now been merged and docker (and/or other similar container libraries) are taking advantage of it. Thus the proper solution, here, would be to update isolate to also use it and thus drop the SUID requirement.

Sorry for the long post: I didn't have time to make it shorter. :)

stefano-maggiolo:
Not saying we shouldn't fix what Luca pointed out, providing this script would be a huge time saver for people already comfortable with docker and that want to try CMS out (I guess especially in the cloud). I vote to include this provided we have some documentation.

wil93:
> I believe "dependency hell" is exaggerated

That's [what it's called](https://en.wikipedia.org/wiki/Dependency_hell) :smile:

> The proper solution to this is, first of all, to have CMS loosen its requirements

It's not very effective IMHO because you are opening to the possibility of "sneaky" bugs that, say, happen only under special circumstances and with a special configuration of dependencies.

> or to use virtualenv and install dependencies via PIP

Yes, if you don't use Docker, then virtualenv is *the* solution (in fact I already began using it myself, and it saved me some insanity over "too new" versions of packages, e.g. werkzeug and sqlalchemy just recently).

> Docker, in this context, seems a lazy and ineffective solution: lazy because instead of tackling the problem one just tries to hide it under the carpet (by moving it to the virtual machine), ineffective because since you'll have to install the right dependencies inside the docker container you'll be facing again exactly the problem you were trying to solve.

I think that "lazy" can be "good" sometimes :stuck_out_tongue:. For the effectiveness part I'm not sure I'm following :confused: the Dockerfile instructs docker to install the exact same things in every case (it doesn't matter if you run Arch or Ubuntu or whatever) so the dependency hell is effectively solved, isn't it?

stefano-maggiolo:
Closing this, as it as no activity, and the commit-history has become a mess. I guess sooner or later we'd like to have something like this but sadly nobody seems to be convinced of the right way to do it.

wil93:
We kinda continued this, but I now believe that it's better to keep the docker stuff in a different repository (see https://github.com/algorithm-ninja/cms-docker for what we use)

PR #371, opened 2014-11-22
Fix untrusted input

giomasce:
During the execution of a task of Batch type, the contestant program could tamper with the input file (usually input.txt). We check it out again from the database to be sure.

### DIFF
--- a/cms/grading/tasktypes/Batch.py
+++ b/cms/grading/tasktypes/Batch.py
@@ -319,6 +319,15 @@ class Batch(TaskType):
                                 manager_filename,
                                 job.managers[manager_filename].digest,
                                 executable=True)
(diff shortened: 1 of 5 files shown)


stefano-maggiolo:
Merged:
a6cbc9aea33f1d2072723683ddcf22218a9acdf5
1fff330a78542d83bba3b15d41d0a47f8bea3454
71db0832ed475bf09cfdd92c25960661c1a6249e
e7d3a38a304a679f3113de3d4bfa572f6bc0808e

Rewriting the input wasn't strictly needed as now the contestant does not have permission to write it. Nonetheless, this cover the case of the grader deleting the input, so it is an improvement anyway.

PR #377, opened 2014-12-24
Fix PriorityQueue items ordering when timestamp is not enough

artikz:
If multiple items are being added to the queue at once, their timestamp are too close to be distinguished. In this case PriorityQueue becomes stack for these items (at least it became in my case).

This commit tries to fix this behaviour by assigning an index to each item when adding it to the queue. So the items with equal priorities and timestamps will be extracted from the queue in the same order as they were added.

### DIFF
--- a/cms/io/priorityqueue.py
+++ b/cms/io/priorityqueue.py
@@ -67,12 +67,13 @@ class QueueEntry(object):
 
     """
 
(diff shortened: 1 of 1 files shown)


stefano-maggiolo:
Merged: 3bb092c26c3f8b4827c0ab9ff5b40c55e821e8ca

Issue #343, opened 2014-08-07
Partial results during evaluation


artikz:
Now contestant will see the submission evaluation results only after all testcases are evaluated. But many times he/she is interested to receive at least some results as early as possible. So displaying evaluation results of already evaluated testcases during the evaluation would be another great improvement.

PR #379, opened 2014-12-26
Provide partial evaluation results to admins and contestants

artikz:
Implementation of https://github.com/cms-dev/cms/issues/343.

What's done:
- submission is being scored after each evaluation job completed;
- AWS always shows current evaluation results, so admins can see the progress of evaluation;
- CWS will show current evaluation results if an option "Show partial results" is set in task settings.

The only possible problem I see here is that a submission can now be "scored" before it was "evaluated". I hope I identified all the cases where "scored" actually means "evaluated" and "scored", but it's possible that I've missed something.

### DIFF
--- a/cms/db/__init__.py
+++ b/cms/db/__init__.py
@@ -79,7 +79,7 @@ __all__ = [
 
 # Instantiate or import these objects.
 
-version = 13
+version = 14
 
 
 engine = create_engine(config.database, echo=config.database_debug,
--- a/cms/db/submission.py
+++ b/cms/db/submission.py
@@ -460,10 +460,11 @@ class SubmissionResult(Base):
         return (bool): True if scored, False otherwise.
 
         """
-        return all(getattr(self, k) is not None for k in [
-            "score", "score_details",
-            "public_score", "public_score_details",
-            "ranking_score_details"])
+        return (self.compilation_failed() or self.evaluated()) and \
+            all(getattr(self, k) is not None for k in [
+                "score", "score_details",
+                "public_score", "public_score_details",
+                "ranking_score_details"])
 
     @staticmethod
     def filter_scored():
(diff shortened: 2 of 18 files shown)


artikz:
Second approach seems nice to me, thanks. I've implemented it in the last commit.

stefano-maggiolo:
I'm merging this (need some rebasing after the score mode commit).

stefano-maggiolo:
Uhm, looking at it, it still doesn't convince me. For start, there are at least one race condition and another synchronization error. Try running the test with batchstdio.py (correct), and reevaluating that submission only. Often you'll get an error in the logs because in AWS you ask for the details but you don't have. More importantly, you'll often get 50/100 because the invalidate_score() in ES at the last testcase is overwritten by the already running thread in SS launched by the first testcase. The new SS thread will not write anything because it sees the submission as scored.

I'll try again with a proposal (if you don't have time, I can do it):
- introduce a new field `score_is_full` (or some better name);
- assume `compilation_outcome != None` <=> all scoring-related fields are present - This needs to be done in ES, on the same alchemy commit (maybe have some method `init_empty_score()` in the submission result);
- flip `score_is_full` in SS, when scoring an `evaluated()` submission result;
- show a spinner or some other indicator next to a score which is not full;
- maybe allow user to see details of a not fully scored submission?

stefano-maggiolo:
Btw, a merged view of the PR can be seen at https://github.com/stefano-maggiolo/cms/commits/master

artikz:
I'm not sure whether I will have time to look at this until Sunday. So, if you are free and wish to fix it, then, please. If not, I'll try to do something on Sunday or Monday.

stefano-maggiolo:
Ok, I'm going to work on this.

stefano-maggiolo:
Update: this is more complicated than expected to do in the current status of division between ES and SS. We might want to unify the two, and do some other minor changes to the way we manage submission results from DB to UI, before implementing this in a proper way.

stefano-maggiolo:
Lots of assumptions that turned out not to be true in this PR (mostly by me :) ). With parallelization of testcases this lost a bit of motivation, but we might give another shot at it in the future. Definitely the code here is not very usable.

PR #380, opened 2014-12-27
Use source and object files from managers file list when compiling solution

artikz:
Now only header files are being copied to the compilation directory. Sometimes it's required to copy source and object files also. For example, when creating a cross-language grader for C++ and Pascal, someone may compile the C++ version to an object file and link it to the Pascal grader to avoid code duplication.

### DIFF
--- a/cms/__init__.py
+++ b/cms/__init__.py
@@ -38,6 +38,7 @@ __all__ = [
     "LANGUAGE_NAMES", "LANGUAGES", "DEFAULT_LANGUAGES",
     "SOURCE_EXT_TO_LANGUAGE_MAP",
     "LANGUAGE_TO_SOURCE_EXT_MAP", "LANGUAGE_TO_HEADER_EXT_MAP",
+    "LANGUAGE_TO_OBJ_EXT_MAP",
     # log
     # Nothing intended for external use, no need to advertise anything.
     # util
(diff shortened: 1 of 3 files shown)


stefano-maggiolo:
I understand the .o usecase, not much the one for including source files. Note also that you are inserting the same file twice: for example in Batch, grader.%l will be inserted in the dictionary twice, and in Communication you will have the same file in stub.%l and in <taskname>.%l.

Also, don't you need to include the .o in the compilation command as well?

artikz:
Actually, I found this usecase when tried to import some contest archive. Authors used an interesting way to create such shared library. They created a graderlib.c grader library, which was included as it is (#include "graderlib.c", note, this is a .c-file, not .h) in C-grader and compiled version of which was used in Pascal grader ({$L graderlib.o}). So they only need this files to be found in compilation directory and don't need additional options in compilation commands.

stefano-maggiolo:
Interesting :)
Given my new knowledge that it is difficult to mess with other files from the C preprocessor, I'd say we can submit this patch. Waiting for a better way to deal with managers [@lerks :)  ]

stefano-maggiolo:
Merged: 4ceb6446d121ef3febecdbad1710fe6c5f30b760

PR #383, opened 2015-01-02
Fix first sweeper run

artikz:
If we have some missed operations when started ES or SS, these operations are not executed. You can check this by shutting down SS, re-evaluating a submission and starting SS after evaluation is complete. The submission will not be scored. Similar thing happens with ES.

This commit tries to fix this by starting the sweeper only after at least one executor was added.

### DIFF
--- a/cms/io/triggeredservice.py
+++ b/cms/io/triggeredservice.py
@@ -212,15 +212,13 @@ class TriggeredService(Service):
 
         self._executors = []
 
-        # Set up and spawn the sweeper.
-        #
-        # TODO: link to greenlet and react to its death.
         self._sweeper_start = None
         self._sweeper_event = Event()
-        gevent.spawn(self._sweeper_loop)
+        # We can't spawn the sweeper here, because there are no executors yet.
+        self._sweeper_spawned = False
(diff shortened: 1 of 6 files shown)


stefano-maggiolo:
Nice catch! It is kind of far fetched, but this still can have problems for services with many executors (only the first might see the change).

I propose a different solution: remove the _sweeper_timeout() method, and let subclasses call a _start_sweeper(timeout) method when they set up all the executors. What do you think?

artikz:
I like your approach too, though we will loose some magic and subclasses will require to not forget to start sweeper.

stefano-maggiolo:
But now they can forget about overriding _sweeper_timeout.

Also, I think the new `_start_sweeper` method needs to have some defense against calling it several times (or better, just do the right thing).

artikz:
Ok, I agree

artikz:
Re-implemented with start_sweeper

stefano-maggiolo:
Merged (just added a warning in case you try to start twice): d2a80c4f8581f1d4331ce7558259a4e742e989aa

PR #389, opened 2015-01-12
Update Japanese translation.

qnighy:
This makes the Japanese translation up-to-date with the current messages.pot file.


stefano-maggiolo:
Thank you Masaki! Merged: 0b7bf1ffd301a6a024d834635f4cfb5fc99e39d3

PR #393, opened 2015-01-15
Update Lithuanian translation

Motiejus:
cc @vytisb


stefano-maggiolo:
Thank you Motiejus! 18a8db9b71f432f3c17b2717ddd55fbcdff08d18

PR #394, opened 2015-01-17
Make French translation compile correctly.

qnighy:
There is a newline at an incorrect position. This commit fixes the syntax error.


stefano-maggiolo:
Thanks, I forgot to try building them... :)

dbde2efaa41f83829b65f374175fddb8ba17f872

PR #395, opened 2015-01-17
Fix broken layout in submission list page.

qnighy:
Submission lists are broken when all testcases in a task are marked public.

snapcrab_noname_2015-1-17_16-58-24_no-00

This commit fixes the problem.

snapcrab_noname_2015-1-17_17-15-12_no-00


stefano-maggiolo:
This was already merged as well :) 3fe3f9562f088b947b176c08500f4dc20c99c2e3

PR #396, opened 2015-01-17
Make sure language cookie is set globally.

qnighy:
Setting language in submission list pages etc. makes the cookie local in the path of the page.

Usually this behavior is not desired. This commit forces the cookie to be set globally.

### DIFF
--- a/cms/server/static/cws_utils.js
+++ b/cms/server/static/cws_utils.js
@@ -217,23 +217,8 @@ CMS.CWSUtils.prototype.update_time = function() {
     }
 };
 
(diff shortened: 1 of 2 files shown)


qnighy:
Simply setting `";path=" + this.url_path` doesn't work, but I've found that `";path=" + this.url_path + "/"` works correctly.
Committed qnighy/cms@b250c0fc28f2c3bc6a2d769cd21f771f81d97242.

(When this is merged, these can be squashed into one commit.)

stefano-maggiolo:
Forgot to close this, it was merged already: a9955d9a66e007f71634bd19d1d7f5211c21fb08

PR #402, opened 2015-01-23
Add submissions page for viewing all submissions in a contest

wafrelka:
Currently, an admin can view all submissions of a user or a task separately in the admin page.
With this commit, an admin can view all submissions in a contest at once.

### DIFF
--- a/cms/server/AdminWebServer.py
+++ b/cms/server/AdminWebServer.py
@@ -1635,6 +1635,26 @@ class RankingHandler(BaseHandler):
             self.render("ranking.html", **self.r_params)
 
 
+class ContestSubmissionsHandler(BaseHandler):
+    """Shows all submissions for this contest.
+
+    """
+    def get(self, contest_id):
+        contest = self.safe_get_item(Contest, contest_id)
+        self.contest = contest
+
+        self.r_params = self.render_params()
+        self.r_params["submissions"] = \
+            self.sql_session.query(Submission).join(Task)\
+                            .filter(Task.contest == contest)\
+                            .options(joinedload(Submission.user))\
+                            .options(joinedload(Submission.files))\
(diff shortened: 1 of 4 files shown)


stefano-maggiolo:
Hello! My apologies for the long delay.

You are completely right that this page was missing, thanks for doing it. I have a couple of request:
- it is missing the usual "reevaluate all" buttons - not mandatory, but since we have them in all other page it would be nice;
- on big contests, this page will create problems to the browser because of the many elements to show; can you add a todo to implement some kind of pagination?

Thank you!

wafrelka:
Updated the commit.

pagination:
- I implemented a simple pagination by reusing and modifying the one in questions.html. It is useful for a mid-size contest.
- I think it is still not enough for a large contest and added TODO comment in contest_submissions.html.

"reevaluate all" buttons:
- Currently, we cannot simply use reevaluation_buttons.html for adding "reevaluate all" buttons. It is unexpected to use the html for implementing them. We should modify reevaluation_buttons.html and related templates.
- I copied, modified and pasted the buttons to contest_submissions.html instead of using reevaluation_buttons.html. I want to refactor it in the future but not in this PR.

stefano-maggiolo:
Thank you! You are right that the pagination should be done server side to avoid huge queries, but this at least would prevent the browser from dying :)

I only changed a small nit in the javascript and restored the score button to the old form (I recently changed and revert it).

Included in 87e5765e61fdd253498ce44283f2625630d4a780 and be2adc94b0a4c1a2a495542437f4cefc7bf283e4.

stefano-maggiolo:
PS: I think you are not in https://github.com/cms-dev/cms/blob/master/AUTHORS.txt ? If you want to be, send me a PR :)

PR #403, opened 2015-01-25
Fix misdetection of score_mode in rankings.

qnighy:
cms-dev/cms@573b9e7c3e85d178c442101a78901962fcc1b3e1 changes names in cms-dev/cms@a169a580634978f1c06bf8662b571aad21be4e54, but names in cmsranking are left unchanged, and  this leads to misunderstanding of score modes in RankingWebServer.

This commit fixes the problem.


stefano-maggiolo:
Merged, thanks!
ffc2a893423f3d1ecde9299d8e63644002aac4d6

PR #408, opened 2015-02-14
Added Korean Translations

hyun-min-choi:
Korean translations have been added. 'ko.po' and 'ko.mo' should be in the cms/server/po directory.


stefano-maggiolo:
Thank you, included it as 4ec469c810a097e114c8e0b3d77c095355910eae

PR #409, opened 2015-02-22
Ranking Server: Contest header abbreviated

rspencer01:
The contest name headers in the ranking webpage are now abbreviated (with `<abbr>`) which endows them with tooltips (like the problem names) if they are too long.


lerks:
Pushed as c6c865c500f80368d381189a7ceeca58ba0ee7e7. Thanks for the contribution!

PR #411, opened 2015-03-08
Fix name of ParameterType form entry.

qnighy:
HTML Templates of some ParameterTypes have whitespaces at wrong positions, and it leads to MissingArgumentError when the form is submitted. This commit fixes the problem.


stefano-maggiolo:
Thank you Masaki, we weren't using those in the main line, so never saw the problem. Included with 1fafeb286ba2d90c6817fd3da8d6b9db5c3a1231.

Sorry also for the long delay, it was a busy time for all of us.

PR #418, opened 2015-04-19
Fix a mistake in the message shown after a question is submitted.

yihangho:
At the very least, there is an extra "the".


giomasce:
The correction is reasonable, but we have to change the original string as well. I have no time to do that now.

yihangho:
Where is the original string?

yihangho:
Oh wait, I think I get you now. Is it in files like [this](https://github.com/cms-dev/cms/blob/d292414429e1081630ed918479fa0fc3499ca15d/cms/server/po/lt.po#L115-L116)?

wil93:
I think it must be changed inside all files:

    [wil93@wil93-arch cms]$ grep -Rn  "Your question has been received, you will be notified when the it will be"
    cms/server/po/zh.po:107:"Your question has been received, you will be notified when the it will be "
    cms/server/po/ru.po:117:"Your question has been received, you will be notified when the it will be "
    cms/server/po/nl.po:109:"Your question has been received, you will be notified when the it will be "
    cms/server/po/messages.pot:109:"Your question has been received, you will be notified when the it will be "
    cms/server/po/lt.po:112:"Your question has been received, you will be notified when the it will be "
    cms/server/po/ja.po:111:"Your question has been received, you will be notified when the it will be "
    cms/server/po/it.po:114:"Your question has been received, you will be notified when the it will be "
    cms/server/po/fr.po:110:"Your question has been received, you will be notified when the it will be "
    cms/server/po/bs.po:104:msgid "Your question has been received, you will be notified when the it will be answered."

BTW, @giomasce what do you think of Transifex? :stuck_out_tongue:

yihangho:
OK, no worries, I'll go ahead and change all of those.

lerks:
The changes you added are all needed but what @giomasce means is that you also have to change it here: https://github.com/cms-dev/cms/blob/master/cms/server/ContestWebServer.py#L868

yihangho:
@lerks @giomasce Done!

giomasce:
Seems to be ok, merged.

PR #419, opened 2015-04-20
Conform requirements.txt filename to standard

wil93:
The requirements.txt filename shouldn't be uppercased. See [here](https://pip.readthedocs.org/en/1.1/requirements.html) and [here](https://github.com/travis-ci/travis-build/blob/a1a111723572f2a302cbf9e36dff515a664de338/lib/travis/build/script/python.rb#L35).


lerks:
Nothing to oppose. Merged as 2d8a97221b3795d6c3fb1402a4fc459f03b10f59. Thanks.

PR #420, opened 2015-05-22
Fix missing template file

wil93:
This was already fixed some time ago in #346, but I think it's better to include this as quickly as possible (because it easily triggers some 500 error codes in AWS).


stefano-maggiolo:
Thank you William, merged.

PR #423, opened 2015-05-25
Use less sudo during installation

wil93:
This commit aims to reduce the use of root access for the CMS installation by splitting all the code that required root away from `setup.py` script and into a new `prerequisites.py` script. This is good for:

- virtualenv users (it's generally good if no files inside a virtualenv are owned by root)
- distribution-specific packages, e.g. a package mantainer may want to add users/groups in "the right way" instead of running `setup.py` (at the moment he/she must patch setup.py)


docs/Installation.rst
```diff
@@ -191,12 +191,29 @@ If you prefer using Python Package Index, you can retrieve all Python dependenci
...
+
+Or, if you prefer to use pip:
+
+.. sourcecode:: bash
+
+    pip install --user .
+
+.. note::
+
+    If you are going to use a `virtual environment <https://virtualenv.pypa.io/en/latest/>`_ then you won't need the ``--user`` flag.
```
stefano-maggiolo: Would you care to expand a little bit more how to install with virtualenv?

docs/Installation.rst
```diff
@@ -191,12 +191,29 @@ If you prefer using Python Package Index, you can retrieve all Python dependenci
...
+
+Both commands will install CMS in your home folder. If you really want to install it globally then you should remove ``--user`` and run the install command as root (but in general that could be a bad idea, because you would be installing files in a system directory bypassing the system package manager).
```
stefano-maggiolo: Maybe something like "depending on your distribution that might be a bad idea, ..." instead of "in general...".

cmstestsuite/TestCleanCheckout.py
```diff
@@ -77,7 +77,8 @@ def setup_cms():
     os.environ["PYTHONPATH"] = "%(TEST_DIR)s" % CONFIG
 
     info("Building cms.")
-    sh("./setup.py build")
+    sh("./prerequisites.py compile_l10n")
+    sh("./prerequisites.py compile_isolate")
```
stefano-maggiolo: Not sure why you need to kill "setup build" for compilation of these things

docs/Installation.rst
wil93: Yeah I meant it as "it's not always a bad idea, because some distribution apply some precautions, but *in general* it can be a bad idea", maybe it can be reworded to be more like this...

cmstestsuite/TestCleanCheckout.py
wil93: Are you suggesting to keep both "setup build" and "prerequisite compile"? I removed "setup build" because it didn't seem to be necessary. With setuptools (at least on version 15 and up) the `build` command seems to be unnecessary.

docs/Installation.rst
wil93: I'll create a separate section then (like the ones for Ubuntu, Archlinux, Debian). But maybe that logically fits better in a separate PR...

cmstestsuite/TestCleanCheckout.py
stefano-maggiolo: I meant we can keep the setup build for these things, instead of adding this new script. Is prerequisite a standard? I feel like setup build is much more common.

wil93: I don't know if it's worth to pollute the setuptools script with the compilation of isolate and l10n files, because:

1. After building, isolate and l10n files will then get installed by prerequisites.py (because they are based on a different path than the one used by the cms python package).
2. Isolate could eventually be separately packaged and installed.
3. Having server-side l10n is not generally a "future-proof" idea either, since with RESTful APIs it's generally better to do client-side localization (see [here](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Intl)), so there is the possibility of switching to JS localization files (which are easily installed in the cms python package without the need of compiling them).

wil93: If desired, one could even add a custom build command (so that the pollution is mitigated), like [this](https://gist.github.com/wil93/e8a4aad9205ab9e772c0). However, I still feel that it's better to separate it (e.g. in this way you would need to run 3 commands in sequence, instead of 2).

prerequisites.py
```diff
@@ -0,0 +1,411 @@
...
+def ask(message):
+    """Ask the user and return True if and only if one of the following holds:
+    - the users responds "Y" or "y"
+    - the "-y" flag was set as a CLI argument
+
+    """
+    return "-y" in sys.argv or raw_input(message) in ["Y", "y"]
+
+
+def check_root(yes):
```
stefano-maggiolo: Please split this into two functions, assert_root() and assert_non_root()

prerequisites.py
```diff
@@ -0,0 +1,411 @@
...
+        print("  %s" % country_code)
+        path = os.path.join("cms", "server", "mo", country_code, "LC_MESSAGES")
+        compiled_path = os.path.join(path, "cms.mo")
+        if not os.path.exists(compiled_path):
+            print("[Error] %s not found" % (compiled_path))
+            print("[Error] You must run the compile_l10n command")
+            exit(1)
+        elif os.path.getmtime(locale) > os.path.getmtime(compiled_path):
+            print("[Warning] %s is newer than %s" % (locale, compiled_path))
+            print("[Warning] Are you sure you ran the compile_l10n command?")
```
stefano-maggiolo: here and before, spell the command out completely (prerequisite ...)

prerequisites.py
```diff
@@ -0,0 +1,411 @@
...
+    check_root(True)
+
+    print("===== Creating user and group cmsuser")
+    os.system("useradd cmsuser -c 'CMS default user' -M -r -s /bin/false -U")
+    cmsuser = pwd.getpwnam("cmsuser")
+    root = pwd.getpwnam("root")
+    cmsuser_grp = grp.getgrnam("cmsuser")
+
+    # Run compile_l10n() and compile_isolate() as *not* root
+    if os.system("sudo -u %s %s compile_l10n" % (os.getenv("SUDO_USER"), sys.argv[0])):
```
stefano-maggiolo: What is SUDO_USER?

stefano-maggiolo: IIUC, SUDO_USER is set when you run something as sudo. The fact that we are root doesn't mean we are running through sudo, though, so you might want to check that this is set before using it.

prerequisites.py
```diff
@@ -0,0 +1,411 @@
...
+- compile_l10n
+- compile_isolate
+- install_l10n  (requires root)
+- install_isolate  (requires root)
+- install  (requires root)
+- uninstall  (requires root)
+""" % (sys.argv[0])
+
+
+if __name__ == "__main__":
```
stefano-maggiolo: Please use argparse, e.g.: http://chase-seibert.github.io/blog/2014/03/21/python-multilevel-argparse.html (but with an explicit dictionary command_name -> function instead of the crazy "let's just execute whatever function the user asks for").

setup.py
```diff
@@ -30,316 +31,110 @@
...
+        os.path.join("tasks", "batch_fileio_managed", "code", "*"),
+        os.path.join("tasks", "batch_fileio_managed", "data", "*.*"),
+        os.path.join("tasks", "communication", "code", "*"),
+        os.path.join("tasks", "communication", "data", "*.*"),
+    ],
+}
+
+# Apparently, pip installs package_data with the permissions they
+# have on the source. We fix the source permissions here. (Though,
+# pip works on a copy, so no changes happen.)
```
stefano-maggiolo: Why our permissions are broken in the first place? Do we have data files != 0644?

prerequisites.py
```diff
@@ -0,0 +1,411 @@
...
+        print("[Warning] Group cmsuser not found")
+
+    print("===== Done")
+
+
+USAGE = """%s <command>
+
+Available commands:
+- compile_l10n
+- compile_isolate
```
stefano-maggiolo: Can we call these build_* and add a build that does all of them in one pass?

cmstestsuite/TestCleanCheckout.py
stefano-maggiolo: Yes, that's exactly what I meant. Anyway, I understand that your idea is to use setup.py only for Python package-related stuff, which is fine. To me setup was a better name than prerequisite but it's not too important.

Why do you say you'd have 3 commands instead of 2?

docs/Installation.rst
```diff
@@ -178,25 +121,142 @@ The following command should now mount the cgroup filesystem:
...
+   :language: python
+
+These are all the python dependencies required to develop CMS:
+
+.. literalinclude:: ../dev-requirements.txt
+   :language: python
+
+There are good reasons to install Python dependencies via pip (Python Package Index) instead of your package manager, for example: two different Linux distributions may "offer" two different versions of ``python-sqlalchemy`` while, when using pip, you can choose to install a version that is known to be working correctly with CMS.
+
+The quick and dirty way of installing Python dependencies is this:
```
stefano-maggiolo: Why "dirty"?

docs/Installation.rst
```diff
@@ -178,25 +121,142 @@ The following command should now mount the cgroup filesystem:
...
+
+.. sourcecode:: bash
+
+    pip install --user -r requirements.txt
+
+This command installs python dependencies in your home folder. If you really want to install them globally then you should remove ``--user`` and run the install command as root (but that could be a bad idea depending on your Linux distribution, because you would be installing files in a system directory bypassing the system package manager).
```
stefano-maggiolo: Nit: "(but, depending on your distribution, this might be a bad idea as it might interfere with the system package manager)"

docs/Installation.rst
```diff
@@ -178,25 +121,142 @@ The following command should now mount the cgroup filesystem:
...
+    .. sourcecode:: bash
+
+        deactivate
+
+Ubuntu
+------
+
+.. warning::
+
+  It's usually possible to install python dependencies using your Linux distribution's package manager. However, keep in mind that the version of each package is controlled by the package mantainers and could be too new or too old for CMS. This is generally not the case on Ubuntu (since the package versions are usually stable and quite old).
```
stefano-maggiolo: Can you do a search and replace for 's and 't to use the non-contracted form?

stefano-maggiolo: This is not generally the case on Ubuntu because we try to build on the packages available at a certain LTS version (in this case is 14.04), not because the packages are old. Please rephrase.

docs/Installation.rst
```diff
@@ -178,25 +121,142 @@ The following command should now mount the cgroup filesystem:
...
+
+    # Optional.
+    # sudo pacman -S python2-yaml python-sphinx python2-pycups
+    # Optionally install the following from AUR.
+    # https://aur.archlinux.org/packages/python2-pypdf2/
 
 Installing CMS
 ==============
 
+In order to run CMS there are some preparation steps to run (like installing the sandbox, compiling localization files, creating the ``cmsuser``, and so on). You can either do all these steps by hand or you can run the following command:
```
stefano-maggiolo: This should go after the "You can download CMS..."

docs/Installation.rst
```diff
@@ -178,25 +121,142 @@ The following command should now mount the cgroup filesystem:
...
+
+.. sourcecode:: bash
+
+    sudo ./prerequisites.py install
```
stefano-maggiolo: FYI: the localization files are picked up also locally, no need to install them. But I think it is better to install everything anyway.

cmstestsuite/TestCleanCheckout.py
wil93: > Why do you say you'd have 3 commands instead of 2?

Because you would need to do:

    ./setup.py <command_to_compile_isolate_and_l10n>
    ./prerequisites.py install
    ./setup.py install

docs/Installation.rst
wil93: Mainly because I wrote that phrase when the code below was still `sudo pip install ...` which then I changed :smile:

However, installing software in your home directory could be seen as "dirty" (if you do that for more than just CMS you could have clashing dependency versions). The "clean" way would be to use virtualenv.

setup.py
wil93: Dunno... I think we can get rid of these lines... What do you think? ([this is the commit that introduced them](https://github.com/cms-dev/cms/commit/d3b443acaf8dcdffc5df72cd19c8252cb09a9cd9#diff-2793981338ea75e5869780501fc50134R66))

docs/Installation.rst
wil93: > Can you do a search and replace for 's and 't to use the non-contracted form?

I removed `'t` for `won't` but left it for `don't` and `didn't`, should it be changed everywhere? (also why no contracted forms?)

docs/Installation.rst
wil93: To be fair, this is more of a pip problem. For example, npm allows to keep multiple versions of the same dependency in your home directory (so, in theory, you don't even need a virtual environment).

lerks:
I'm skeptical about one point. I guess we're not the only Python package that does localization, and that does it with .po and .mo files. All these other packages also have to compile .po files into .mo files, like we have to. Among them there will be some that use setuptools to build and install. How do they handle the compilation of .po files? Do they have a prerequisite files like the one you want to introduce? If they don't, why do you want to diverge from the standard practice?

setup.py
stefano-maggiolo: I didn't realize they were already there, let's keep it for the moment, and see if they are needed later.

stefano-maggiolo:
Maybe we should investigate this: http://askubuntu.com/questions/144122/when-should-mo-files-be-generated ?

The point about isolate remains, though, unless we want to move its compilation and installation entirely to the other repo.

docs/Installation.rst
stefano-maggiolo: I'd prefer so. It's more professional :P

setup.py
stefano-maggiolo: I did not get that it was already there. Let's keep it for the moment and see if we can get rid of it later.

wil93:
I don't know much about this type of compiled l10n (I did it client-side, in a web application, so I didn't need to compile anything).

> Maybe we should investigate this: http://askubuntu.com/questions/144122/when-should-mo-files-be-generated ?

They seem to do it as I've done in [the snippet I linked yesterday](https://gist.github.com/wil93/e8a4aad9205ab9e772c0) (they just call `build_l10n` what I called `custom_build`). It's a possibility, but I'm not sure about it...

setup.py
wil93: No problem. Anyway I think we can get rid of them: it seems that they fix the permissions directly on the file system, so it looks like it's "enough" to run them once and commit the permissions that changed (that, in my test, happened to be none).

Btw, we could avoid using package_data completely and [switch to a MANIFEST.in file](http://blog.codekills.net/2011/07/15/lies,-more-lies-and-python-packaging-documentation-on--package_data-/).

wil93: Nope, package_data must stay, or else `./setup.py install` will not work (we could add `MANIFEST.in` though, that is needed if one wants to do `./setup.py sdist`).

stefano-maggiolo: Ok to get rid of it then if you tested. If there isn't a need for MANIFEST now let's not add it. I'd say.

docs/Installation.rst
stefano-maggiolo: I'd rephrase this sentence to avoid the informal feel, but I'll let you decide if and how :)

stefano-maggiolo:
I'd say we submit this as it is (mod style change if any) - since the prerequisite makes sense at least for isolate. If we manage to find a good pythonic way to build and package the mo files, we can move it to the setup.

stefano-maggiolo:
Ok, can you rebase these commits, and retest everything? When you're done I'll do some more testing and merge

wil93:
Done: https://github.com/wil93/cms/commits/less_sudo_rebased

stefano-maggiolo:
Ok, I have (at least) a few things before pushing:
1. in Ubuntu at least activate is not an executable in virtualenv, so you need "source ...activate" (or equivalent) in the docs;
2. there are some dependencies required to install everything via pip: at least libpq-dev and libcups2-dev in Ubuntu; it should be noted before mentioning pip install -r...
3. if possible - pip tries to compile our python testcases, is there a way to avoid it?

wil93:
1. The `. activate` should be equivalent to `source activate`.
2. Right, [these](https://github.com/cms-dev/cms/pull/363/files#diff-3254677a7917c6c01f55212f86c57fbfR30), I forgot about them... I will add them today.
3. Doesn't also `setup.py install` try to do that?

stefano-maggiolo:
1. Ok
2. Ok
3. I think so, I was just asking if we can improve easily (otherwise np). I tried filtering out the cmstestsuite packages but no luck.

wil93:
> I tried filtering out the cmstestsuite packages but no luck.

Maybe if `__init__.py` is removed from test directories?

stefano-maggiolo:
Merged, thank you.

aa8f5a66f303992135f4ea29f558656b2228ec21
fb29b1e520c9a251cd17bfdc438d5d486e949775

PR #429, opened 2015-05-31
AWS stuff

stefano-maggiolo:
Ignore the first two, that are in the other PR.


lerks:
Basic suggestion: why don't we organize the file hierarchy differently? Instead of having `templates/contest`, `templates/admin`, `handlers/admin` we could have `contest/templates`, `admin/templates` and `admin/handlers`. This would make `contest` and `admin` two sub-packages of `cms.server`, each containing its dedicated files. This would reduce mix-ups and confusions. Also static files could be properly split up and moved to `contest/static` and `admin/static`.

stefano-maggiolo:
Merged starting from 4432d33b6d02b1397d44a0775cdfe16146cfb7f5

PR #430, opened 2015-06-01
Split CWS' handlers

stefano-maggiolo:


stefano-maggiolo:
Finally GH understand when I push a PR without pressing the button!

PR #431, opened 2015-06-03
Add desktop notification support to AWS and CWS

wil93:
notif

Currently, it's kinda annoying because every operation in AWS triggers a notification (you will get a lot of "Operation successful"). However is very handy for contestants (and for admins who may forget to look at the "questions" page).

### DIFF
--- a/cms/server/static/aws_utils.js
+++ b/cms/server/static/aws_utils.js
@@ -40,7 +40,9 @@ CMS.AWSUtils = function(url_root, timestamp,
     this.file_asked_url = "";
 
     // Ask permission for desktop notifications
-    Notification.requestPermission();
+    if ("Notification" in window) {
+        Notification.requestPermission();
+    }
 };
 
 
--- a/cms/server/static/aws_utils.js
+++ b/cms/server/static/aws_utils.js
@@ -38,6 +38,9 @@ CMS.AWSUtils = function(url_root, timestamp,
     this.remaining_div = null;
     this.file_asked_name = "";
     this.file_asked_url = "";
+
+    // Ask permission for desktop notifications
+    Notification.requestPermission();
 };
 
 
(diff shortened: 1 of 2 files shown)


wil93:
It seems to be broken on Firefox (it works only if the `icon` field is removed from the "options" object passed to `new Notification()`).

lerks:
I think it's very useful for CMS (contestants usually spend time coding, not on the browser, and it's good they can see notifications anyway). On the other hand I also think it's annoying for AWS. It should only show the notification that are not triggered by an action the admin just performed (these can be new questions, failed evaluations, offline workers, ...).

wil93:
Yep, the right way to do that is probably to add a boolean parameter to the JS notification function (show_in_desktop) or to just decide based on the "level" of the notification (I don't know how that is currently used).

However, it's not *that* annoying after all. What I'd really like to change is the behavior of `Notification.onclick`. Currently, it does nothing apart from hiding the notification (which, since it has not been closed, will stack up in my DE's bottom panel). Instead, I'd like to make it so that when you click a notification you're brought to the browser, specifically to the browser tab that triggered the notification.

[Here](https://github.com/cryptocat/cryptocat/issues/421) a method is devised, but it seems to be chrome-specific. Maybe the best way is to just open up a new window? (this could be better because this new window could point directly to the "communication" page, and because a contestant could have more than a single tab on the contest page when they click the notification).

lerks:
Of course, for this feature to be merged, it has to work on Firefox as well. If you cannot find a way to have the `icon` work, please remove it.

As for when showing the notification, I have two proposals. Either show only the notifications that are not "informative" (i.e. show the warnings and the errors), or show a notification only when the user is not currently looking at the page [1]. As I'm not currently sure how that API behaves (e.g. what happens when the page is a foreground tab of a background window?), some testing would be needed. We could also combine the two ideas.

I wouldn't open a new window or perform any strange behavior upon click: telling the user there's something he has to take care of is enough; he'll then know how to access the page and look for the details.

[1] https://developer.mozilla.org/en-US/docs/Web/Guide/User_experience/Using_the_Page_Visibility_API

lerks:
Forget the visibility API: I just did some tests and it seems that, by specification, when "in doubt" (another window is covering the browser) it is preferred to assume the page is visible, rather than hidden. It's the opposite of what we'd like to have.

stefano-maggiolo:
Picking up this from Luca. I tested it and works with both Firefox and Chrome, but no icon in both (not that I'm worried, as soon as they will support it, we already have it there). I agree with the fact that AWS is not suited to send its notifications in this form, so I'll submit after removing AWS from this.

If you want to have it in AWS for non-admin-triggered action feel free to send another PR, I'd like that.

stefano-maggiolo:
Merged here: 70ca55065fb90fc556f12e7ad34b67ba8bbdb66a - as said please send an updated version for AWS! Thanks.

PR #440, opened 2015-06-27
Handle multistep compilation better

artikz:
Keep each compilation step's stdout and stderr in the sandbox (for future analysis), and
properly stop after first fail.


stefano-maggiolo:
Merged (with small changes) at e55e3ffe53123e299692578209e8640122dad8a8

PR #444, opened 2015-07-05
Add option to allow / disallow submissions download

artikz:
Admins can control from AWS whether contestants are allowed to download their submissions. This option maybe useful for security reasons.

### DIFF
--- a/cms/db/__init__.py
+++ b/cms/db/__init__.py
@@ -79,7 +79,7 @@ __all__ = [
 
 # Instantiate or import these objects.
 
-version = 15
+version = 16
 
 
 engine = create_engine(config.database, echo=config.database_debug,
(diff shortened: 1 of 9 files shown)


stefano-maggiolo:
Just curious, why do you need this?

artikz:
For example, you gave somebody access to CMS to solve a problem and he submitted several solutions. Then username and password of the somebody were stolen by another somebody. I don't want that another somebody to be able to get anything from this :)

It doesn't meant to use in a real supervised contest, more in a contest where you don't have control on who has access to the system and whether they exchange their credentials (voluntarily or unintentionally).

wil93:
I, for one, needed this option sometime ago (and I ended up editing ContestWebServer.py by raising a 403 error if a user *which wasn't hidden* tried to download a submission)

stefano-maggiolo:
Make sense. Merged: c8c3a4b37a10589a72d12a0badee0c8c448af227

PR #447, opened 2015-07-14
Administrative accounts

stefano-maggiolo:
This was longer than expected... but I believe that the code is better organized now.

The end structure is comprised of:
- a cookie-based authentication middleware that can be passed to WebService; this translate authentication cookies into authentication headers for the underlying WSGI apps, and viceversa (the app can say when it wants a cookie to be set, deleted, or refreshed);
- the tornado app is similar to CWS, just it doesn't do any cookie-based authentication; the templates change based on the capabilities of the current user;
- the rpc middleware has a new parameter, a function that does authorization based on the cookie.

The capability _jobs that Giovanni suggested is missing, but it is simple enough to add it now (just add the field, change the reevaluation template to enable the buttons in that case, and add the rpc methods to the whitelist).

As for the commits, they are as self-contained as possible. Feel free to review them one at a time, but I think you'd like to see the complete result as well.

One thing that is missing is the possibility of revoking a cookie-based authentication (that in CWS happens when you change the password), because I wanted to avoid DB access in the authentication middleware. The admins can be disabled, or deleted and recreated, so I think that's not a big problem.

[Review on Reviewable](https://reviewable.io/reviews/cms-dev/cms/447)

### DIFF
--- a/cms/server/admin/handlers/base.py
+++ b/cms/server/admin/handlers/base.py
@@ -131,14 +131,17 @@ def parse_ip_address_or_subnet(value):
     return value
 
 
-def require_permission(permission="authenticated"):
+def require_permission(permission="authenticated", self_allowed=False):
     """Return a decorator requiring a specific admin permission level
 
     The default value, "authenticated", just checkes that the admin is
     logged in. All other values also implicitly require it. Therefore,
     there is no need to use tornado.web.authenticated if using this.
 
-    permission (string): one of the permission levels
+    permission (string): one of the permission levels.
+    self_allowed (bool): if true, interpret the first argument as the
+       admin id that can execute the method regardless of their
+       permission.
 
     """
     if permission not in [BaseHandler.PERMISSION_ALL,
@@ -164,7 +167,13 @@ def require_permission(permission="authenticated"):
             if user.permission_all or user.__getattribute__(permission_key):
                 return func(self, *args, **kwargs)
             else:
-                raise tornado.web.HTTPError(403, "Admin is not authorized")
+                if self_allowed and len(args) > 0 and int(args[0]) == user.id:
+                    # First argument is assumed to be the admin id,
+                    # encoded as a unicode object, and should match
+                    # the current user id.
+                    return func(self, *args, **kwargs)
+                else:
+                    raise tornado.web.HTTPError(403, "Admin is not authorized")
 
         return newfunc
 
--- a/cms/server/admin/handlers/base.py
+++ b/cms/server/admin/handlers/base.py
@@ -35,16 +35,18 @@ import logging
 import traceback
 
 from datetime import datetime, timedelta
+from functools import wraps
 
 import tornado.web
 
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm import joinedload
 
-from cms.db import Contest, Participation, Question, Session, \
+from cms.db import Admin, Contest, Participation, Question, \
     Submission, SubmissionFormatElement, SubmissionResult, Task, Team, User
 from cms.grading.scoretypes import get_score_type_class
 from cms.grading.tasktypes import get_task_type_class
+from cms.io import WebService
 from cms.server import CommonRequestHandler, file_handler_gen, get_url_root
 from cmscommon.datetime import make_datetime
 
@@ -129,6 +131,46 @@ def parse_ip_address_or_subnet(value):
     return value
 
 
+def require_permission(permission="authenticated"):
+    """Return a decorator requiring a specific admin permission level
+
+    The default value, "authenticated", just checkes that the admin is
+    logged in. All other values also implicitly require it. Therefore,
+    there is no need to use tornado.web.authenticated if using this.
+
+    permission (string): one of the permission levels
+
+    """
+    if permission not in [BaseHandler.PERMISSION_ALL,
+                          BaseHandler.PERMISSION_MESSAGING,
+                          BaseHandler.AUTHENTICATED]:
+        raise ValueError("Invalid permission level %s." % permission)
+
+    def decorator(func):
+        """Decorator for requiring a permission level
+
+        """
+        @wraps(func)
+        @tornado.web.authenticated
+        def newfunc(self, *args, **kwargs):
+            """Check if the permission is present before calling the function.
+
+            """
+            if permission == BaseHandler.AUTHENTICATED:
+                return func(self, *args, **kwargs)
+
+            user = self.current_user
+            permission_key = "permission_%s" % permission
+            if user.permission_all or user.__getattribute__(permission_key):
+                return func(self, *args, **kwargs)
+            else:
+                raise tornado.web.HTTPError(403, "Admin is not authorized")
+
+        return newfunc
+
+    return decorator
+
+
 class BaseHandler(CommonRequestHandler):
     """Base RequestHandler for this application.
 
@@ -136,6 +178,9 @@ class BaseHandler(CommonRequestHandler):
     child of this class.
 
     """
+    PERMISSION_ALL = "all"
+    PERMISSION_MESSAGING = "messaging"
+    AUTHENTICATED = "authenticated"
 
     def try_commit(self):
         """Try to commit the current session.
@@ -157,6 +202,43 @@ class BaseHandler(CommonRequestHandler):
                 make_datetime(),
                 "Operation successful.", "")
             return True
+
+    def get_current_user(self):
+        """Gets the current admin from cookies.
+
+        return (Admin|None): if a valid cookie is retrieved, return
+            the Admin object, otherwise None.
+
+        """
+        admin_id = self.request.headers.get(
+            WebService.AUTHENTICATED_USER_HEADER, None)
+        if admin_id is None:
+            self.set_cookie_action("delete")
+            return None
+
+        # Load admin.
+        admin = self.sql_session.query(Admin)\
+            .filter(Admin.id == admin_id)\
+            .filter(Admin.enabled.is_(True))\
+            .first()
+        if admin is None:
+            self.set_cookie_action("delete")
+            return None
+
+        # Maybe refresh the cookie.
+        if self.refresh_cookie:
+            self.set_cookie_action("refresh")
+
+        return admin
+
+    def set_cookie_action(self, action):
+        """Set the header to inform the upper WSGI layer to do the specified
+        action on the login cookie.
+
+        action (unicode): the value of the cookie.
+
+        """
+        self.set_header(WebService.AUTHENTICATED_COOKIE_HEADER, action)
 
     def safe_get_item(self, cls, ident, session=None):
         """Get item from database of class cls and id ident, using
@@ -183,12 +265,7 @@ class BaseHandler(CommonRequestHandler):
         """This method is executed at the beginning of each request.
 
         """
-        # Attempt to update the contest and all its references
-        # If this fails, the request terminates.
-        self.set_header("Cache-Control", "no-cache, must-revalidate")
-
-        self.sql_session = Session()
-        self.sql_session.expire_all()
+        super(BaseHandler, self).prepare()
         self.contest = None
 
     def render_params(self):
@@ -451,16 +528,24 @@ class BaseHandler(CommonRequestHandler):
 FileHandler = file_handler_gen(BaseHandler)
 
 
-def SimpleHandler(page):
-    class Cls(BaseHandler):
-        def get(self):
-            self.r_params = self.render_params()
-            self.render(page, **self.r_params)
+def SimpleHandler(page, authenticated=True):
+    if authenticated:
+        class Cls(BaseHandler):
+            @require_permission(BaseHandler.AUTHENTICATED)
+            def get(self):
+                self.r_params = self.render_params()
+                self.render(page, **self.r_params)
+    else:
+        class Cls(BaseHandler):
+            def get(self):
+                self.r_params = self.render_params()
+                self.render(page, **self.r_params)
     return Cls
 
 
 def SimpleContestHandler(page):
     class Cls(BaseHandler):
+        @require_permission(BaseHandler.AUTHENTICATED)
         def get(self, contest_id):
             self.contest = self.safe_get_item(Contest, contest_id)
 
--- a/cms/server/admin/handlers/base.py
+++ b/cms/server/admin/handlers/base.py
@@ -287,6 +287,8 @@ class BaseHandler(CommonRequestHandler):
         params["timestamp"] = make_datetime()
         params["contest"] = self.contest
         params["url_root"] = get_url_root(self.request.path)
+        if self.current_user is not None:
+            params["current_user"] = self.current_user
         if self.contest is not None:
             params["phase"] = self.contest.phase(params["timestamp"])
             # Keep "== None" in filter arguments. SQLAlchemy does not
--- a/cms/conf.py
+++ b/cms/conf.py
@@ -107,6 +107,7 @@ class Config(object):
         # AdminWebServer.
         self.admin_listen_address = ""
         self.admin_listen_port = 8889
+        self.admin_cookie_duration = 10 * 60 * 60  # 10 hours
 
         # ProxyService.
         self.rankings = ["http://usern4me:passw0rd@localhost:8890/"]
--- a/cms/db/__init__.py
+++ b/cms/db/__init__.py
@@ -54,6 +54,8 @@ __all__ = [
     "Contest", "Announcement",
     # user
     "User", "Team", "Participation", "Message", "Question",
+    # admin
+    "Admin",
     # task
     "Task", "Statement", "Attachment", "SubmissionFormatElement", "Dataset",
     "Manager", "Testcase",
@@ -93,6 +95,7 @@ from .types import RepeatedUnicode
 from .base import metadata, Base
 from .contest import Contest, Announcement
 from .user import User, Team, Participation, Message, Question
+from .admin import Admin
 from .task import Task, Statement, Attachment, SubmissionFormatElement, \
     Dataset, Manager, Testcase
 from .submission import Submission, File, Token, SubmissionResult, \
--- a/cms/db/admin.py
+++ b/cms/db/admin.py
@@ -1,0 +1,88 @@
+#!/usr/bin/env python2
+# -*- coding: utf-8 -*-
+
+# Contest Management System - http://cms-dev.github.io/
+# Copyright © 2015 Stefano Maggiolo <s.maggiolo@gmail.com>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Admin-related database interfaces for SQLAlchemy.
+
+"""
+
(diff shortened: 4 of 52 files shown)


cms/server/admin/handlers/base.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+    #!/usr/bin/env python2
```
wil93: Why these leading spaces?

stefano-maggiolo: Fixed and rebased immediately, thanks. If you want to review, I'll appreciate.

wil93:
I think this needs to be rebased because of #449.


---

Reviewed 2 of 2 files at r1, 1 of 1 files at r2, 6 of 6 files at r3, 3 of 3 files at r4, 6 of 6 files at r5, 19 of 19 files at r6, 2 of 2 files at r7, 4 of 4 files at r8, 15 of 15 files at r9, 4 of 4 files at r10.

---

<sup>**[cms/server/admin/handlers/admin.py, line 132 \[r7\]](https://reviewable.io:443/reviews/cms-dev/cms/447#-K25Fom_2NpMAIWlu-sP)** ([raw file](https://github.com/cms-dev/cms/blob/09b9cf58953f38aa2933c12e5ae2eab2714fd521/cms/server/admin/handlers/admin.py#L132)):</sup>
This TODO can be deleted now?


---

<sup>**[cms/server/admin/handlers/main.py, line 61 \[r6\]](https://reviewable.io:443/reviews/cms-dev/cms/447#-K24a4TiUW_R_MDw1qgR)** ([raw file](https://github.com/cms-dev/cms/blob/e51033248fdee624b537cbd479bd3c8cc0a88c2f/cms/server/admin/handlers/main.py#L61)):</sup>
"Nonexistent" maybe? Unexisting is less common. (not very important)


---

<sup>**[cmscontrib/AddAdmin.py, line 81 \[r2\]](https://reviewable.io:443/reviews/cms-dev/cms/447#-K24SDCAAmjHI9VtDdza)** ([raw file](https://github.com/cms-dev/cms/blob/1ab563c2eb09e021a940cc676afe51a8d53461ec/cmscontrib/AddAdmin.py#L81)):</sup>
You probably want to add `"cmsAddAdmin=cmscontrib.AddAdmin:main"` to the entry_points list in setup.py.

By the way, if it will eventually be assumed that CMS is installed, then all these `if`s can be deleted (because setuptools creates runnable scripts that import the module and call the right method, so the condition is always false).


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cms-dev/cms/447)

giomasce:
Reviewed 2 of 2 files at r1, 1 of 1 files at r2, 6 of 6 files at r3, 3 of 3 files at r4.
Review status: all files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[cms/db/admin.py, line 38 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/447#-K7YmTgV2Z8ieeHyrxKV:-K7YmTgWuYPkUmsy3INq:-1850175796) ([raw file](https://github.com/cms-dev/cms/blob/a690478d9abfd8a6e8f74b72c1d5d51ac99e8151/cms/db/admin.py#L38)):*
Small typo.

---

*[cms/server/admin/templates/admins.html, line 48 \[r3\]](https://reviewable.io:443/reviews/cms-dev/cms/447#-K7YpIe7OSlqKZLOn66b:-K7YpIe891FfVcc6Hlk6:321067739) ([raw file](https://github.com/cms-dev/cms/blob/00efc18ad1efb98fc31f5cd920135da32749353a/cms/server/admin/templates/admins.html#L48)):*
Probably we should hide this snippet until we actually have that feature!

---

*[requirements.txt, line 14 \[r4\]](https://reviewable.io:443/reviews/cms-dev/cms/447#-K7YphLL75CKPnulvnSg:-K7YphLL75CKPnulvnSh:1058852015) ([raw file](https://github.com/cms-dev/cms/blob/e4c7ac5f52131a715828663d8a43d869664bb123/requirements.txt#L14)):*
This must be also written in the documentation

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cms-dev/cms/447)*

giomasce:
Reviewed 5 of 6 files at r5.
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[cms/io/web_service.py, line 73 \[r5\]](https://reviewable.io:443/reviews/cms-dev/cms/447#-K7YtQQIeFYX9cR7CmDq:-K7YtQQJQOE_GTAfbIoC:417915456) ([raw file](https://github.com/cms-dev/cms/blob/c1f8018574b75d35be49d11c93fe4e01104cfcc8/cms/io/web_service.py#L73)):*
Can we erase all headers matching `X-Cms-*`? I find it more future-proof with respect to security.

---

*[cms/server/admin/server.py, line 130 \[r5\]](https://reviewable.io:443/reviews/cms-dev/cms/447#-K7Yu4qwfp-HHBTHgd3e:-K7Yu4qwfp-HHBTHgd3f:-765603841) ([raw file](https://github.com/cms-dev/cms/blob/c1f8018574b75d35be49d11c93fe4e01104cfcc8/cms/server/admin/server.py#L130)):*
This constant is already set in `web_service.py`. Can we deduplicate it?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cms-dev/cms/447)*

giomasce:
Reviewed 1 of 6 files at r5.
Review status: all files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[cms/server/admin/server.py, line 111 \[r5\]](https://reviewable.io:443/reviews/cms-dev/cms/447#-K7a2HPU8TpmtZq4bgXu:-K7a2HPVakPXjI3RCiKB:-189351632) ([raw file](https://github.com/cms-dev/cms/blob/c1f8018574b75d35be49d11c93fe4e01104cfcc8/cms/server/admin/server.py#L111)):*
It would be nice to have specific tests for security-related features, but I do not have time to write them.

---

*[cms/server/admin/server.py, line 181 \[r5\]](https://reviewable.io:443/reviews/cms-dev/cms/447#-K7a24H17oAGtB7FmnWY:-K7a24H3JoepRzR4cszu:-1929970197) ([raw file](https://github.com/cms-dev/cms/blob/c1f8018574b75d35be49d11c93fe4e01104cfcc8/cms/server/admin/server.py#L181)):*
What does `httponly` do? I could not find any reference.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cms-dev/cms/447)*

giomasce:
Reviewed 11 of 19 files at r6.
Review status: all files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[cms/server/admin/handlers/base.py, line 143 \[r6\]](https://reviewable.io:443/reviews/cms-dev/cms/447#-K7a3vUrBO4Ku1vpjGlI:-K7a3vUsnL7x3ONajc7r:482752150) ([raw file](https://github.com/cms-dev/cms/blob/e51033248fdee624b537cbd479bd3c8cc0a88c2f/cms/server/admin/handlers/base.py#L143)):*
Is there a reason for not reusing the same constants when the decorator is invoked? Using a constant instead of an in-line string seems more robust to me.

---

*[cms/server/admin/handlers/base.py, line 220 \[r6\]](https://reviewable.io:443/reviews/cms-dev/cms/447#-K7a4XEMfQeRi9dxBsue:-K7a4XEMfQeRi9dxBsuf:964887620) ([raw file](https://github.com/cms-dev/cms/blob/e51033248fdee624b537cbd479bd3c8cc0a88c2f/cms/server/admin/handlers/base.py#L220)):*
`admin_id` should already be an `int` (it is converted in `AWSAuthmiddleware._authenticate()`.

---

*[cms/server/admin/handlers/base.py, line 220 \[r6\]](https://reviewable.io:443/reviews/cms-dev/cms/447#-K7a6T0m81UyfUkbb43F:-K7a6T0nXzGYTcHXVdG-:2017400722) ([raw file](https://github.com/cms-dev/cms/blob/e51033248fdee624b537cbd479bd3c8cc0a88c2f/cms/server/admin/handlers/base.py#L220)):*
Shouldn't we filter out disabled users here? Otherwise, where does that happen?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cms-dev/cms/447)*

giomasce:
Review status: all files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[cms/server/admin/handlers/base.py, line 220 \[r6\]](https://reviewable.io:443/reviews/cms-dev/cms/447#-K7a6T0m81UyfUkbb43F:-K7blvywHQ5dqxNfQo7x:-1721044443) ([raw file](https://github.com/cms-dev/cms/blob/e51033248fdee624b537cbd479bd3c8cc0a88c2f/cms/server/admin/handlers/base.py#L220)):*
Ok, I found where it happens (in the login handler). Still, I would expect that disabling a user has immediate effect, so the same check should happen also here.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cms-dev/cms/447)*

giomasce:
Reviewed 3 of 19 files at r6.
Review status: all files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[cms/server/admin/handlers/submission.py, line 91 \[r6\]](https://reviewable.io:443/reviews/cms-dev/cms/447#-K7bmil7GQgBuZNqlldR:-K7bmil8xJxPHr9iNfsl:899405071) ([raw file](https://github.com/cms-dev/cms/blob/e51033248fdee624b537cbd479bd3c8cc0a88c2f/cms/server/admin/handlers/submission.py#L91)):*
Shouldn't be `"messaging"` here?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cms-dev/cms/447)*

stefano-maggiolo:
I'm going to push once solving all the comments, then rebase and push again. Will comment here when finished.

---

Review status: 31 of 51 files reviewed at latest revision, 13 unresolved discussions, some commit checks failed.

---

*[cms/db/admin.py, line 38 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/447#-K7YmTgV2Z8ieeHyrxKV:-K7jGkR0pjMcnppqtNhV:-499146009) ([raw file](https://github.com/cms-dev/cms/blob/a690478d9abfd8a6e8f74b72c1d5d51ac99e8151/cms/db/admin.py#L38)):*
Done.

---

*[cms/io/web_service.py, line 73 \[r5\]](https://reviewable.io:443/reviews/cms-dev/cms/447#-K7YtQQIeFYX9cR7CmDq:-K7jHXzplUGDf6k-P5O6:1673924650) ([raw file](https://github.com/cms-dev/cms/blob/c1f8018574b75d35be49d11c93fe4e01104cfcc8/cms/io/web_service.py#L73)):*
It doesn't look like with the fixers used here: http://werkzeug.pocoo.org/docs/0.11/contrib/fixers/#werkzeug.contrib.fixers.HeaderRewriterFix

I'll add a TODO.

---

*[cms/server/admin/handlers/admin.py, line 132 \[r7\]](https://reviewable.io:443/reviews/cms-dev/cms/447#-K25Fom_2NpMAIWlu-sP:-K7jItLhORrdQqQzNY0d:-499146009) ([raw file](https://github.com/cms-dev/cms/blob/09b9cf58953f38aa2933c12e5ae2eab2714fd521/cms/server/admin/handlers/admin.py#L132)):*
Done.

---

*[cms/server/admin/handlers/base.py, line 143 \[r6\]](https://reviewable.io:443/reviews/cms-dev/cms/447#-K7a3vUrBO4Ku1vpjGlI:-K7jMwpJuyBAdla5UiFq:-499146009) ([raw file](https://github.com/cms-dev/cms/blob/e51033248fdee624b537cbd479bd3c8cc0a88c2f/cms/server/admin/handlers/base.py#L143)):*
Done.

---

*[cms/server/admin/handlers/base.py, line 220 \[r6\]](https://reviewable.io:443/reviews/cms-dev/cms/447#-K7a4XEMfQeRi9dxBsue:-K7jOtUw_yax_l_qiDgs:-499146009) ([raw file](https://github.com/cms-dev/cms/blob/e51033248fdee624b537cbd479bd3c8cc0a88c2f/cms/server/admin/handlers/base.py#L220)):*
Done.

---

*[cms/server/admin/handlers/base.py, line 220 \[r6\]](https://reviewable.io:443/reviews/cms-dev/cms/447#-K7a6T0m81UyfUkbb43F:-K7jPDc1dvmgPLUbZ3_F:-1832984908) ([raw file](https://github.com/cms-dev/cms/blob/e51033248fdee624b537cbd479bd3c8cc0a88c2f/cms/server/admin/handlers/base.py#L220)):*
Good catch, thanks.

---

*[cms/server/admin/handlers/main.py, line 61 \[r6\]](https://reviewable.io:443/reviews/cms-dev/cms/447#-K24a4TiUW_R_MDw1qgR:-K7jPScx-yY7GGDDOaVe:-499146009) ([raw file](https://github.com/cms-dev/cms/blob/e51033248fdee624b537cbd479bd3c8cc0a88c2f/cms/server/admin/handlers/main.py#L61)):*
Done.

---

*[cms/server/admin/handlers/submission.py, line 91 \[r6\]](https://reviewable.io:443/reviews/cms-dev/cms/447#-K7bmil7GQgBuZNqlldR:-K7jPYfFmrh3yAJj62zS:-626595490) ([raw file](https://github.com/cms-dev/cms/blob/e51033248fdee624b537cbd479bd3c8cc0a88c2f/cms/server/admin/handlers/submission.py#L91)):*
Comment is not meant to be contestant-visible

---

*[cms/server/admin/server.py, line 130 \[r5\]](https://reviewable.io:443/reviews/cms-dev/cms/447#-K7Yu4qwfp-HHBTHgd3e:-K7jSFVpfelRpIfQg05B:1137445462) ([raw file](https://github.com/cms-dev/cms/blob/c1f8018574b75d35be49d11c93fe4e01104cfcc8/cms/server/admin/server.py#L130)):*
It was actually a leftover...

---

*[cms/server/admin/server.py, line 181 \[r5\]](https://reviewable.io:443/reviews/cms-dev/cms/447#-K7a24H17oAGtB7FmnWY:-K7jPrEtR_1sAxZIUyB8:-1823542770) ([raw file](https://github.com/cms-dev/cms/blob/c1f8018574b75d35be49d11c93fe4e01104cfcc8/cms/server/admin/server.py#L181)):*
That the browser should enforce that JS cannot access the cookie.

---

*[cms/server/admin/templates/admins.html, line 48 \[r3\]](https://reviewable.io:443/reviews/cms-dev/cms/447#-K7YpIe7OSlqKZLOn66b:-K7jQ-23fbqjvYsbYNVi:2101406455) ([raw file](https://github.com/cms-dev/cms/blob/00efc18ad1efb98fc31f5cd920135da32749353a/cms/server/admin/templates/admins.html#L48)):*
Right.

---

*[cmscontrib/AddAdmin.py, line 81 \[r2\]](https://reviewable.io:443/reviews/cms-dev/cms/447#-K24SDCAAmjHI9VtDdza:-K7jQId1aUJjm3NnhGTw:1765455268) ([raw file](https://github.com/cms-dev/cms/blob/1ab563c2eb09e021a940cc676afe51a8d53461ec/cmscontrib/AddAdmin.py#L81)):*
Not going to be assumed, but thanks

---

*[requirements.txt, line 14 \[r4\]](https://reviewable.io:443/reviews/cms-dev/cms/447#-K7YphLL75CKPnulvnSg:-K7jRCmOKa2sbttGrORN:-499146009) ([raw file](https://github.com/cms-dev/cms/blob/e4c7ac5f52131a715828663d8a43d869664bb123/requirements.txt#L14)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cms-dev/cms/447#-:-K7jRbGNRX_vlXm57TKQ:1368813601)*

stefano-maggiolo:
Done. Reviewable merged both pushes to the same revision, but luckily the diff is nice enough (r10 -> r11). As expected, the rebase introduced changes almost only in the tests. Feel free to check this last revision.

---

Review status: 12 of 52 files reviewed at latest revision, 13 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cms-dev/cms/447#-:-K7kPT_MXuAhdgPWaPv7:804179481)*

giomasce:
Would it be possibile to have an option (i.e., in `cms.conf`) to disable AWS login altogether? In many use cases it was simply not needed and having to manage another set of username and password is annoying.

---

Reviewed 5 of 19 files at r6, 2 of 2 files at r7, 4 of 4 files at r8.
Review status: 12 of 52 files reviewed at latest revision, 12 unresolved discussions.

---

*[cms/server/admin/templates/base.html, line 69 \[r6\]](https://reviewable.io:443/reviews/cms-dev/cms/447#-K7raBZ8eY8-cOEIvq5z:-K7raBZ8eY8-cOEIvq6-:579343879) ([raw file](https://github.com/cms-dev/cms/blob/e51033248fdee624b537cbd479bd3c8cc0a88c2f/cms/server/admin/templates/base.html#L69)):*
When can this actually happen? Is it in order to support the possibility to use AWS without login or does it do anything else?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cms-dev/cms/447#-:-K7rZ8mkEGXO1j3Zz1rD:-361280952)*

giomasce:
Reviewed 15 of 15 files at r9, 4 of 4 files at r10.
Review status: 12 of 52 files reviewed at latest revision, 11 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cms-dev/cms/447)*

giomasce:
Reviewed 21 of 42 files at r11.
Review status: 33 of 52 files reviewed at latest revision, 6 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cms-dev/cms/447)*

giomasce:
Reviewed 21 of 42 files at r11.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cms-dev/cms/447)*

stefano-maggiolo:
I agree (maybe as a command line switch for aws, rather than as a configuration option, I'd like to hear your preference too). But in a different PR :)

---

---

*[cms/server/admin/templates/base.html, line 69 \[r6\]](https://reviewable.io:443/reviews/cms-dev/cms/447#-K7raBZ8eY8-cOEIvq5z:-K7tcaX_hdqCEcVOrl4l:1062755509) ([raw file](https://github.com/cms-dev/cms/blob/e51033248fdee624b537cbd479bd3c8cc0a88c2f/cms/server/admin/templates/base.html#L69)):*
I think nothing uses it, but maybe it's still useful to prevent errors if we forget to redirect to the login page

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cms-dev/cms/447#-:-K7tc6gdktbcJNP5B9nR:1293007229)*

stefano-maggiolo:
This one I just pushed is probably the last update, just a couple of cosmetic changes. 

---

Review status: 49 of 52 files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cms-dev/cms/447#-:-K7tfX2KrHRBZnOO_GbN:701940464)*

giomasce:
Ok. I think everything's good, you can push.

I personally would prefer a configuration option, because usually either you want AWS admins or you don't. It is not something you may want to change often between two runs of the same installation.

Another thing that has to be fixed is the user documentation, particularly with respect to the part where we say that AWS has to be protected because it has no login.

---

Reviewed 3 of 3 files at r12.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cms-dev/cms/447#-:-K7uLgPQBV8CVIwmMfQr:1381851627)*

giomasce:
Also documentation can go in another PR. Please open issues to remind us of missing docs and login disabling option.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cms-dev/cms/447#-:-K7uMa0Cir-7f51biOGX:-420119912)*

stefano-maggiolo:
Merged

PR #448, opened 2015-07-15
Fix bulk submissions invalidation

artikz:
This fixes bulk submissions invalidation buttons on task, participation and submissions pages.


stefano-maggiolo:
fa58fe0d73278f86395d8abb87862662b6aeb6ee

Thanks!

PR #452, opened 2015-07-19
Pagination

stefano-maggiolo:
[Review on Reviewable](https://reviewable.io/reviews/cms-dev/cms/452)


cms/server/admin/handlers/base.py
```diff
@@ -410,6 +411,40 @@ def get_score_type(self, dest, name, params):
...
+        page_size(int): the number of submissions per page.
+
+        """
+        query = query\
+            .options(joinedload(Submission.task))\
+            .options(joinedload(Submission.participation))\
+            .options(joinedload(Submission.files))\
+            .options(joinedload(Submission.token))\
+            .options(joinedload(Submission.results))\
+            .order_by(Submission.timestamp.desc())
```
lerks: Some of the code that uses this function will then access, for example, the SubmissionResult's evaluations. Maybe we should join-load these here as well, to avoid a query being fired for each submission in those cases. (For example, the contestsubmission handler shows a table with all evaluations in case the submission is still scoring)

cms/server/admin/handlers/views.py
```diff
@@ -0,0 +1,83 @@
...
+        if submission_id is not None:
+            # A specific submission for a specific dataset (if
+            # specified) or for the live dataset.
+            url += "submission/%s" % submission_id
+            invalidate_arguments["submission_id"] = submission_id
+            if dataset_id is not None:
+                url += "/%s" % dataset_id
+                invalidate_arguments["dataset_id"] = dataset_id
+        elif participation_id is not None:
+            # All submissions of the participation of this participation
```
lerks: Typo?

cms/server/admin/handlers/views.py
```diff
@@ -0,0 +1,83 @@
...
+            url += "contest/%s/user/%s" % (contest_id, user_id)
```
lerks: As arguments are mutually exclusive, aren't these supposed to be None?

cms/server/admin/handlers/views.py
```diff
@@ -0,0 +1,83 @@
...
+        contest_id (int): the id of the contest containing the
+            submission results to invalidate.
+        submission_id (int|None): id of the submission to invalidate,
+            or None.
+        dataset_id (int|None): id of the dataset to invalidate, or
+            None.
+        participation_id (int|None): id of the participation to
+            invalidate, or None.
+        user_id (int|None): id of the user (mandatory if participation_id
+            is given, to write the redirect).
```
lerks: So, they're not really all mutually exclusive. I was tricked by that first sentence, as I didn't read the docstring through. Maybe say that there isn't just that "only" exception. Moreover, in the code you don't consider the case in which only user_id is provided: is that a case you allow?

lerks: Also, I see that in all uses you make of this class you always have the whole objects at your disposal, not just the IDs. Why not pass the objects as arguments then? This way you could, for example, extract the user ID from the participation directly.

cms/server/admin/templates/fragments/submission_rows.html
```diff
@@ -1,10 +1,22 @@
...
+    {% else %}
+  {{ i + 1 }}
+    {% end %}
+  {% end %}
+</div>
+{% end %}
+
 <table class="bordered">
   <thead>
     <tr>
```
lerks: Could you remove the  id="paged_content_submissions" from the tbody? It shouldn't be necessary anymore.

cms/server/admin/templates/fragments/submission_rows.html
```diff
@@ -1,10 +1,22 @@
...
-{% if submissions == [] %}
+{% if submission_count == 0 %}
 <p>No submissions found.</p>
 
 {% else %}
-{# TODO: implement a more useful pagination for a large contest #}
-<div id="page_selector_submissions"></div>
+
+{% if submission_count > len(submissions) %}
+<div id="page_selector_submissions">
```
lerks: And this id="" as well, I guess... (see below)

cms/server/admin/templates/fragments/submission_rows.html
```diff
@@ -1,10 +1,22 @@
...
+  Pages:
+  {% for i in range(submission_pages) %}
+    {% if i != submission_page %}
+  <a href="{{ url_root }}{{ submission_url }}?page={{ i }}">{{ i + 1 }}</a>
```
lerks: I think that "page_url" would be a better name than "submission_url": this got me a little confused at first.

cms/server/admin/templates/participation.html
```diff
@@ -37,125 +37,18 @@
...
-        <td>
-          {% set reevaluation_par_name = "submission" %}
-          {% set reevaluation_par_value = s.id %}
-          {% set reevaluation_par_dataset_id = None %}
-          {% include fragments/reevaluation_buttons.html %}
-        </td>
-      </tr>
-      {% end %}
-    </tbody>
-  </table>
```
lerks: I totally agree that this code, that was almost duplicated, had to be removed. The "almost" is that, here, the user column was removed, as it value was constant. Whereas now you reintroduced it. No big harm. I suggest that, to handle this properly, the submission table could also be made a UIModule. (I'm not saying you should do it for this PR, it's a proposal for the future)

cms/server/admin/handlers/views.py
```diff
@@ -0,0 +1,83 @@
...
+               user_id=None):
+        """Render reevaluation buttons for the given filters.
+
+        Apart from contest_id, the parameters are mutually exclusive
+        (i.e., only one can be provided). The only exception is when
+        providing a submission, in which case a specific dataset can
+        also be passed (the live one is used by default).
+
+        url_root (unicode): path to the root of the server.
+        contest_id (int): the id of the contest containing the
```
lerks: Can contest_id be None? Here you say no, but you give it a None default value, and in activate_dataset.html in fact you don't pass any explicit contest_id

cms/server/admin/handlers/base.py
stefano-maggiolo: Done.

cms/server/admin/templates/fragments/submission_rows.html
stefano-maggiolo: Done (also in the three calling files of course).

cms/server/admin/templates/fragments/submission_rows.html
stefano-maggiolo: Done.

cms/server/admin/templates/fragments/submission_rows.html
stefano-maggiolo: Done.

cms/server/admin/handlers/views.py
stefano-maggiolo: I double checked where I call this, and documented following that.

cms/server/admin/handlers/views.py
stefano-maggiolo: Yep, this method is a bit confusing. One additional problem is that the URL for the participation is using contest and user instead of just participation.

cms/server/admin/handlers/views.py
stefano-maggiolo: Done.

cms/server/admin/handlers/contestuser.py
```diff
@@ -140,9 +140,12 @@ def get(self, contest_id, user_id):
         if participation is None:
             raise tornado.web.HTTPError(404)
 
-        self.r_params = self.render_params()
+        submission_query = self.sql_session.query(Submission)\
+            .filter(Submission.participation_id == participation.id)
```
lerks: Any reason to put the equality on the IDs instead of on the objects themselves? I think we usually prefer the latter if there's no big disadvantage... (e.g. an additional query firing)

cms/server/admin/handlers/views.py
```diff
@@ -0,0 +1,92 @@
...
+            or None.
+        dataset_id (int|None): id of the dataset to invalidate, or
+            None.
+        participation_id (int|None): id of the participation to
+            invalidate, or None.
+        user_id (int|None): id of the user (mandatory if participation_id
+            is given, to write the redirect).
+
+        """
+        url = "%s/" % url_root
```
lerks: This looks bad, as you're prepending the url_root to the url in the template as well.

stefano-maggiolo: Right. For some reason (Chrome, I'd guess) having more ../ than needed doesn't give any error. I should set up my testing env in such a way that url_root is not trivial... Anyway, for the moment I removed the url_root from the template.

cms/server/admin/handlers/contestuser.py
stefano-maggiolo: Done.

PR #454, opened 2015-07-24
Update italy_yaml loader to support get_statement boolean

wil93:
After `get_statement` boolean parameter was added to `get_task` method, the cmsAddTask command is broken for the italy_yaml loader. This commit fixes that.


stefano-maggiolo:
Merged in 2e3c7f9a31e1cc40ce485da1cce6063eecb96939, thank you!

PR #457, opened 2015-08-04
Update italy_yaml and cmsMake

veluca93:
Now italy_yaml works with the changes made in c09c052.
Moreover, cmsMake uses the "no_statement" feature required to loaders
and ignores the statement if it's missing, instead of creating an empty
one.


wil93:
Part of these changes are in #454

stefano-maggiolo:
Merged in 4f11113068c8249e5fb5046a56c222477205ed7e - thanks

PR #461, opened 2015-08-06
Make cmsMake work with input.txt/output.txt batch tasks

stefano-maggiolo:


cmstaskenv/cmsMake.py
```diff
@@ -466,17 +466,44 @@ def make_output(n, assume=None):
...
+        copied_infile = os.path.join(
+            temp_dir,
+            "input.txt" if use_stdin else yaml_conf.get("infile"))
+        copied_outfile = os.path.join(
+            temp_dir,
+            "output.txt" if use_stdout else yaml_conf.get("outfile"))
+
+        shutil.copy(infile, copied_infile)
+        fin = io.open(copied_infile, "rb") if use_stdin else None
+        fout = io.open(copied_outfile, 'wb') if use_stdout else None
```
wil93: The 2 lines above should use the `with ... as fin:` and `with ... as fout:` syntax

stefano-maggiolo: Sadly that doesn't work with the fact that we are "maybe" opening those files.

stefano-maggiolo:
Merged in a7fda46f26206272b82d063d195d56c33a790631

PR #464, opened 2015-08-22
Adapt DumpImporter to the fact that Users are also toplevel objects in the dump

stefano-maggiolo:


stefano-maggiolo:
Done. Please check.

PR #465, opened 2015-08-26
Refactored the functional tests

stefano-maggiolo:
A new class to manage similar tests has been extracted, which we will
use in the future to do ES-stress tests. The other advantage is that
now all submissions are submitted before waiting for the results,
speeding up considerably the running time.


stefano-maggiolo:
Merged in 812ca25afde12014f089035f674e02c6b0909a76

PR #468, opened 2015-09-09
Use .editorconfig

wil93:
When used in conjunction with the editorconfig plugin, it helps maintaining consistent code formatting in the project.


stefano-maggiolo:
Merged! 52af128ca94eb3964cd4048528ba42ec88a3da02

PR #472, opened 2015-09-12
Fixed problem with empty partecipation password

### LINKED ISSUE #471: Hitting "update" inside AWS (even without changing anything) could have side effects
Steps to reproduce:

1. Create a user and a contest.
2. Add a participation for the user in the contest and go to the participation page.
3. The password field is "None" (because `Participation.password == None`).
4. Hit "update".
5. Now `Participation.password == "None"` (the actual string of length 4)

edomora97:
When a user is created and added to a contest it's contest password is None, but in the admin page the None value is converted to the 'None' string causing an unwanted side effect. If the user update the record the None value is converted to the 'None' password. With this patch the special '' (empty) password is converted internally to None.

Fix #471

### DIFF
--- a/cms/server/admin/handlers/contestuser.py
+++ b/cms/server/admin/handlers/contestuser.py
@@ -162,9 +162,7 @@ class ParticipationHandler(BaseHandler):
         try:
             attrs = participation.get_attrs()
 
-            self.get_string(attrs, "password")
-            if attrs["password"] == "":
-                attrs["password"] = None
+            self.get_string(attrs, "password", empty=None)
             self.get_ip_address_or_subnet(attrs, "ip")
             self.get_datetime(attrs, "starting_time")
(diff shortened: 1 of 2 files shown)


cms/server/admin/handlers/contestuser.py
```diff
@@ -163,6 +163,8 @@ def post(self, contest_id, user_id):
             attrs = participation.get_attrs()
 
             self.get_string(attrs, "password")
+            if attrs["password"] == "":
+                attrs["password"] = None
```
stefano-maggiolo: I think you can use get_string with empty=None - can you please check and amend the commit if true? Thank you Edoardo!

stefano-maggiolo:
Thank you, merged in 7ec6db4e99495083eb7af227a3ae71256d48dbdb

PR #473, opened 2015-09-13
Enable C11 support

dariost:
Since the **C++11** standard has already been enabled, the **C11** standard should be enabled too.


stefano-maggiolo:
Merged in a92264b634b95b7405c388949b0acf7075d6fda2 thanks!

PR #477, opened 2015-09-27
Add --as-root option to prerequisites.py

wil93:
This is useful when using docker, since you are root in the container

### DIFF
--- a/prerequisites.py
+++ b/prerequisites.py
@@ -145,9 +145,12 @@ def assert_root():
 
 def assert_not_root():
     """Check if the current user is *not* root, and exit with an error message
-    if needed.
+    if needed. If the --as-root flag is set, this function does nothing.
 
     """
+    if "--as-root" in sys.argv:
+        return
+
     if os.geteuid() == 0:
         print("[Error] You must *not* be root to do this, try avoiding 'sudo'")
         exit(1)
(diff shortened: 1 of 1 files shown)


prerequisites.py
```diff
@@ -148,6 +148,9 @@ def assert_not_root():
     if needed.
 
     """
+    if "--as-root" in sys.argv:
```
stefano-maggiolo: I guess you need to update the doc message.

prerequisites.py
```diff
@@ -159,6 +162,9 @@ def get_real_user():
     In the case of a user *actually being root* we return an error.
 
     """
+    if "--as-root" in sys.argv:
```
stefano-maggiolo: Ditto.

prerequisites.py
wil93: Done, however I'm not sure if it's necessary: the "-y" flag was undocumented as well, and both "-y" and "--as-root" should not be used (at least by an "end user" of cms), I think they are mainly for internal use. Also, I think that as cms gets simpler, the prerequisites.py script will gradually become smaller and less important.

stefano-maggiolo: Sorry for not being very precise - I meant the function documentation. Feel free to leave what you added though, I think it's nice to have it in the usage.

wil93: Oh ok, done :smile:

wil93:
Rebased

stefano-maggiolo:
Merged in 872082a9f18d6ecdfcc07f7be561e1913d0b8851

PR #478, opened 2015-09-28
Simplify installation and retrieval of translation files

wil93:
At the moment, `./prerequisites.py build_l10n` creates the `.mo` files, and `./prerequisites.py install_l10n` installs them.

With this PR, `./prerequisites.py install_l10n` is not needed anymore because the `.mo` files are installed by setup.py

This is cleaner because we can retrieve the files with pkg_resources instead of looking for specific paths.

[Review on Reviewable](https://reviewable.io/reviews/cms-dev/cms/478)

PR #479, opened 2015-09-29
Assume that CMS is installed

wil93:
Depends on #478.

At the moment, CMS checks if it's installed or run directly from source. This is messy because, right now, CMS believes that it's *not* installed when it's installed and run in a virtual environment.

[Review on Reviewable](https://reviewable.io/reviews/cms-dev/cms/479)


stefano-maggiolo:
I agree it's not the most nice way of doing things, but removing everything doesn't just remove the possibility of running it without having access to /*/local/*/cms?

wil93:
Yes.. maybe we should expand this section from the config

```
"_section": "System-wide configuration",

"temp_dir": "/tmp",
```

to include also "log_dir" and so on...

I think it would be the quickest solution to implement. The best thing to do, however, is probably to just manage the needed directories in a better way (logs, for example, could just be sent to a syslog server, instead of written to a file).

stefano-maggiolo:
Wouldn't it be simpler to find a way of detecting virtualenv + installed?

stefano-maggiolo:
Sorry, even if you have a point re installed in virtualenv not being detected correctly, this breaks other stuff. Closing for now.

Issue #460, opened 2015-08-06
Remaining tasks for 1.3


stefano-maggiolo:
This is a list of the remaining bugfixes / minor features before releasing CMS 1.3. It is mostly about things discovered at the IOI. The release date is planned to be within August.
- [ ] [patch prosed] Merge IP autologin feature.
- [X] Accept the AWS pagination branch (which includes a refactoring of the reevaluation buttons).
- [X] Fix RWS initial loading of data.
- [X] Pick up partially evaluated submission in ES's sweep.
- [ ] [patch existing] Add an "invalidate outcome" to reevaluate without deleting the existing evaluations objects (and add relevant documentation).
- [X] Fix the participation password field in AWS (weird when None is selected).
- [X] Remove any mention of localization from AWS.
- [X] [patch presented] PS sends submissions before tasks in certain occasions (resulting in a 400 in RWS).
- [X] [not reproducible] Delete attachments redirects to a non-existing page in AWS.
- [X] [working as intended] Look at DumpExporter argument parsing ("-c <id> target" does not work).

There are also these three bigger topics, which may or may not end up in 1.3.
- [x] Think if there is a simple way to fix the evaluation_tries (which were not updated after the split of evaluations).
- [ ] Reinstate JobGroups, and use them to distribute more ESOperations to a Worker in case the queue is big.
- [ ] Check if enabling background judging on a hidden dataset with ES active causes problems.

stefano-maggiolo:
Let's update this at < 2 months from the release (this does not include the PRs already in flight):

- [ ] Real Java support (status: patch existing, but needs improvements).
- [ ] Reinstate JobGroup (status: todo).
- [ ] Command line switch in ES to optimize for throughput or latency, i.e. whether to send one testcase at a time or all at once to workers (status: todo).
- [ ] Use a cache of recent workers results in ES and write to DB only every so often (status: patch existing, but needs improvements).
- [ ] Remove all checks for "job not done" in ES (status: waiting for other patches).

Nice to have, but not required, in reverse order of importance:
- [ ] A simple implementation of analysis mode.
- [ ] Reorganize cmscontrib, in particular differentiate between the scripts types (importing, inserting from command line, removing).
- [ ] RWS visualizations (PR in flight, but stalled).
- [ ] Admin account (PR in flight, but stalled).

lerks:
1.3 has been released, this is obsolete.

PR #480, opened 2015-10-02
Fix RWS's error 400 when starting ProxyService with more than 0 subs

wil93:
Sometimes RWS will get stuck and produce warnings about "invalid data".
Apparently, this can be caused by the fact that ProxyService starts by
sending the submissions/subchanges related data first, and then it will
proceed to the contests/teams/users related data. Now, if there are
already more than 0 submissions, this means that ProxyService will send
data about these submissions even though it has never initialized the
rankings.

This commit enqueues the initialization operation and *then* starts the
sweeper. I'm not sure if this is the right thing to do, but it works.

### DIFF
--- a/cms/service/ProxyService.py
+++ b/cms/service/ProxyService.py
@@ -262,10 +262,11 @@ class ProxyService(TriggeredService):
         self.rankings = list()
         for ranking in config.rankings:
             self.add_executor(ProxyExecutor(ranking.encode('utf-8')))
-        self.start_sweeper(347.0)
 
-        # Send some initial data to rankings.
+        # Enqueue the dispatch of some initial data to rankings.
         self.initialize()
+
+        self.start_sweeper(347.0)
 
     def _missing_operations(self):
         """Return a generator of data to be sent to the rankings..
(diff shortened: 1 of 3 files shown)


stefano-maggiolo:
Awesome, this was marked as "not reproducible" in https://github.com/cms-dev/cms/issues/460 - thanks for tracking this down!

wil93:
Great :smile:

By the way, I think I've fixed another (somewhat unrelated) bug, which caused RWS to give a status 500 when trying to merge new submissions.

To reproduce the bug:

1. Empty the /var/local/lib/cms/ranking directory.
2. Start cmsRankingWebServer.
3. Start cmsResourceService and wait for RWS to get all the data
4. Shut RWS down and then start it again (here the bug is triggered, because RWS won't load the *submissions* data from disk).
5. Kill cmsProxyService and let it restart.
6. RWS will merge the new data for contests, tasks, users, but will give a status 500 when trying to merge the subsmissions.

wil93:
I noticed that, maybe, the status 500 bug was the same as the "Fix RWS initial loading of data" one present in the list (for which a patch allegedly exists).

stefano-maggiolo:
Yeah, I have a bunch of small bugfixes that still need to be included. I'll
try to get to them soon(ish). Sorry 'bout that.

stefano-maggiolo:
Merged both (after changing a little bit the comments & messages). Grazie!
f3c624f87b559d5bb504bf670df9fd1a74e1aee9
e3951d3ec06c4e82788ab8440f977af6a930bc31

PR #482, opened 2015-10-05
Add desktop notification support to AWS

wil93:
Continuation of #431.


stefano-maggiolo:
Oh, I didn't realize it was just one line more. Merged in 9f00b7c7087d4a888ded86509edd5f8c65ff5e43

wil93:
I did't either, actually :smile: I initially thought that we had to add a show_in_desktop argument to the display_notification function, but then I noticed that the type parameter was enough.

PR #487, opened 2015-10-22
Fix minor issue that sometimes happened when using symlinks

wil93:
For example, if you had this folder structure:

```
tasks/
  mytask/
    ...
    mytask2.yaml
contest/
  {mytask2 --> ../tasks/mytask}
```

You would get and error when trying to import mytask2 from contest folder.

[Review on Reviewable](https://reviewable.io/reviews/cms-dev/cms/487)

### DIFF
--- a/cmscontrib/AddContest.py
+++ b/cmscontrib/AddContest.py
@@ -73,7 +73,7 @@ class ContestImporter(BaseImporter):
         self.no_statements = no_statements
         self.file_cacher = FileCacher()
 
-        self.loader = loader_class(os.path.realpath(path), self.file_cacher)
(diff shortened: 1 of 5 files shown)


stefano-maggiolo:
I guess this makes sense (even if the setup is a bit weird. @giomasce, do you see any problem in not following symlinks?


---

Reviewed 5 of 5 files at r1.
Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cms-dev/cms/487)

stefano-maggiolo:
59936621261668b9f90dd6df3716bf0e46e2524e

PR #492, opened 2015-10-30
Avoid spawning too many desktop notifications

wil93:
Apparently, AWS wasn't the only one spawning notifications for just
about everything. CWS does that too, when you:

 * send a submission
 * use a token on a submission

[Review on Reviewable](https://reviewable.io/reviews/cms-dev/cms/492)


stefano-maggiolo:
Merged in bc268b3697e807d2e4622e103a1e2678662f7550

PR #500, opened 2015-11-23
Fix cmsAddUser: do not try to add a user that already exists

lucach:
This aligns AddUser with AddContest and AddTask.

Adding a user without checking if it is a duplicate wastes an ID each time.

[Review on Reviewable](https://reviewable.io/reviews/cms-dev/cms/500)

### DIFF
--- a/cmscontrib/AddUser.py
+++ b/cmscontrib/AddUser.py
@@ -7,6 +7,7 @@
 # Copyright © 2010-2011 Stefano Maggiolo <s.maggiolo@gmail.com>
 # Copyright © 2010-2011 Matteo Boscariol <boscarim@hotmail.com>
 # Copyright © 2014-2015 William Di Luigi <williamdiluigi@gmail.com>
+# Copyright © 2015 Luca Chiodini <luca@chiodini.org>
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as
@@ -43,9 +44,8 @@ import logging
 import os
 
 from cms import utf8_decoder
-from cms.db import SessionGen
+from cms.db import SessionGen, User
 from cms.db.filecacher import FileCacher
-from sqlalchemy.exc import IntegrityError
 
 from cmscontrib.loaders import choose_loader, build_epilog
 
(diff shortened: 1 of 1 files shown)


wil93:
If I'm not mistaken this doesn't change the script's behavior, right?

If that's the case, I think the old approach (with `try`-`catch`) was more pyhonic, as for [EAFP](https://docs.python.org/2/glossary.html#term-eafp).

lucach:
You're right, it's equivalent except that this way we do not create "holes" in the IDs progression.

wil93:
That's a good point... Maybe the most clean solution is to add `session.rollback()` just after `except IntegrityError:`, I'm not sure if it does what I think it does (it should).

wil93:
Reviewed 1 of 1 files at r1.

---



_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cms-dev/cms/500)_

wil93:
Reviewed 1 of 1 files at r2.

---



_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cms-dev/cms/500)_

lucach:
Rebased after f8ba5b9
If we want to keep this style, it should be also changed the new cmsAddTeam.

stefano-maggiolo:
I'm honestly agnostic with respect to this change :)
IDs are not guaranteed nor required to be consecutive, and on the other hand I don't particularly prefer try-except over if-else... Maybe slightly in favor of if-else as IntegrityError in theory could be raised by other problems.

If we want to make this an abstract style discussion though, I'd prefer the fast-return pattern:

```python
old_user = [...]
if old_user is not None:
  [log]
  return
[add user
```

Mostly because in other situations the else branch could be large.
Can you change it?

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cms-dev/cms/500#-:-K4EDJC8Q8Z0o1M-snfL:-1725016436)*

lucach:
Done.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cms-dev/cms/500#-:-K4M_XPZxFPMC05Sr3lo:-499146009)*

stefano-maggiolo:
Reviewed 1 of 1 files at r4.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cms-dev/cms/500)*

PR #502, opened 2015-11-29
Move IP restrictions and hidden user block to contest configuration

stefano-maggiolo:
[Review on Reviewable](https://reviewable.io/reviews/cms-dev/cms/502)


lerks:
Apart from comments, OK to push.

lerks:
OK to push

PR #505, opened 2015-11-30
Use team.id instead of team.code

wil93:
[Review on Reviewable](https://reviewable.io/reviews/cms-dev/cms/505)


stefano-maggiolo:
Merged.

PR #506, opened 2015-12-13
Quick fix to support changing what version of python to use

wil93:
I know that we should look into making compilers "separate" from CMS (with a plugin system). In the meantime, though, I think it's good to have a simple way to switch the python version used when evaluating python submissions.

[Review on Reviewable](https://reviewable.io/reviews/cms-dev/cms/506)

### DIFF
--- a/cms/conf.py
+++ b/cms/conf.py
@@ -57,6 +57,7 @@ class Config(object):
         self.temp_dir = "/tmp"
         self.backdoor = False
         self.file_log_debug = False
+        self.python_version = "2"
 
         # Database.
         self.database = "postgresql+psycopg2://cmsuser@localhost/cms"
--- a/cms/grading/__init__.py
+++ b/cms/grading/__init__.py
@@ -243,13 +243,19 @@ def get_compilation_commands(language, source_filenames, executable_filename,
         # The executable name is fixed, and there is no way to specify
         # the name of the pyc, so we need to bundle together two
         # commands (compilation and rename).
-        # In order to use Python 3 change them to:
-        # /usr/bin/python3 -m py_compile %s
-        # mv __pycache__/%s.*.pyc %s
-        py_command = ["/usr/bin/python2", "-m", "py_compile",
-                      source_filenames[0]]
-        mv_command = ["/bin/mv", "%s.pyc" % os.path.splitext(os.path.basename(
-                      source_filenames[0]))[0], executable_filename]
+        py_command = ["/usr/bin/python" + config.python_version, "-m",
+                      "py_compile", source_filenames[0]]
+
+        basename = os.path.splitext(os.path.basename(
+                                    source_filenames[0]))[0]
+
+        if config.python_version == "3":
+            pyc_name = "__pycache__/%s.cpython-34.pyc" % basename
+        else:
+            pyc_name = "%s.pyc" % basename
+
+        mv_command = ["/bin/mv", pyc_name, executable_filename]
+
         commands.append(py_command)
         commands.append(mv_command)
     elif language == LANG_PHP:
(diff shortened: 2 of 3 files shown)


stefano-maggiolo:
Thanks for doing this, indeed I don't think we'll get around to change to a plugin system before 1.3. Do you see a good place where to document this new feature?

---

Reviewed 2 of 2 files at r1.

---

*[cms/grading/__init__.py, line 250 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/506#-K5_Gi4cHrjavaCxDmnP:-K5_Gi4cHrjavaCxDmnQ:-116978710) ([raw file](https://github.com/cms-dev/cms/blob/6cc94e21572ed61560eec2c8e68122853a517e7f/cms/grading/__init__.py#L250)):*
Can we not depend on the specific string "cpython-34", which I assume means python 3.4? https://www.python.org/dev/peps/pep-3147/#proposal suggests that that string identifies the version and I think it can be treated as an opaque id.

---

*[cms/grading/__init__.py, line 254 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/506#-K5_GaelXB13xvjCe9o6:-K5_GaelXB13xvjCe9o7:369948937) ([raw file](https://github.com/cms-dev/cms/blob/6cc94e21572ed61560eec2c8e68122853a517e7f/cms/grading/__init__.py#L254)):*
It would be cleaner to extract the basename and apply it immediately to both pyc_name

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cms-dev/cms/506#-:-K5_HYlI6vJ-MziQjszU:-323640886)*

wil93:
We could put it in cms.conf.sample, or just hint at the existence of that option here http://cms.readthedocs.org/en/v1.2/Configuring%20a%20contest.html#programming-languages

(In the meantime I've done both)

---

Review status: 1 of 4 files reviewed at latest revision, 2 unresolved discussions.

---

*[cms/grading/__init__.py, line 250 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/506#-K5_Gi4cHrjavaCxDmnP:-K5aRy_sn1uErB9L2BbM:850909643) ([raw file](https://github.com/cms-dev/cms/blob/6cc94e21572ed61560eec2c8e68122853a517e7f/cms/grading/__init__.py#L250)):*
I tried to use `*` but it doesn't work... so I thought it was enough to use the current version (for ubuntu 14.04)

---

*[cms/grading/__init__.py, line 254 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/506#-K5_GaelXB13xvjCe9o6:-K5aUPiqrgd3GpsCD6mV:-499146009) ([raw file](https://github.com/cms-dev/cms/blob/6cc94e21572ed61560eec2c8e68122853a517e7f/cms/grading/__init__.py#L254)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cms-dev/cms/506#-:-K5aR9TuX9jZCRUgYuC9:-230321549)*

stefano-maggiolo:
Honestly I would revert it from cms.conf.sample - it is unlikely that a random user would use it and on the other hand it might be confusing.

---

Reviewed 3 of 3 files at r2.

---

*[cms/grading/__init__.py, line 250 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/506#-K5_Gi4cHrjavaCxDmnP:-K5akx-lZrDvtBmyp-HI:-1273396913) ([raw file](https://github.com/cms-dev/cms/blob/6cc94e21572ed61560eec2c8e68122853a517e7f/cms/grading/__init__.py#L250)):*
I think you can inspect the content of the directory and find out the right name :)

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cms-dev/cms/506#-:-K5akfHNIcU1ELrJ3wdr:1677893882)*

wil93:
Done.

---

---

*[cms/grading/__init__.py, line 250 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/506#-K5_Gi4cHrjavaCxDmnP:-K5aoQJpktCaTJEMgT4Q:-614850810) ([raw file](https://github.com/cms-dev/cms/blob/6cc94e21572ed61560eec2c8e68122853a517e7f/cms/grading/__init__.py#L250)):*
Mmmm actually I'm not sure that the directory even exists at the time of executing that line of code :confused:

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cms-dev/cms/506#-:-K5aoKJZ8G7i2jlFwYpQ:-499146009)*

stefano-maggiolo:
This will be superseded by the language plugins that are in the works.

PR #507, opened 2015-12-13
Fix cmscommon.archive when self.temp_dir is None

wil93:
[Review on Reviewable](https://reviewable.io/reviews/cms-dev/cms/507)


stefano-maggiolo:
Merged :)

PR #508, opened 2015-12-14
A few more ES-related things

stefano-maggiolo:
Apart from a few logging changes, and a small bugfix, there are two more interesting changes:
1. one to try to limit the integrity errors, now that we have removed the check; this is not big, but needs some thoughts as the effecs might be larger than expected;
2. change of the sweeper from querying all submissions to python and then filtering to filtering in the db; this is a large change, but luckily it has tests :)

[Review on Reviewable](https://reviewable.io/reviews/cms-dev/cms/508)


stefano-maggiolo:
Closing this as I made some more changes. Will repropose later.

PR #512, opened 2015-12-22
Fix regression when activating a dataset

wil93:
Because of this, users whose score had changed weren't being
autoselected anymore.

[Review on Reviewable](https://reviewable.io/reviews/cms-dev/cms/512)


stefano-maggiolo:
Reviewed 2 of 2 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cms-dev/cms/512)*

stefano-maggiolo:
Merged.

PR #513, opened 2015-12-24
Add script to allow exporting submissions to files

wil93:
[Review on Reviewable](https://reviewable.io/reviews/cms-dev/cms/513)

### DIFF
--- a/cmscontrib/ExportSubmissions.py
+++ b/cmscontrib/ExportSubmissions.py
@@ -1,0 +1,213 @@
+#!/usr/bin/env python2
+# -*- coding: utf-8 -*-
+
+# Contest Management System - http://cms-dev.github.io/
+# Copyright © 2015-2016 William Di Luigi <williamdiluigi@gmail.com>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Utility to export submissions to a folder.
+
+"""
+
+from __future__ import absolute_import
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import argparse
+import logging
+import os
+import sys
+import codecs
+
+from cms import utf8_decoder, LANG_C, LANG_CPP, LANG_JAVA, LANG_PASCAL, \
+    LANG_PHP, LANG_PYTHON
+from cms.db import Dataset, File, FSObject, Participation, SessionGen, \
+    Submission, SubmissionResult, Task, User
+
+
+logger = logging.getLogger(__name__)
+
+
+_RAW_TEMPLATE_DATA = """
+* user:  %s
+* fname: %s
+* lname: %s
+* task:  %s
+* score: %s
+* date:  %s
+"""
+
+
+TEMPLATE = {
+    LANG_C: "/**%s*/\n" % _RAW_TEMPLATE_DATA,
+    LANG_PASCAL: "(**%s*)\n" % _RAW_TEMPLATE_DATA,
+    LANG_PYTHON: "\"\"\"%s\"\"\"\n" % _RAW_TEMPLATE_DATA,
+    LANG_PHP: "<?php\n/**%s*/\n?>" % _RAW_TEMPLATE_DATA,
+}
+TEMPLATE[LANG_CPP] = TEMPLATE[LANG_C]
+TEMPLATE[LANG_JAVA] = TEMPLATE[LANG_C]
+
+
+def filter_top_scoring(results, unique):
+    """Filter results keeping only the top scoring submissions for each user
+    and task
+
+    results ([Submission]): the starting list of submissions
+    unique (bool): if True, keep only the first top-scoring submission
+
+    return ([Submission]): the filtered submissions
+
+    """
+    usertask = {}
+    for row in results:
+        key = (row[6], row[10])  # u_id, t_id
+        value = (-row[3], row[2], row)  # sr_score, s_timestamp
+        if unique:
+            if key not in usertask or usertask[key][0] > value:
+                usertask[key] = [value]
+        else:
+            if key not in usertask or usertask[key][0][0] > value[0]:
+                usertask[key] = [value]
+            elif usertask[key][0][0] == value[0]:
+                usertask[key].append(value)
+
+    results = []
+    for key, values in usertask.iteritems():
+        for value in values:
+            results.append(value[2])  # the "old" row
+
+    return results
+
+
+def main():
+    """Parse arguments and launch process.
+
+    """
+    parser = argparse.ArgumentParser(
+        description="Export CMS submissions to a folder.")
+    parser.add_argument("-c", "--contest-id", action="store", type=int,
+                        help="id of contest (default: all contests)")
+    parser.add_argument("-t", "--task-id", action="store", type=int,
+                        help="id of task (default: all tasks)")
+    parser.add_argument("-u", "--user-id", action="store", type=int,
+                        help="id of user (default: all users)")
+    parser.add_argument("-s", "--submission-id", action="store", type=int,
+                        help="id of submission (default: all submissions)")
+    parser.add_argument("--add-info", action="store_true",
+                        help="if set, information on the submission will"
+                             " be added in the first lines of each file")
+    parser.add_argument("--min-score", action="store", type=float,
+                        help="ignore submissions which scored strictly"
+                             " less than this (default: 0.0)",
+                        default=0.0)
+    parser.add_argument("--filename", action="store", type=utf8_decoder,
+                        help="the filename format to use"
+                             " (default: {id}.{name}.{ext})",
+                        default="{id}.{name}.{ext}")
+    parser.add_argument("output_dir", action="store", type=utf8_decoder,
+                        help="directory where to save the submissions")
+
+    group = parser.add_mutually_exclusive_group(required=False)
+    group.add_argument("--unique", action="store_true",
+                       help="if set, only the earliest best submission"
+                            " will be exported for each (user, task)")
+    group.add_argument("--best", action="store_true",
+                       help="if set, only the best submissions will be"
+                            " exported for each (user, task)")
+
+    args = parser.parse_args()
+
+    if not os.path.exists(args.output_dir):
+        os.mkdir(args.output_dir)
+    if not os.path.isdir(args.output_dir):
+        logger.critical("The output-dir parameter must point to a directory")
+        sys.exit(1)
+
+    with SessionGen() as session:
+        q = session.query(Submission)\
+            .join(Submission.task)\
+            .join(Submission.files)\
+            .join(Submission.results)\
+            .join(SubmissionResult.dataset)\
+            .join(Submission.participation)\
+            .join(Participation.user)\
+            .filter(Dataset.id == Task.active_dataset_id)\
+            .filter(SubmissionResult.score >= args.min_score)\
+            .with_entities(Submission.id, Submission.language, Submission.timestamp,
+                           SubmissionResult.score, File.filename, File.digest,
+                           User.id, User.username, User.first_name, User.last_name,
+                           Task.id, Task.name)
+
+        if args.contest_id:
+            q = q.filter(Participation.contest_id == args.contest_id)
+
+        if args.task_id:
+            q = q.filter(Submission.task_id == args.task_id)
+
+        if args.user_id:
+            q = q.filter(Participation.user_id == args.user_id)
+
+        if args.submission_id:
+            q = q.filter(Submission.id == args.submission_id)
+
+        results = q.all()
+
+        if args.unique or args.best:
+            results = filter_top_scoring(results, args.unique)
+
+        print("%s file(s) will be created." % len(results))
+        if raw_input("Continue? [Y/n] ").lower() not in ["y", ""]:
+            sys.exit(0)
+
+        done = 0
+        for row in results:
+            s_id, s_language, s_timestamp, sr_score, f_filename, f_digest, \
+                u_id, u_name, u_fname, u_lname, t_id, t_name = row
+
+            name = f_filename
+            if name.endswith(".%l"):
+                name = name[:-3]  # remove last 3 chars
+
+            filename = args.filename.format(id=s_id, name=name, ext=s_language,
+                                            time=s_timestamp, user=u_name)
+            filename = os.path.join(args.output_dir, filename)
+            if os.path.exists(filename):
+                logger.warning("Skipping file '%s' because it already exists",
+                               filename)
+
+            fso = FSObject.get_from_digest(f_digest, session)
+            assert fso is not None
+            with fso.get_lobject(mode="rb") as file_obj:
+                data = file_obj.read().decode('utf-8')
+
+                if args.add_info:
+                    data = TEMPLATE[s_language] % (
+                        u_name,
+                        u_fname,
+                        u_lname,
+                        t_name,
+                        sr_score,
+                        s_timestamp
+                    ) + data
+
+                with codecs.open(filename, "w", encoding="utf-8") as file_out:
+                    file_out.write(data)
+
+            done += 1
+            print(done, "/", len(results))
+
+
+if __name__ == "__main__":
+    sys.exit(main())
--- a/setup.py
+++ b/setup.py
@@ -112,6 +112,7 @@ setup(
             "cmsAddTask=cmscontrib.AddTask:main",
             "cmsRemoveTask=cmscontrib.RemoveTask:main",
             "cmsComputeComplexity=cmscontrib.ComputeComplexity:main",
+            "cmsExportSubmissions=cmscontrib.ExportSubmissions:main",
             "cmsAddContest=cmscontrib.AddContest:main",
             "cmsAddAdmin=cmscontrib.AddAdmin:main",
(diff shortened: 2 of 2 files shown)


stefano-maggiolo:
Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[cmscontrib/ExportSubmissions.py, line 60 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/513#-K7gkYJZVGIeLj8WfVZp:-K7gkYJZVGIeLj8WfVZq:62786199) ([raw file](https://github.com/cms-dev/cms/blob/dcb96d6c3936a0ee39f2cd4de708f68e546274a6/cmscontrib/ExportSubmissions.py#L60)):*
Please support all languages (see cms/__init__.py)

---

*[cmscontrib/ExportSubmissions.py, line 100 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/513#-K7gkmLPqJdohEHErE_J:-K7gkmLQxmlipX7Ms3A3:2013377617) ([raw file](https://github.com/cms-dev/cms/blob/dcb96d6c3936a0ee39f2cd4de708f68e546274a6/cmscontrib/ExportSubmissions.py#L100)):*
argparse handles this on its own: https://docs.python.org/2/library/argparse.html#argument-groups

---

*[cmscontrib/ExportSubmissions.py, line 139 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/513#-K7gtWqJQJw0cixEzX3c:-K7gtWqJQJw0cixEzX3d:-692850208) ([raw file](https://github.com/cms-dev/cms/blob/dcb96d6c3936a0ee39f2cd4de708f68e546274a6/cmscontrib/ExportSubmissions.py#L139)):*
This is weird. I think you want
```python
q = session.query(Submission)\
    .join(Submission.task)\
    .join(Submission.files)\
    .join(Submission.results)\
    .join(SubmissionResult.dataset)\
    .join(Submission.participation)\
    .join(Participation.user)\
    .filter(Dataset.id == Task.active_dataset_id)\
    .filter(SubmissionResult.score >= args.min_score)
```

and later results are submission, so you can use alchemy object instead of tuples. Or if you prefer using tuples, add a `.withentities(...)` at the end.

---

*[cmscontrib/ExportSubmissions.py, line 155 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/513#-K7gwDPhe7bJKxxu_Vk_:-K7gwDPhe7bJKxxu_Vka:-1632517715) ([raw file](https://github.com/cms-dev/cms/blob/dcb96d6c3936a0ee39f2cd4de708f68e546274a6/cmscontrib/ExportSubmissions.py#L155)):*
I'd move this to a function
```python
def filter_top_scoring(results, unique):
    """Filter results keeping only the top scoring submissions for each user and task

    results ([Submission]): [...]
    unique (bool): if True, keep only the first top-scoring submission

    return ([Submission]): [...]

    """
```

---

*[cmscontrib/ExportSubmissions.py, line 160 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/513#-K7gvKosgFmeQcbgg_FM:-K7gvKotkN_BCW_Z3x0A:1442796468) ([raw file](https://github.com/cms-dev/cms/blob/dcb96d6c3936a0ee39f2cd4de708f68e546274a6/cmscontrib/ExportSubmissions.py#L160)):*
Why `-row[3]`?

---

*[cmscontrib/ExportSubmissions.py, line 178 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/513#-K7gveIu27bHqV7pm060:-K7gveIvStCv9cWrxbS5:2126929950) ([raw file](https://github.com/cms-dev/cms/blob/dcb96d6c3936a0ee39f2cd4de708f68e546274a6/cmscontrib/ExportSubmissions.py#L178)):*
`"%s file(s) will be created." % len(result)`

---

*[cmscontrib/ExportSubmissions.py, line 180 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/513#-K7gvpoR5mZ9dnjv8Bs3:-K7gvpoR5mZ9dnjv8Bs4:-195056732) ([raw file](https://github.com/cms-dev/cms/blob/dcb96d6c3936a0ee39f2cd4de708f68e546274a6/cmscontrib/ExportSubmissions.py#L180)):*
sys.exit

---

*[cmscontrib/ExportSubmissions.py, line 191 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/513#-K7gx0Vs6u-oUwoovj6P:-K7gx0Vs6u-oUwoovj6Q:-1122443893) ([raw file](https://github.com/cms-dev/cms/blob/dcb96d6c3936a0ee39f2cd4de708f68e546274a6/cmscontrib/ExportSubmissions.py#L191)):*
Don't you want to offer more formatting options? Timestamp, username are both useful, I guess.

---

*[cmscontrib/ExportSubmissions.py, line 216 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/513#-K7gxMiVvxZPAsADVljW:-K7gxMiWC586vQWejW21:-1263979176) ([raw file](https://github.com/cms-dev/cms/blob/dcb96d6c3936a0ee39f2cd4de708f68e546274a6/cmscontrib/ExportSubmissions.py#L216)):*
```
if __name__ == "__main__":
    sys.exit(main())
```

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cms-dev/cms/513)*

wil93:
Review status: 1 of 2 files reviewed at latest revision, 9 unresolved discussions.

---

*[cmscontrib/ExportSubmissions.py, line 60 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/513#-K7gkYJZVGIeLj8WfVZp:-K8SgWm5MPirG1jEWrax:-499146009) ([raw file](https://github.com/cms-dev/cms/blob/dcb96d6c3936a0ee39f2cd4de708f68e546274a6/cmscontrib/ExportSubmissions.py#L60)):*
Done.

---

*[cmscontrib/ExportSubmissions.py, line 100 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/513#-K7gkmLPqJdohEHErE_J:-K8SgWHfBiNFt4-Cl3dD:-499146009) ([raw file](https://github.com/cms-dev/cms/blob/dcb96d6c3936a0ee39f2cd4de708f68e546274a6/cmscontrib/ExportSubmissions.py#L100)):*
Done.

---

*[cmscontrib/ExportSubmissions.py, line 139 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/513#-K7gtWqJQJw0cixEzX3c:-K8SkwUU26h0OW-we2rg:-499146009) ([raw file](https://github.com/cms-dev/cms/blob/dcb96d6c3936a0ee39f2cd4de708f68e546274a6/cmscontrib/ExportSubmissions.py#L139)):*
Done.

---

*[cmscontrib/ExportSubmissions.py, line 155 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/513#-K7gwDPhe7bJKxxu_Vk_:-K8Slp8xEAA4_HCc-oMk:-499146009) ([raw file](https://github.com/cms-dev/cms/blob/dcb96d6c3936a0ee39f2cd4de708f68e546274a6/cmscontrib/ExportSubmissions.py#L155)):*
Done.

---

*[cmscontrib/ExportSubmissions.py, line 160 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/513#-K7gvKosgFmeQcbgg_FM:-K7hrg3A0ji69g8rBC7v:-585566516) ([raw file](https://github.com/cms-dev/cms/blob/dcb96d6c3936a0ee39f2cd4de708f68e546274a6/cmscontrib/ExportSubmissions.py#L160)):*
It's to have that for two submissions s1 and s2 it will hold s1 < s2 when s1's score is greater than s2's score, or when the scores are equal but s1's timestamp is less. So, the minimum = the earliest among the best submissions.

---

*[cmscontrib/ExportSubmissions.py, line 178 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/513#-K7gveIu27bHqV7pm060:-K8SkS12wsA-M3Yt96mj:-499146009) ([raw file](https://github.com/cms-dev/cms/blob/dcb96d6c3936a0ee39f2cd4de708f68e546274a6/cmscontrib/ExportSubmissions.py#L178)):*
Done.

---

*[cmscontrib/ExportSubmissions.py, line 180 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/513#-K7gvpoR5mZ9dnjv8Bs3:-K8SkWaNlwxXbcO1Eq6d:-499146009) ([raw file](https://github.com/cms-dev/cms/blob/dcb96d6c3936a0ee39f2cd4de708f68e546274a6/cmscontrib/ExportSubmissions.py#L180)):*
Done.

---

*[cmscontrib/ExportSubmissions.py, line 191 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/513#-K7gx0Vs6u-oUwoovj6P:-K8Sld3XjgKiVzACiCx0:-499146009) ([raw file](https://github.com/cms-dev/cms/blob/dcb96d6c3936a0ee39f2cd4de708f68e546274a6/cmscontrib/ExportSubmissions.py#L191)):*
Done.

---

*[cmscontrib/ExportSubmissions.py, line 216 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/513#-K7gxMiVvxZPAsADVljW:-K8Sldn-3cFRZQqEpKfl:-499146009) ([raw file](https://github.com/cms-dev/cms/blob/dcb96d6c3936a0ee39f2cd4de708f68e546274a6/cmscontrib/ExportSubmissions.py#L216)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cms-dev/cms/513)*

stefano-maggiolo:
Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[cmscontrib/ExportSubmissions.py, line 139 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/513#-K7gtWqJQJw0cixEzX3c:-K8u8YCwj-6P6OcLU2cs:1739308068) ([raw file](https://github.com/cms-dev/cms/blob/dcb96d6c3936a0ee39f2cd4de708f68e546274a6/cmscontrib/ExportSubmissions.py#L139)):*
Just curious, why do you prefer the raw fields here instead of using the objects? Do we have performance issues?

---

*[cmscontrib/ExportSubmissions.py, line 160 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/513#-K7gvKosgFmeQcbgg_FM:-K8uAE154Kz9UQnDAgwD:793851708) ([raw file](https://github.com/cms-dev/cms/blob/dcb96d6c3936a0ee39f2cd4de708f68e546274a6/cmscontrib/ExportSubmissions.py#L160)):*
I find it quite ugly, but if there is no other succinct way it's fine

---

*[cmscontrib/ExportSubmissions.py, line 42 \[r2\]](https://reviewable.io:443/reviews/cms-dev/cms/513#-K8u7JNGBsovBKqkYNEH:-K8u7JNHvZS_QM7nqZvk:1765718262) ([raw file](https://github.com/cms-dev/cms/blob/6b2aca3bfa28025a1bb081e2fae34626e5b7e6b8/cmscontrib/ExportSubmissions.py#L42)):*
All uppercase; also two blank lines between toplevel stuff (even for variables, either you remove all whitelines, or add two).

---

*[cmscontrib/ExportSubmissions.py, line 55 \[r2\]](https://reviewable.io:443/reviews/cms-dev/cms/513#-K8u7d9ZD_a4gvQJgoi7:-K8u7d9_wmz8tHHMK0nM:1751194523) ([raw file](https://github.com/cms-dev/cms/blob/6b2aca3bfa28025a1bb081e2fae34626e5b7e6b8/cmscontrib/ExportSubmissions.py#L55)):*
I'd prefer to use %s

---

*[cmscontrib/ExportSubmissions.py, line 78 \[r2\]](https://reviewable.io:443/reviews/cms-dev/cms/513#-K8uA2JGizjKOUQ7c6jT:-K8u9nSj6BLFNd9IW38D:2055470720) ([raw file](https://github.com/cms-dev/cms/blob/6b2aca3bfa28025a1bb081e2fae34626e5b7e6b8/cmscontrib/ExportSubmissions.py#L78)):*
If you set [value] here later you can avoid the ```if unique``` (and in the previous line you should have ```or usertask[key][0] > value```

---

*[cmscontrib/ExportSubmissions.py, line 86 \[r2\]](https://reviewable.io:443/reviews/cms-dev/cms/513#-K8u9u1mAI9W2H-j1Oly:-K8u9u1mAI9W2H-j1Olz:355433174) ([raw file](https://github.com/cms-dev/cms/blob/6b2aca3bfa28025a1bb081e2fae34626e5b7e6b8/cmscontrib/ExportSubmissions.py#L86)):*
key, values, and later for value in values

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cms-dev/cms/513)*

wil93:
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[cmscontrib/ExportSubmissions.py, line 139 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/513#-K7gtWqJQJw0cixEzX3c:-K9oT8OnBiphfKNTnp-X:-1647949333) ([raw file](https://github.com/cms-dev/cms/blob/dcb96d6c3936a0ee39f2cd4de708f68e546274a6/cmscontrib/ExportSubmissions.py#L139)):*
I initially used objects but then I started testing it with many submissions (about 500) and noticed that even the first select was somewhat slow. I thought that could be partially caused by the numerous `.join()` statements, so I tried to restrict the selection as to reduce the data fetched from "500 x [a lot]" to "500 x [much less]".

Since that query was followed by many more queries (one to fetch each file) I decided that it should have been optimized. At the time I didn't really test exactly how much faster it was, but I recall it "feeling" faster.

I'm gonna try to test this...

...done, I just run this:

    explain analyze select * from submissions s, tasks t, files f, submission_results r, datasets d, participations p, users u where p.user_id = u.id and d.id = t.active_dataset_id and r.submission_id = s.id and f.submission_id = s.id and s.participation_id = p.id and s.task_id = t.id and t.contest_id = p.contest_id;

and this:

    explain analyze select s.id, s.language, s.timestamp, r.score, f.filename, f.digest, u.id, u.username, u.first_name, u.last_name, t.id, t.name from submissions s, tasks t, files f, submission_results r, datasets d, participations p, users u where p.user_id = u.id and d.id = t.active_dataset_id and r.submission_id = s.id and f.submission_id = s.id and s.participation_id = p.id and s.task_id = t.id and t.contest_id = p.contest_id;

and I consistently got ~230ms for the first and ~140ms for the second.

I'm not sure how meaningful this "raw" test is, but I'm confident that SQLAlchemy would yield similar results.

---

*[cmscontrib/ExportSubmissions.py, line 42 \[r2\]](https://reviewable.io:443/reviews/cms-dev/cms/513#-K8u7JNGBsovBKqkYNEH:-K8uCRBDA5qEU1mPPusr:-499146009) ([raw file](https://github.com/cms-dev/cms/blob/6b2aca3bfa28025a1bb081e2fae34626e5b7e6b8/cmscontrib/ExportSubmissions.py#L42)):*
Done.

---

*[cmscontrib/ExportSubmissions.py, line 55 \[r2\]](https://reviewable.io:443/reviews/cms-dev/cms/513#-K8u7d9ZD_a4gvQJgoi7:-K8uCk6RMBKyO06O4W4T:-499146009) ([raw file](https://github.com/cms-dev/cms/blob/6b2aca3bfa28025a1bb081e2fae34626e5b7e6b8/cmscontrib/ExportSubmissions.py#L55)):*
Done.

---

*[cmscontrib/ExportSubmissions.py, line 78 \[r2\]](https://reviewable.io:443/reviews/cms-dev/cms/513#-K8uA2JGizjKOUQ7c6jT:-K9oaftDShnpFJ4xRJ1h:-499146009) ([raw file](https://github.com/cms-dev/cms/blob/6b2aca3bfa28025a1bb081e2fae34626e5b7e6b8/cmscontrib/ExportSubmissions.py#L78)):*
Done.

---

*[cmscontrib/ExportSubmissions.py, line 86 \[r2\]](https://reviewable.io:443/reviews/cms-dev/cms/513#-K8u9u1mAI9W2H-j1Oly:-K9oT1l4ZIisN5L59keM:-499146009) ([raw file](https://github.com/cms-dev/cms/blob/6b2aca3bfa28025a1bb081e2fae34626e5b7e6b8/cmscontrib/ExportSubmissions.py#L86)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cms-dev/cms/513)*

cmscontrib/ExportSubmissions.py
```diff
@@ -0,0 +1,213 @@
...
+                                            time=s_timestamp, user=u_name)
+            filename = os.path.join(args.output_dir, filename)
+            if os.path.exists(filename):
+                logger.warning("Skipping file '%s' because it already exists",
+                               filename)
+
+            fso = FSObject.get_from_digest(f_digest, session)
+            assert fso is not None
+            with fso.get_lobject(mode="rb") as file_obj:
+                data = file_obj.read().decode('latin-1')
```
wafrelka: I want to use utf-8 instead of latin-1. Is there any room for encoding selection?

wil93: Done. For some reason I remember that `decode('latin-1')` was necessary, but I just tested it now with `decode('utf-8')` and it works anyway, so...

stefano-maggiolo:
Reviewed 1 of 1 files at r4.
Review status: 1 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[cmscontrib/ExportSubmissions.py, line 139 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/513#-K7gtWqJQJw0cixEzX3c:-KAW85NKqHygoaCYAGiI:1369561699) ([raw file](https://github.com/cms-dev/cms/blob/dcb96d6c3936a0ee39f2cd4de708f68e546274a6/cmscontrib/ExportSubmissions.py#L139)):*
I think the difference in the query time is the minimum, probably Alchemy adds a lot more than 100ms.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cms-dev/cms/513)*

stefano-maggiolo:
Merged.

PR #526, opened 2016-01-23
Simplified Chinese version of cms.po updated

vanshady:
We used to only have traditional Chinese, but now I translate it to simplified Chinese, which is the language used in mainland China.

[Review on Reviewable](https://reviewable.io/reviews/cms-dev/cms/526)


stefano-maggiolo:
Thank you, merged!

vanshady:
I didn't change the following lines and I didn't use any Python, where there be a problem?
"POT-Creation-Date: 2014-07-01 07:17+0100\n"
"PO-Revision-Date: 2014-07-01 07:24+0100\n"

stefano-maggiolo:
No problem for the dates (if you prefer, feel free to send a PR to change the revision, but it's not required). Not sure I understand what you mean with "I didn't use any python" (my guess: you did not recreate the pot file extracting the strings from the source code; if so, that's also fine as the strings haven't changed for a long time).

vanshady:
I see. Thanks for the reply.

PR #527, opened 2016-01-25
Fix bug in cmsAddTask when -c is specified

wil93:
[Review on Reviewable](https://reviewable.io/reviews/cms-dev/cms/527)


stefano-maggiolo:
As discussed offline, we implicitly assume that tasks' nums are from 0 to len(tasks) - 1. This is enforced by AWS, but sadly not by RemoveTask, which shall be changed. Adding an issue for it.

---

Reviewed 1 of 1 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cms-dev/cms/527#-:-K8uBLWgCY9nwmuuCuoY:63826371)*

stefano-maggiolo:
Merged. The issue is https://github.com/cms-dev/cms/issues/530

PR #533, opened 2016-02-06
Add an option to disable testing interface

wafrelka:
This is useful if admins want to prohibit server side testing (like IOI-style).

[Review on Reviewable](https://reviewable.io/reviews/cms-dev/cms/533)

### DIFF
--- a/cms/conf.py
+++ b/cms/conf.py
@@ -91,6 +91,7 @@ class Config(object):
         self.max_input_length = 5000000
         self.stl_path = "/usr/share/doc/stl-manual/html/"
         self.allow_questions = True
+        self.allow_testing = True
         # Prefix of 'iso-codes'[1] installation. It can be found out
         # using `pkg-config --variable=prefix iso-codes`, but it's
         # almost universally the same (i.e. '/usr') so it's hardly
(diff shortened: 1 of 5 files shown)


wil93:
This is a very good improvement. However, note that CMS is currently trying to "migrate" options from cms.conf to the database. I think that the `allow_testing` option should be implemented as a new boolean field in the `contests` table, without updating config/cms.conf.sample or cms/conf.py.

stefano-maggiolo:
Thanks for this, we had a few requests about it :)

Seconding William here, this is definitely a contest-level configuration. Would you like to move it there? A sample commit doing the same for two other configuration parameters is at https://github.com/cms-dev/cms/commit/d60c9876034309204e305b2a8ab5e8fc04e5c8bb
[Note that I just changed a bit the contest edit page, so you want to rebase over it, and you will have to change only template&handler for the edit contest, not for add contest which doesn't contain all the configurations anymore.]

stefano-maggiolo:
Ping :)

If you prefer, I'll follow up with an additional commit moving the configuration to the DB, just let me know if you want to do it yourself!

wafrelka:
I'm not familiar with the code around DB. I'd appreciate it if you could write additional commits.

stefano-maggiolo:
SG, no problem. Will do it soon!

stefano-maggiolo:
Done (merged yours and followed up with the move to DB). Thanks. Btw, similarly to Myungwoo, feel free to add your name to AUTHORS.txt if you want :)

stefano-maggiolo:
(or pseudonym)

wafrelka:
I'm afraid the `allow_testing` option still remains in `cms.conf.sample`.

stefano-maggiolo:
Wops, good catch. Thanks!

PR #535, opened 2016-02-10
Add missing 'return' keyword at function 'get_evaluation'

wafrelka:
Without 'return' keyword, the function always returns None ...

[Review on Reviewable](https://reviewable.io/reviews/cms-dev/cms/535)

### DIFF
--- a/cms/db/submission.py
+++ b/cms/db/submission.py
@@ -405,7 +405,7 @@ class SubmissionResult(Base):
         # and spare a query.
         # (We could use .one() and avoid a LIMIT but we would need to
         # catch a NoResultFound exception.)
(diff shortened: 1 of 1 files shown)


stefano-maggiolo:
Hi! wow, this is embarassing :)

Since this method has no callers (luckily!), can you remove it entirely instead? Unless you want to use it for something else in the future, in which case I'm happy to merge this. Let me know!

wafrelka:
We want to use this function to create some tools for our domestic OI contest.

stefano-maggiolo:
No problem, thanks! (merged)

wafrelka:
thank you!

PR #544, opened 2016-02-16
Fix user tests

wil93:
[Review on Reviewable](https://reviewable.io/reviews/cms-dev/cms/544)


stefano-maggiolo:
Any of you'd like to write a test for user tests?

stefano-maggiolo:
Unittests were failing because of a mistake in them :(

Hope you don't mind, I amended your commit to include the fix for the tests as well.

stefano-maggiolo:
Merged.

PR #546, opened 2016-02-17
Reduce memory usage in AWS

wafrelka:
If a task has many test cases, using `joinedload` with `SubmissionResult.evaluations` consumes a lot of unnecessary memory because `joinedload` creates a lot of duplicated columns.
This patch replaces some `joinedload`s with `subqueryload`s and reduces the memory usage of AWS significantly.

Although memory usage varies based on a contest configuration (the amount of problems / test cases / submissions etc...), in my environment this patch reduces the memory usage of AWS from 1.3GB to 130MB when opening submissions page of a contest.

[Review on Reviewable](https://reviewable.io/reviews/cms-dev/cms/546)

### DIFF
--- a/cms/server/admin/handlers/base.py
+++ b/cms/server/admin/handlers/base.py
@@ -40,7 +40,7 @@ from functools import wraps
 import tornado.web
 
 from sqlalchemy.exc import IntegrityError
-from sqlalchemy.orm import joinedload
+from sqlalchemy.orm import subqueryload
 
 from cms import __version__
 from cms.db import Admin, Contest, Participation, Question, \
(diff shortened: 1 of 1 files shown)


stefano-maggiolo:
Wow, super! I'm not sure I fully understand why this saves so much memory - is it because details such as the submission timestamp are stored for each evaluation in the original source? I expected this not being a problem... anyway this definitely improves things and it also decrease the time for the request!

If you see more of these please keep them coming! :D

---

Reviewed 1 of 1 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cms-dev/cms/546#-:-KB3tz0sQfSsN83V_t65:1890039891)*

stefano-maggiolo:
Merged!

wafrelka:
I don't know the detail of the difference between `subqueryload` and `joinedload`, but [SQLAlchemy Documentation](http://docs.sqlalchemy.org/en/latest/orm/loading_relationships.html#what-kind-of-loading-to-use) describes it.

> - When using joined loading, (...) Each row will also contain the full set of columns represented by the parents, repeated for each collection item - SQLAlchemy does not re-fetch these columns other than those of the primary key, however most DBAPIs (with some exceptions) will transmit the full data of each parent over the wire to the client connection in any case. Therefore joined eager loading only makes sense when the size of the collections are relatively small.
> - When using subquery loading, (...) An INNER JOIN is used, a minimum of parent columns are requested, only the primary keys. So a subquery load makes sense when the collections are larger.

wafrelka:
Thank you for merging.

PR #547, opened 2016-02-19
Fix: reject bad RPC auth request

wil93:
The header could not be defined (TypeError), could be non-integer (ValueError), and len(path_info) could be less than 3 (IndexError)

[Review on Reviewable](https://reviewable.io/reviews/cms-dev/cms/547)


stefano-maggiolo:
Reviewed 1 of 1 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cms-dev/cms/547)*

stefano-maggiolo:
Merged!

Issue #548, opened 2016-02-21
Wrong message for delayed user.


myungwoo:
It's the screen for 2 hours delayed user.

![default](https://cloud.githubusercontent.com/assets/9026019/13203895/4cf5e832-d905-11e5-9822-b6be96d59405.PNG)

Problem 1)
"Until contest ends" should be modified to "Until contest starts".

Problem 2)
It seems it needs to delay starting time and ending time in overview page either.

p.s)
I'm going to run APIO(Asia-Pacific Informatics Olympiad)2016 with CMS.
Each country may have different starting time because it's not a off-line competition.
I'm planning to handle this with delay time for user instead of running bunch of contests.
Is there any better idea to handle this?

stefano-maggiolo:
As mentioned in the PR, https://github.com/stefano-maggiolo/cms/commit/a1372aab6fd9a9d282c97158bd15dcfeb1d6ad0a should fix both. Please take a look!

myungwoo:
Okay. I didnt consider about usaco-like contest. Your commit looks perfect!

By the way,, I have two more questions..
1) When your commit get pulled to cms-dev/cms master branch?
2) Could you please tell me how to add users to contest or add teams in 1.3dev0?
I used to solve it by cmsAddUser script in v1.2 but the script has been changed so I have no idea.

stefano-maggiolo:
Hello again! The AddStuff scripts caused a lot of confusion, so we decided to change them: now everything that works with a loader (i.e., importers) is called ImportStuff, and everything that works by looking only at the arguments in the command line is called AddStuff. I also added AddUser, AddTeam and AddParticipation to make scripting easier.

As for when they'll be in master... they are now :)

It would be nice if you'd test these new scripts! Thanks.

PR #549, opened 2016-02-21
Fix countdown message for delayed user before contest starts.

myungwoo:
Fix problem 1 describing at #548 .

[Review on Reviewable](https://reviewable.io/reviews/cms-dev/cms/549)


stefano-maggiolo:
Hi Myungwoo, unfortunately it's not that simple, as the same message is used when the user has not started its time in a so-called "virtual contest" or "usaco-like contests" (i.e., when they can use x seconds within a window of y, but they need to start explicitly the time). I'm going to close this, but please take a look at https://github.com/stefano-maggiolo/cms/commit/a1372aab6fd9a9d282c97158bd15dcfeb1d6ad0a which I believe fixes both issues in #548

PR #556, opened 2016-02-24
Fix typo that broke killing services via AWS

### LINKED ISSUE #554: Cannot kill services in adminwebserver
When i clicked kill button to kill any services(e.g. proxy service, worker, etc),
refresh symbol appeared for a few seconds.
But the service is still running.

wil93:
Fixes #554

[Review on Reviewable](https://reviewable.io/reviews/cms-dev/cms/556)


bokse:
Now it works. thanks.

stefano-maggiolo:
Merged, thanks both for reporting and solving :)

PR #561, opened 2016-02-26
Set basic language to fallback & Move allow_questions from conf to db

### LINKED ISSUE #560: Fallback to 'en' even if 'en' is not in allowed_localizations
Fallback to 'en' even if 'en' is not in allowed_localizations when browser language is 'en'.

myungwoo:
Set basic language to fallback -> Fix #560 . First commit
Move allow_questions from conf to db . Second commit

Updater testing finished.

[Review on Reviewable](https://reviewable.io/reviews/cms-dev/cms/561)

### DIFF
--- a/cms/server/contest/handlers/base.py
+++ b/cms/server/contest/handlers/base.py
@@ -224,7 +224,9 @@ class BaseHandler(CommonRequestHandler):
         # TODO We fallback on basic_lang if no language matches: we could
         # return 406 Not Acceptable instead.
         # Select the one the user likes most.
-        basic_lang = 'en' if 'en' in lang_codes else lang_codes[0]
+        # TODO specify about basic_lang in docs.
+        basic_lang = lang_codes[0].replace("_", "-") \
+                    if len(self.contest.allowed_localizations) else 'en'
         http_langs = [lang_code.replace("_", "-") for lang_code in lang_codes]
         self.browser_lang = parse_accept_header(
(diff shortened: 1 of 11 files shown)


cms/server/contest/handlers/base.py
```diff
@@ -221,13 +221,14 @@ def get_user_locale(self):
             lang_codes = filter_language_codes(
                 lang_codes, self.contest.allowed_localizations)
 
-        # TODO We fallback on "en" if no language matches: we could
+        # TODO We fallback on basic_lang if no language matches: we could
         # return 406 Not Acceptable instead.
         # Select the one the user likes most.
+        basic_lang = 'en' if 'en' in lang_codes else lang_codes[0]
```
wil93: Mmm, what about just `basic_lang = lang_codes[0]`? I think it would be more consistent... It could also be specified in the documentation (e.g. "the first language in the JSON list will be used as a fallback")

myungwoo: It's better ;)
However, the order of lang_codes is diffirent with order of JSON list.
It's possible to make same. I'll do that next morning.

myungwoo: Now please check 1513eb6

stefano-maggiolo:
I meant *only* 4 spaces :)

Anyway, I merged the commits with just a few spacing changed. Thank you!

PR #562, opened 2016-02-27
Add option to avoid installing configuration files

wil93:
I think I forgot to open a PR for this minor change. It's useful when the configuration files will be provided externally (e.g. with docker run -v)

[Review on Reviewable](https://reviewable.io/reviews/cms-dev/cms/562)


stefano-maggiolo:
Merged.

PR #563, opened 2016-02-28
Remove login page, logout btn for ip_autologin

myungwoo:
Login page and logout button should be removed when ip_autologin is on, since ip_autologin restrict all actions for login & logout.

[Review on Reviewable](https://reviewable.io/reviews/cms-dev/cms/563)


wil93:

---

*[cms/server/contest/templates/base.html, line 69 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/563#-KBcNKY7vCsnw43Xe4Kf:-KBcNKY7vCsnw43Xe4Kg:331844467) ([raw file](https://github.com/cms-dev/cms/blob/33505a989e045e5eb0fe112e9177ffa68a5f2b9d/cms/server/contest/templates/base.html#L69)):*
Maybe this (and the `{% end %}` below) should be indented, because they're inside another `if`

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cms-dev/cms/563)*

stefano-maggiolo:
Actually, with ip_autologin (and/or ip_lock), we do not remove the possibility to log in with username and password, so we cannot remove all these.

PR #564, opened 2016-02-28
Add default message when checker gives empty message

myungwoo:
When checker gives empty stderr, translator prints po file information in non-English language.

[Review on Reviewable](https://reviewable.io/reviews/cms-dev/cms/564)

### DIFF
--- a/cms/grading/__init__.py
+++ b/cms/grading/__init__.py
@@ -8,6 +8,7 @@
 # Copyright © 2013 Bernard Blackham <bernard@largestprime.net>
 # Copyright © 2013-2014 Luca Wehrstedt <luca.wehrstedt@gmail.com>
 # Copyright © 2014 Fabian Gundlach <320pointsguy@gmail.com>
+# Copyright © 2016 Myungwoo Chun <mc.tamaki@gmail.com>
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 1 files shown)


stefano-maggiolo:
Merged!

stefano-maggiolo:
I'm sorry this too is not general enough to get in: depending on the scoretype, 0 and 1 might or might not mean failure and success, respectively. For example, see GroupThreshold, where the outcome represent "how much of a resource has been used" and so 0.0 is quite positive!

stefano-maggiolo:
Does https://github.com/stefano-maggiolo/cms/commit/0291079569118064641119c37b39f0fa454b58a1 solve the issue of showing the po file?

myungwoo:
Yep. Now empty message is printed correctly.
However, making checker print translatable message(e.g. Output is correct, Output isn't correct or something else) would be nice.
There's no method to handle this now.

myungwoo:
@stefano-maggiolo .
How about making translatable checker message like myungwoo@832d539 ?

When checker print "trans:correct" (without quotes) to stderr, then translatable "Output is correct" message will print.
Similarly, when checker print "trans:wrong" (without quotes) to stderr, then translatable "Output isn't correct" message will print.

Directly print "Output is correct", "Output isn't correct" is not translatable probably because of newline character end of the string.

stefano-maggiolo:
On one hand I'd like to have translated checker messages, on the other I see some problems.

Of course, we cannot be very granular with the messages we handle there, because we cannot anticipate all the possible messages a checker can print. And if we handle only "correct/not correct", then what's the difference with the status that the score type already prints (the one with the green background in your screenshot)?

myungwoo:
I agree with your opinion, there's no difference between score type prints & translated checker message.
However, I'm still thinking it's meaningful even if we just support "correct/not correct".

White-diff evaluation also just support "correct/not correct".
Problem author may want his/her checker acts like white-diff (just for checker message).
To be honest, I did...

Again, sorry about unskilled English.. :(
If you have problem to understand what I'm saying, please excuse me.

stefano-maggiolo:
No problems with English, your help is much more valuable than any trouble we'll have understanding each other :)

I think you convinced me, with the analogy with white_diff. What do you think of https://github.com/stefano-maggiolo/cms/commit/d6a9825085244ed4b6e7ec24d4f5c44fa4d27731 ?

If you need other messages let me know, we can add one for "partially correct", for example.

myungwoo:
Adding "partially correct" would be great!

PR #567, opened 2016-03-01
Allow manager to request string translations

myungwoo:
Discussion moved from #564 .
In addtion, I add partially correct message for checker.

[Review on Reviewable](https://reviewable.io/reviews/cms-dev/cms/567)


stefano-maggiolo:
Thanks, merged!

PR #568, opened 2016-03-02
Set maximum question subject length in html

myungwoo:
Maximum subject length is 50, not 250.
[communicaton.py](https://github.com/cms-dev/cms/blob/master/cms/server/contest/handlers/communication.py) line 72.

[Review on Reviewable](https://reviewable.io/reviews/cms-dev/cms/568)


stefano-maggiolo:
Thanks, merged!

PR #574, opened 2016-03-05
Fix bug which non-login user cannot switch language

myungwoo:
utils was not defined when uesr is not logged in.

my solution -> define util with dummy timestamp when user is not logged in.

[Review on Reviewable](https://reviewable.io/reviews/cms-dev/cms/574)


stefano-maggiolo:
Merged, thanks

PR #577, opened 2016-03-06
Update ajax redirect link to relative

myungwoo:
When aws is mapping by nginx prefix with proxy (like [/config/nginx.conf.sample](https://github.com/cms-dev/cms/blob/master/config/nginx.conf.sample): line 114-115), ajax redirect link should be changed to relative.
It is redirected by javascript(window.location.replace).

[Review on Reviewable](https://reviewable.io/reviews/cms-dev/cms/577)

### DIFF
--- a/cms/server/admin/handlers/admin.py
+++ b/cms/server/admin/handlers/admin.py
@@ -3,6 +3,7 @@
 
 # Contest Management System - http://cms-dev.github.io/
 # Copyright © 2015 Stefano Maggiolo <s.maggiolo@gmail.com>
+# Copyright © 2016 Myungwoo Chun <mc.tamaki@gmail.com>
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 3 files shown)


stefano-maggiolo:
Hi Myungwoo, is there a bug for which this change is necessary?

myungwoo:
Yes, there is.

If I use AdminWebServer with proxy such as http://example.com/aws/.
It redirects to http://example.com/admins because it redirects by javascript instead of Handler.redirect method.

So, when I delete announcement, attachment, statement, or admin 404 not found page appear, however deletion works well.

myungwoo:
Reminder! @stefano-maggiolo 

Please consider this PR.
Sorry for my unskilled English. It is my best words to describe the situation.

Again,
when I delete admin account at "http://example.com/aws/admin/1",
after the deletion it redirects to "http://example.com/admins",
even the correct redirect url is "http://example.com/aws/admins".

This PR updates ajax redirect to relative url to fix above situation.

stefano-maggiolo:
Hi Myungwoo! I'm really sorry for taking so long in going through the remaining PRs (I'm quite busy at work and also in my free time due to personal reasons), but rest assured I'm definitely considering your PRs :) 
As you might have noticed I did some progress this weekend, and I intend to continue during the week.

stefano-maggiolo:
(Finally) merged, thank you!

PR #578, opened 2016-03-06
Improve permission checking in AWS

myungwoo:
Add permission check for PERMISSION_ALL at SimpleHandler &
Remove User add menu from non-permission_all admin.

[Review on Reviewable](https://reviewable.io/reviews/cms-dev/cms/578)


stefano-maggiolo:
Merged

PR #583, opened 2016-03-11
Add --prefix to cmsImportTask

wil93:
This is useful when you want to add a task to a contest, but you have already added it to another contest.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/cms-dev/cms/583)

### DIFF
--- a/cmscontrib/ImportTask.py
+++ b/cmscontrib/ImportTask.py
@@ -60,8 +60,9 @@ class TaskImporter(BaseImporter):
 
     """
 
-    def __init__(self, path, update, no_statement, contest_id, loader_class):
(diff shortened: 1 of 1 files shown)


stefano-maggiolo:
Wouldn't it be more flexible to specify the whole name? What if you want a suffix instead?

wil93:
This way you can do just `cmsImportTask --prefix "something_" .` without having to specify the name (handy if you have to run the command on many tasks).

stefano-maggiolo:
Well, anyhow, this is for you so you know better...

stefano-maggiolo:
Merged.

wil93:
In retrospect, maybe it really would have been better to override the whole name :sweat_smile: see #616

Issue #582, opened 2016-03-11
Downloading tests with / in codename fails


akmohtashami:
In admin interface, downloading input/output of a test which its codename contains '/' character, fails with error "File request failed". This is because the url used to retrieve the file returns 404 error.

myungwoo:
What did you mean input/output of a test?
Did you mean input/output of a testcase, or input/output of a user test?

It seems codename of a testcase should be satisfied the following RE ([a-zA-Z0-9_.-]+).
You may see this RE in AWS URL matching([cms/server/admin/handlers/\__init\__.py](https://github.com/cms-dev/cms/blob/master/cms/server/admin/handlers/__init__.py)).

I think it should be improved with codename validation in adding testcase or adding testcases.

myungwoo:
How about commit myungwoo@0b6d94c?

akmohtashami:
I meant input/output of a testcase.
Commit myungwoo@0b6d94c prevents to add testcases with invalid codename in the admin interface. But I think this must be done at a lower level since admin interface isn't the only way of adding testcases. e.g. different loaders might create tests with invalid codenames.

stefano-maggiolo:
Maybe a better solution would be to url-encode the testcase codename in aws - I think that is sufficient to handle codenames with special chars. I will try to see if it is enough.

akmohtashami:
I just remembered that in output only tasks output_<codename> is used for the name of submitted files. if codenames contain characters such as `/` I think it might be a problem. So the admin might have to manually rename the data using bash, etc.

lerks:
Fixed in #738.

PR #584, opened 2016-03-12
Added testcase codename validator

akmohtashami:
For details, please refer to #582.

---

This change is [Review on Reviewable](https://reviewable.io/reviews/cms-dev/cms/584)

### DIFF
--- a/cms/db/task.py
+++ b/cms/db/task.py
@@ -35,12 +35,14 @@ from sqlalchemy.schema import Column, ForeignKey, CheckConstraint, \
     UniqueConstraint, ForeignKeyConstraint
 from sqlalchemy.types import Boolean, Integer, Float, String, Unicode, \
     Interval, Enum
-from sqlalchemy.orm import backref, relationship
+from sqlalchemy.orm import backref, relationship, validates
 from sqlalchemy.ext.orderinglist import ordering_list
 
 from . import Base, Contest
 from .smartmappedcollection import smart_mapped_collection, smc_sa10_workaround
 from cms import SCORE_MODE_MAX, SCORE_MODE_MAX_TOKENED_LAST
+
+import re
 
 
 class Task(Base):
@@ -555,3 +557,8 @@ class Testcase(Base):
     output = Column(
         String,
         nullable=False)
+
+    @validates("codename")
+    def validate_codename(self, key, codename):
+        codename = re.sub(r'[^a-zA-Z0-9_.-]+', '_', codename)
+        return codename


cms/db/task.py
```diff
@@ -555,3 +557,8 @@ class Testcase(Base):
     output = Column(
         String,
         nullable=False)
+
+    @validates("codename")
+    def validate_codename(self, key, codename):
```
stefano-maggiolo: Codacy Issue found: [Method could be a function](https://www.codacy.com/app/s-maggiolo/cms/file/2228608531/issues/source?bid=3167167&fileBranchId=3175304#l562)

lerks:
Hey, sorry for letting you wait so long to have a review (we're very short on manpower). Thanks for your contribution!

I like your idea of putting the validation at a low level. The way you went in this PR was what I had first in mind, but then I wondered if we could go any lower and in fact we can: we can have PostgreSQL enforce the constraint internally, so that the database is consistent even if accessed without SQLAlchemy. One way to do that is as follows:
```
codename = Column(
    Unicode,
    CheckConstraint("codename ~ '^[^a-zA-Z0-9_.-]+$'"),
    nullable=False)
```

Now, there's two more improvements I'd like to have. First, there are many fields whose alphabet should probably be restricted (filenames, usernames, task and contest codenames, and many more). It would be nice if we could have a decorator that put before the field achieves the same effect of putting that CheckConstraint (I'm not sure it's possible, but it's worth investigating).

Another thing I'm wondering about is what alphabet we want to allow. These codenames (of all sorts) appear in all kinds of places: file path components on disk, URL components, IDs of HTML tags, JSON dictionary keys, etc. I don't know exactly what constraints are enforced by each of these, but I'd rather err on the safe side. We should discuss this.

lerks:
Also, I am not sure how the violation of a CheckConstraint shows up in AWS but it would be nice to fail gracefully, so something along the lines of https://github.com/myungwoo/cms/commit/0b6d94c may be desirable as well (in addition to what I said before, not instead of it).

lerks:
I started working on this myself. Current progress is here: https://github.com/lerks/cms/tree/safe_codenames.

wil93:
> we can have PostgreSQL enforce the constraint internally, so that the database is consistent even if accessed without SQLAlchemy

By the way, not really related to this PR but: this could also be done for `default` attributes by using `server_default` instead. I think it would make sense to do it.

lerks:
Yes we could. I think originally the reason we chose not to do that was because we had the default password for Users be a randomly generated string (I don't know if this is still the case) and therefore for that we had to use the Python-enabled SQLAlchemy client-side defaults, and it would have been weird to mix the two.

lerks:
Done in #738.

PR #589, opened 2016-03-16
Allow auto-marking all testcases as public

wil93:
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/589)


stefano-maggiolo:
Merged

PR #592, opened 2016-03-20
Add escape around first and last names in RWS.

qnighy:
Since we (Japanese OI committee) use a registration form for online contest, we don't trust neither last names nor first names. This patch adds escapes around first and last names in RWS.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/592)

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -34,6 +34,10 @@ Copyright: 2008-2012 Dmitry Baranovskiy (http://raphaeljs.com)
 License: Expat
 Comment: See http://raphaeljs.com/license.html
 
+Files: cmsranking/static/lib/he.js
+Copyright: 2013-2015 Mathias Bynens <https://mathiasbynens.be/>
+License: Expat
+
 Files: cms/server/static/jq/jquery-1.7.1.min.js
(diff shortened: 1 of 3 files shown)


wil93:
Why not use the existing [`escape`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/escape) function?

wil93:
Mmm nevermind, escape is not useful here. However [this could be used](https://github.com/mathiasbynens/he) (demo [here](https://mothereff.in/html-entities))

qnighy:
I don't know which way is the best, but I also prepared the one you suggested. The former one is simpler, I think.

qnighy:
We should avoid building raw HTML and use jQuery in the end, but it's too much for this single issue.

stefano-maggiolo:
If I tracked back Hara-san's code correctly, that is used in mustache.js. I would prefer not to pull in another library for just 10 lines of JS, so I'll include the original version.

stefano-maggiolo:
Merged with two changes:
- removed last comma in the map definition (not valid JS);
- added escaping to task and contest names.

wil93:
If #459 is merged then we could easily use the external library

Issue #599, opened 2016-04-10
Default value for submission format


akmohtashami:
In the previous versions submission format of a task was automatically field with "<task_name>.%l" on task creation. However in the current version this feature has been removed. Was this intentional or it's OK to re-enable this feature again?

myungwoo:
I've sent PR to resolve this.

stefano-maggiolo:
Done! :)

PR #600, opened 2016-04-11
Set default submission format

myungwoo:
#599 suggests default value of submission format as previous version.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/600)


stefano-maggiolo:
Superfast! Thanks, merged

Fixes #599

Issue #591, opened 2016-03-18
Some text in CWS isn't translated


wafrelka:
Some text (for example, "Execution killed with signal %d (could be triggered by violating memory limits)" message in the submission detail page) isn't translated even if corresponding text seems to exist in `cms/locale/{lang}/LC_MESSAGES/cms.po`.
I tested in Japanese, Korean, Russian, Chinese and Italian language setting but in all cases the message "Execution killed with ..." in the submission detail page isn't translated into local language.
Replacement of %d in the message seems to be performed before translation. Is it the cause?

myungwoo:
Yeah. You are right.
There was a minor bug.
Thanks for reporting!

wafrelka:
The bug seems to be fixed. Thank you.

PR #601, opened 2016-04-11
Fix bug in submission message translation

myungwoo:
Fix bug which reported at #591 .

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/601)


stefano-maggiolo:
Merged!

Issue #61, opened 2012-10-14
Rethink phase management


lw:
A contest can be in three phases: pre-contest, contest and post-contest. CWS, in addition, defines a user-related actual_phase which takes the contest.per_user_time and user.starting_time into consideration for USACO-like contests. So we have the following "actual" phases:
- phase **-2**: the contest has not started yet
- phase **-1**: the contest is running but the user has not started its per_user_time yet
- phase **0**: the contest is running and the user can send submissions
- phase **+1**: the contest is running but the user has already finished its per_user_time
- phase **+2**: the contest has already finished

At the moment all phases except phase 0 are handled almost the same by CWS: the user can't do anything except reading the documentation, view announcements and messages and ask questions (and read answers). On phase -1 there's a small difference: the user can start his/hers per_user_time. 
On phase 0 the user can do almost anything: read statements, download attachments, send submissions and usertests, use tokens, etc. (but not start his/her per_user_time).

We may want to change this behavior (at least partially). For example we may want to give a user access to his/her submissions even on phases > 0, to allow to see the results of last-minute submissions which may get fully evaluated only after the contest ends (of course we would continue to block new submissions). Also, it could make sense to allow to use tokens even after the contest.
We should also decide if we want to implement the analysis mode properly, or if we think it's enough to do as we did at the IOI (extend the contest length and give infinite tokens).

This issue is the right place to discuss all of this.

PR #604, opened 2016-04-13
Access to submissions and resources after contest finishes

akmohtashami:
Users maintain their access to resources that they could access during the contest but are unable to make any changes(e.g. submit solutions/tests)
Currently they are not allowed to play their remaining tokens.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/604)

### DIFF
--- a/cms/server/contest/handlers/tasksubmission.py
+++ b/cms/server/contest/handlers/tasksubmission.py
@@ -369,7 +369,7 @@ class TaskSubmissionsHandler(BaseHandler):
 
     """
     @tornado.web.authenticated
-    @actual_phase_required(0)
+    @actual_phase_required(0, 1, 2)
     def get(self, task_name):
         participation = self.current_user
 
@@ -423,7 +423,7 @@ class SubmissionStatusHandler(BaseHandler):
     refresh_cookie = False
 
     @tornado.web.authenticated
-    @actual_phase_required(0)
+    @actual_phase_required(0, 1, 2)
     def get(self, task_name, submission_num):
         participation = self.current_user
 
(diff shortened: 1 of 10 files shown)


cms/server/contest/handlers/tasksubmission.py
```diff
@@ -579,6 +579,7 @@ class UseTokenHandler(BaseHandler):
     """Called when the user try to use a token on a submission.
 
     """
+    # TODO: Decide if tokens can be applied after contest ends.
```
wil93: Since tokens are sometimes used to determine the score, I'd say that it's reasonable to *not* allow users to apply tokens after the contest has ended.

Furthermore, when tokens are used to decide the score, playing a token is a something like sending a submission (i.e. if you play a token when it's too late, you won't get points, and that's how it's meant to be).

akmohtashami: The reason I mentioned this is because in #61 it was said that allowing this could make sense.

However I agree with you that letting users play tokens after the contest isn't a good idea. Usually we want the scores to be final when the contest ends.

stefano-maggiolo:
Sorry, contest data is intentionally not available after the contest finishes - even if maybe there isn't a real security reason, changing this behaviour seems a bit risky. Also in big onsite contests you want participants to vacate the hall asap and I think it helps if they don't have things to look at in the interface.

A while ago I discussed a proper analysis mode with @lucach - basically it means adding a new time window for analysis in the contest properties; if we are in that window (which might be represented as a new phase), contestants should be able to see full feedback (but not play token as you discussed). As a second step, it would be nice to allow contestants to submit in that window, but this requires quite some changes in the way we compute the score (since now we'd have to ignore all submissions outside of the valid time for phase 0 of the contest).

PR #606, opened 2016-04-18
Fix bad usage of translation function

myungwoo:
There's a minor bug.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/606)


stefano-maggiolo:
Thanks! On the positive news, I'm back :)

myungwoo:
Welcome back! I was waiting you.

PR #609, opened 2016-05-15
[cmsImportContest] Allow to import a contest without participations

lucach:
Users and contests are now not strictly tied together. It is reasonable to import a contest without participations (and add them later).

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/609)

### DIFF
--- a/cmscontrib/ImportContest.py
+++ b/cmscontrib/ImportContest.py
@@ -7,7 +7,7 @@
 # Copyright © 2010-2012 Matteo Boscariol <boscarim@hotmail.com>
 # Copyright © 2013 Luca Wehrstedt <luca.wehrstedt@gmail.com>
 # Copyright © 2014-2015 William Di Luigi <williamdiluigi@gmail.com>
-# Copyright © 2015 Luca Chiodini <luca@chiodini.org>
+# Copyright © 2015-2016 Luca Chiodini <luca@chiodini.org>
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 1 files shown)


stefano-maggiolo:
Reviewed 1 of 1 files at r1.

---

*[cmscontrib/ImportContest.py, line 161 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/609#-KIasDwYYi-sf6itveEo:-KIasDwZ5rq69XDv8t44:-553663968) ([raw file](https://github.com/cms-dev/cms/blob/5f9a5ecef579b2e6906e7e8b24b30146b2c74a07/cmscontrib/ImportContest.py#L161)):*
> ```Python
> 
>             # Check needed participations
>             if participations is not None:
> ```

To avoid an indentation, what about
```python
participations = participations if participations is not None else []
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/609)*

lucach:

---

*[cmscontrib/ImportContest.py, line 161 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/609#-KIasDwYYi-sf6itveEo:-KIiK5eIwhfDaAPqdNml:965693301) ([raw file](https://github.com/cms-dev/cms/blob/5f9a5ecef579b2e6906e7e8b24b30146b2c74a07/cmscontrib/ImportContest.py#L161)):*
> To avoid an indentation, what about
> ```python
> participations = participations if participations is not None else []
> ```
</details>
The method is already very nested: I agree on avoiding indenting more.
The one-liner you have proposed is longer than 80 cols (and wrapping it decreases readability), what about 

```
if participations is None:
    participations = []
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/609)*

stefano-maggiolo:
Merged with your proposed change, sorry for the delay!

PR #610, opened 2016-05-20
Send pending submission to RWS

myungwoo:
I'm considering the CMS for ICPC.

It is the first step for it.
Almost all ICPC scoreboard needs to display pending(not evaluated yet) submission also.
Submission is considered as pending in RWS when it has no subchange(score data) but submission.

+)
My idea for CMS - ICPC is, the submission is accepted when it got full score,
no special setting in CMS (new scoring method, penalty etc...),
just send RWS data to already made ICPC ScoreBoard([spotboard](https://github.com/wookayin/spotboard), [spotboard feedserver for cms](https://github.com/myungwoo/spotboard-cms)).

---

This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/610)

PR #612, opened 2016-05-21
Simple submission detail (ICPC Mode)

myungwoo:
One different between IOI and ICPC is the submission detail.
1) ICPC only shows the detail of first incorrect testcase, but IOI shows detail of all testcases.
2) In ICPC, there are only Accepted & Rejected (Correct or Incorrect), but in IOI there is score for the submission.

In this pull request,
I made **Simple Submission Detail mode** for a contest (I don't call it ICPC mode).

1) It is just available for grouping score types (GroupMin, GroupMul, but not sure for GroupThresold)
2) It only shows the detail of first incorrect(which is not "Correct") testcase, when I got full score there's no detail of any testcases.
3) It changed "score" column to "result" column in submissions table.
3-1) score-class remain same (red / yellow / green cell)
3-2) print message "Accepted" if submission got full-score.
3-3) print message "Rejected" otherwise

Finally, this pull request & #610 make CMS all ready to use for ICPC!

Please comment if there is any idea to handle this.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/612)

### DIFF
--- a/cms/grading/ScoreType.py
+++ b/cms/grading/ScoreType.py
@@ -73,7 +73,8 @@ class ScoreType(object):
         self.max_score, self.max_public_score, self.ranking_headers = \
             self.max_scores()
 
-    def get_html_details(self, score_details, translator=None):
+    def get_html_details(self, score_details, simple_submission_detail, \
+            translator=None):
         """Return an HTML string representing the score details of a
         submission.
 
@@ -95,6 +96,9 @@ class ScoreType(object):
                          "Try invalidating scores.")
             return translator("Score details temporarily unavailable.")
         else:
+            if simple_submission_detail:
+                return Template(self.SIMPLE_TEMPLATE).generate(
+                    details=score_details, _=translator)
             return Template(self.TEMPLATE).generate(details=score_details,
                                                     _=translator)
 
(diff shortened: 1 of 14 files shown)


cms/grading/ScoreType.py
```diff
@@ -95,6 +96,9 @@ def get_html_details(self, score_details, translator=None):
                          "Try invalidating scores.")
             return translator("Score details temporarily unavailable.")
         else:
+            if simple_submission_detail:
```
stefano-maggiolo: Codacy Issue found: [Line too long (85/80)](https://www.codacy.com/app/s-maggiolo/cms/file/2611551932/issues/source?bid=3167167&fileBranchId=3310990#l100)

cms/server/contest/handlers/tasksubmission.py
```diff
@@ -466,11 +466,24 @@ def get(self, task_name, submission_num):
...
-                data["max_public_score"] = "%g" % \
-                    round(score_type.max_public_score, task.score_precision)
-            data["public_score"] = "%g" % \
-                round(sr.public_score, task.score_precision)
+            if submission.task.contest.simple_submission_detail and \
+                score_type.max_public_score == score_type.max_score:
+                if sr.public_score == score_type.max_public_score:
+                    data["public_message"] = self._("Accepted")
+                    data["score_class"] = "score_100"
+                else if sr.public_score > 0:
```
stefano-maggiolo: Codacy Issue found: [invalid syntax](https://www.codacy.com/app/s-maggiolo/cms/file/2611551984/issues/source?bid=3167167&fileBranchId=3310990#l474)

cms/grading/ScoreType.py
myungwoo: Done

cms/server/contest/handlers/tasksubmission.py
myungwoo: It was a terrible mistake, sorry, fixed it.

lerks:
Hello, and thanks for your effort into making CMS suitable for ACM-ICPC competition! It's not something we've been actively seeking, but it's a welcome extra feature. Please keep us posted about your progress! (BTW, I'm also looking at #610, expect a comment there soon).

I haven't looked in detail into the code you've posted, but from reading the description I'm wondering whether the same result could be achieved simply with a custom ScoreType. In fact, it's the ScoreType that decides the score and how much information about the testcases is shown to the users. Could you check if that is enough to accomplish your goals? (If it does, we may be interested in including it in the main repository)

lerks:
I've looked at the code more carefully. As it is I cannot accept it, it would need some cleanup. Yet, I confirm that writing a custom ScoreType would achieve most of what you're looking for. The only exception is showing "Accepted" or "Rejected" instead of the numerical score. I will think whether there's a simple way to achieve it.

myungwoo:
Thank you for suggesting a new idea. Making new ScoreType seems much better than I did. I'll implement it soon.

Which method to handle showing "Accepted" or "Rejected" instead of the numerical score is the best do you think?

stefano-maggiolo:
To show strings, I think you just need to define the template you want in the scoretype, marking the translatable strings as in https://github.com/cms-dev/cms/blob/master/cms/grading/scoretypes/Sum.py#L42

lerks:
No, we're talking about a different thing. A ScoreType can provide custom HTML for the details (and can thus put its own custom text for testcase and subtask scores), but the score for the whole submission is rendered by CMS in a standard way, see [1]. That's the text that we want to override here, showing `"Accepted" if score == 100 else "Rejected"`.

[1] https://github.com/cms-dev/cms/blob/master/cms/server/contest/templates/submission_row.html#L63-85

myungwoo:
Thanks, lerks.
And also [1] too.
[1] is for ajax submission status updating.

[1] https://github.com/myungwoo/cms/blob/eaccdd94a2e55e22116f7e000876ebc2deae3947/cms/server/contest/handlers/tasksubmission.py#L469-L479

lerks:
@myungwoo, I implemented the required change in https://github.com/cms-dev/cms/pull/622. Could you have a look at it?

A sample ScoreType for ACM-ICPC-style contest could be the following (I didn't test it, so there may be mistakes). It doesn't show any information about testcases, just a global "accepted" or "rejected" verdict.
```
class ICPCScoreType(ScoreTypeAlone):
    def format_score(self, score, max_score, unused1, unused2, _=None):
        if score == max_score:
            return _("Accepted")
        else:
            return _("Rejected")

    def max_scores(self):
        return 1.0, 1.0, []

    def compute_score(self, sr):
        if sr.evaluated() and \
                all(float(ev.outcome) == 1.0 for ev in sr.evaluations):
            return 1.0, "{}", 1.0, "{}", []
        else:
            return 0.0, "{}", 0.0, "{}", []
```

lerks:
#622 was pushed.

PR #614, opened 2016-05-27
Support JSON5 for the cms.conf and cms.ranking.conf files

wil93:
There was some discussion some time ago about moving to YAML, but the change was regarded as too drastic. This commit allows to use a better version of JSON (which allows comments, commas at the end of arrays/objects, and multi-line strings) while still keeping backwards compatibility.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/614)

### DIFF
--- a/cms/conf.py
+++ b/cms/conf.py
@@ -28,7 +28,10 @@ from __future__ import unicode_literals
 
 import errno
 import io
-import json
+try:
+    import json5 as json
+except ImportError:
+    import json
 import logging
 import os
 import sys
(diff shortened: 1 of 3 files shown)


lerks:
I don't really support this. One reason is that the Python lib for JSON5 isn't available on Ubuntu (I know that you'll answer that it's only an optional dependency). I also have some reservations about using a non-standard format.

I'd rather use YAML (but I'm not gonna start that conversation again), or maybe something like JSmin [1], which BTW is available in Ubuntu. Or just keep it as it is.

[1] https://pypi.python.org/pypi/jsmin

stefano-maggiolo:
Closing this, sorry.

PR #615, opened 2016-05-29
Support using IP lock behind more than one proxy

wil93:
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/615)


stefano-maggiolo:
I'm merging it, with a small modification to make sure that num_proxy_used is preferred to is_proxy_used even when the former is set to 0.

stefano-maggiolo:
Done

PR #620, opened 2016-06-14
Add new GroupSum scoretype

wil93:
I can't test this right now, but it should be correct... I only changed `*` to `+` from `GroupMul.py`.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/620)


stefano-maggiolo:
I don't think there is enough demand for this :>

PR #627, opened 2016-06-24
ExportSubmissions: Avoid utf-8 decoding when --add-info is not given

wafrelka:
With this patch, we can avoid utf-8 conversion errors when a submission includes invalid chars. (It can happen when a contestant writes some comments in source code with an encoding other than utf-8)
Additionally, we can save time of unnecessary utf-8 conversion.

Maybe it is good to deal with utf-8 conversion error when --add-info is given, but I don't know how to.
(Ignoring invalid chars using `codecs.decode` function with `errors=ignore`, or concatenating utf-8 encoded additional info and source code as byte sequences?)

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/627)

### DIFF
--- a/cmscontrib/ExportSubmissions.py
+++ b/cmscontrib/ExportSubmissions.py
@@ -193,9 +193,9 @@ def main():
             fso = FSObject.get_from_digest(f_digest, session)
             assert fso is not None
             with fso.get_lobject(mode="rb") as file_obj:
-                data = file_obj.read().decode('utf-8')
 
                 if args.add_info:
+                    raw_data = file_obj.read().decode('utf-8')
                     data = TEMPLATE[s_language] % (
                         u_name,
                         u_fname,
@@ -203,9 +203,13 @@ def main():
                         t_name,
                         sr_score,
                         s_timestamp
-                    ) + data
+                    ) + raw_data
+                    out_options = {"encoding": "utf-8"}
+                else:
+                    data = file_obj.read()
+                    out_options = {}
 
-                with codecs.open(filename, "w", encoding="utf-8") as file_out:
+                with codecs.open(filename, "w", **out_options) as file_out:
                     file_out.write(data)
 
             done += 1


wil93:
I'm not sure about this workaround...

I think the real issue here is that this line:

```py
with fso.get_lobject(mode="rb") as file_obj:
    data = file_obj.read()
```

returns raw bytes, and we don't know which encoding those bytes are following.

So, we should either try some different encodings (in the vast majority of cases one of 'utf-8' or 'latin-1' will be the correct one but, in general, it depends on the user... e.g. which editor was used to write the code), *or*, we could enforce all large object to be stored as 'utf-8'.

For the time being I think the best solution is to modify the ExportSubmissions script so that it tries both 'utf-8' and 'latin-1', which are the most common encodings, I will open a PR in a moment.

wafrelka:
I want to export user submissions as-is, because I think integrity of submissions is important, and don't want to deal with encoding problems.

There's no problem if you want to deal with encoding problems when `--add-info` is given. #628 looks good to some extent. However, I don't like bringing the problem where it can be avoided (like non-`--add-info` case).

If you don't like the deference of encoding methods depending on the `--add-info` option, it may be good to introduce `--raw-output` flag or something like it to express the deference explicitly.

wil93:
> I think integrity of submissions is important

I'm not sure I understand what you mean by that... I think it makes sense to require a specific encoding (or maybe "one of several" accepted encodings). How would you display text data in the admin page without knowing the encoding? At some point we need to know the encoding...

> and don't want to deal with encoding problems

Neither do I, and that's why I suggest to only ever use utf-8 in all cases :smile: 

I think that exporting a bunch of files with several different encodings is more error-prone than exporting a bunch of utf-8 encoded files (what if you want to feed the exported files to another program? that program will need to handle a lot of encodings, instead of one).

> If you don't like the deference of encoding methods depending on the --add-info option, it may be good to introduce --raw-output flag or something like it to express the deference explicitly.

Mmm... Yes, that would be far cleaner. However, I still think it makes sense to change ContestWebServer so that we always store the source files in utf-8.

wil93:
@wafrelka I edited PR #628, now there is a `--utf8` flag available which is the opposite of the `--raw-output` flag you suggested (it is optional, but it automatically becomes mandatory when `--add-info` is specified). I tested it and it seems to work. What do you think?

wafrelka:
Personally, I think CMS should be just a pipe between contestants and core judging system including compilers, managers and output checkers. It shouldn't handle encoding problems at all so that users and compilers can use any encodings they want. Converting and modifying submissions may result in the difference of judging result between CMS and contestants side.
If we want to show user submissions in the admin page, we should guess encodings and convert submissions at the rendering of utf-8 encoded html page.

The word "integrity" I used means that we can always reproduce the same result at a local side with exported submission files as CMS produces at the server side. Converting submissions may break it and that's why I don't want to avoid utf-8 encoding.

wafrelka:
Now PR #628 seems good to me.

wafrelka:
…… %s/deference/difference/g

wafrelka:
PR #628 now includes the feature of this patch and I close this PR.

stefano-maggiolo:
Thank you for the bugreport and the contribution to the other PR!

PR #628, opened 2016-06-27
When exporting a submission file, try to guess the encoding if utf-8 fails

wil93:
Attempt to fix the issue described in #627

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/628)

### DIFF
--- a/cms/util.py
+++ b/cms/util.py
@@ -73,7 +73,7 @@ def utf8_decoder(value):
     """Decode given binary to text (if it isn't already) using UTF8, and
     falling back to other encodings when possible (using chardet to guess).
 
-    value (bytes): value to decode.
+    value (string): value to decode.
 
     return (unicode): decoded value.
 
--- a/cms/util.py
+++ b/cms/util.py
@@ -86,9 +86,12 @@ def utf8_decoder(value):
         try:
             return value.decode("utf-8")
         except UnicodeDecodeError:
-            return value.decode(chardet.detect(value)["encoding"])
-    else:
-        raise TypeError("Not a string.")
+            try:
+                return value.decode(chardet.detect(value).get("encoding"))
+            except TypeError:
+                pass
+
+    raise TypeError("Not a string.")
 
 
 class Address(namedtuple("Address", "ip port")):
--- a/cms/util.py
+++ b/cms/util.py
@@ -84,7 +84,7 @@ def utf8_decoder(value):
         return value
     elif isinstance(value, six.binary_type):
         try:
-            return value.decode('utf-8')
+            return value.decode("utf-8")
         except UnicodeDecodeError:
             return value.decode(chardet.detect(value)["encoding"])
     else:
--- a/cms/util.py
+++ b/cms/util.py
@@ -24,6 +24,7 @@ from __future__ import absolute_import
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import chardet
 import errno
 import logging
 import netifaces
@@ -69,9 +70,10 @@ def mkdir(path):
 
 
 def utf8_decoder(value):
-    """Decode given binary to text (if it isn't already) using UTF8.
+    """Decode given binary to text (if it isn't already) using UTF8, and
+    falling back to other encodings when possible (using chardet to guess).
 
-    value (string): value to decode.
+    value (bytes): value to decode.
 
     return (unicode): decoded value.
 
@@ -81,7 +83,10 @@ def utf8_decoder(value):
     if isinstance(value, six.text_type):
         return value
     elif isinstance(value, six.binary_type):
-        return value.decode('utf-8')
+        try:
+            return value.decode('utf-8')
+        except UnicodeDecodeError:
+            return value.decode(chardet.detect(value)["encoding"])
     else:
         raise TypeError("Not a string.")
 
--- a/cmscontrib/ExportSubmissions.py
+++ b/cmscontrib/ExportSubmissions.py
@@ -217,12 +217,12 @@ def main():
                         ) + data
 
                     # Print utf8-encoded, possibly altered data
-                    with codecs.open(filename, "w", encoding="utf-8") as file_out:
-                        file_out.write(data)
+                    with codecs.open(filename, "w", encoding="utf-8") as f_out:
+                        f_out.write(data)
                 else:
                     # Print raw, untouched binary data
-                    with open(filename, "wb") as file_out:
-                        file_out.write(data)
+                    with open(filename, "wb") as f_out:
+                        f_out.write(data)
 
             done += 1
             print(done, "/", len(results))
--- a/cmscontrib/ExportSubmissions.py
+++ b/cmscontrib/ExportSubmissions.py
@@ -204,7 +204,13 @@ def main():
                 data = file_obj.read()
 
                 if args.utf8:
-                    data = utf8_decoder(data)
+                    try:
+                        data = utf8_decoder(data)
+                    except TypeError:
+                        logger.critical("Could not guess encoding of file "
+                                        "'%s'. Aborting.",
+                                        filename)
+                        sys.exit(1)
 
                     if args.add_info:
                         data = TEMPLATE[s_language] % (
--- a/cmscontrib/ExportSubmissions.py
+++ b/cmscontrib/ExportSubmissions.py
@@ -193,11 +193,7 @@ def main():
             fso = FSObject.get_from_digest(f_digest, session)
             assert fso is not None
             with fso.get_lobject(mode="rb") as file_obj:
-                data = file_obj.read()
-                try:
-                    data = data.decode('utf-8')
-                except UnicodeDecodeError:
-                    data = data.decode('latin-1')
+                data = utf8_decoder(file_obj.read())
 
                 if args.add_info:
                     data = TEMPLATE[s_language] % (
--- a/cmscontrib/ExportSubmissions.py
+++ b/cmscontrib/ExportSubmissions.py
@@ -105,6 +105,9 @@ def main():
                         help="id of user (default: all users)")
     parser.add_argument("-s", "--submission-id", action="store", type=int,
                         help="id of submission (default: all submissions)")
+    parser.add_argument("--utf8", action="store_true",
+                        help="if set, the files will be encoded in utf8"
+                             " when possible")
     parser.add_argument("--add-info", action="store_true",
                         help="if set, information on the submission will"
                              " be added in the first lines of each file")
@@ -128,6 +131,11 @@ def main():
                             " exported for each (user, task)")
 
     args = parser.parse_args()
+
+    if args.add_info and not args.utf8:
+        logger.critical("If --add-info is specified, then --utf8 must be"
+                        " specified as well.")
+        sys.exit(1)
 
     if not os.path.exists(args.output_dir):
         os.mkdir(args.output_dir)
@@ -193,20 +201,28 @@ def main():
             fso = FSObject.get_from_digest(f_digest, session)
             assert fso is not None
             with fso.get_lobject(mode="rb") as file_obj:
-                data = utf8_decoder(file_obj.read())
+                data = file_obj.read()
 
-                if args.add_info:
-                    data = TEMPLATE[s_language] % (
-                        u_name,
-                        u_fname,
-                        u_lname,
-                        t_name,
-                        sr_score,
-                        s_timestamp
-                    ) + data
+                if args.utf8:
+                    data = utf8_decoder(data)
 
-                with codecs.open(filename, "w", encoding="utf-8") as file_out:
-                    file_out.write(data)
+                    if args.add_info:
+                        data = TEMPLATE[s_language] % (
+                            u_name,
+                            u_fname,
+                            u_lname,
+                            t_name,
+                            sr_score,
+                            s_timestamp
+                        ) + data
+
+                    # Print utf8-encoded, possibly altered data
+                    with codecs.open(filename, "w", encoding="utf-8") as file_out:
+                        file_out.write(data)
+                else:
+                    # Print raw, untouched binary data
+                    with open(filename, "wb") as file_out:
+                        file_out.write(data)
 
             done += 1
             print(done, "/", len(results))
--- a/cmscontrib/ExportSubmissions.py
+++ b/cmscontrib/ExportSubmissions.py
@@ -193,7 +193,11 @@ def main():
             fso = FSObject.get_from_digest(f_digest, session)
             assert fso is not None
             with fso.get_lobject(mode="rb") as file_obj:
-                data = file_obj.read().decode('utf-8')
+                data = file_obj.read()
+                try:
+                    data = data.decode('utf-8')
+                except UnicodeDecodeError:
+                    data = data.decode('latin-1')
 
                 if args.add_info:
                     data = TEMPLATE[s_language] % (
--- a/docs/Installation.rst
+++ b/docs/Installation.rst
@@ -214,7 +214,7 @@ To install CMS python dependencies on Ubuntu, you can issue:
          python-sqlalchemy python-psutil python-netifaces python-crypto \
          python-tz python-six python-beautifulsoup python-mechanize \
          python-coverage python-mock python-requests python-werkzeug \
-         python-gevent python-bcrypt patool
+         python-gevent python-bcrypt python-chardet patool
 
     # Optional.
     # sudo apt-get install python-yaml python-sphinx python-cups python-pypdf2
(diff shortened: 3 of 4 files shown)


qnighy:
I prefer saving the raw binary data to detecting text encodings. I don't think we can safely distinguish between latin1 and ms932 (or, almost equivalently, Shift JIS) or other encodings.

wil93:
Yeah, and exporting raw binary data has also the advantage of maintaining the original encoding, but the problem is when we use `--add-info`, which in some cases can be useful.

Adding information to the file means adding some bytes to it. These new bytes should be encoded with the same encoding used by the file, but this is not always possible. That's why I think the best think is to just export utf-8-encoded source files (even if this means forgetting what was the original encoding of the file), or even better: start storing source files directly in utf-8 (by modifying the contest interface / server).

So the best solution, IMHO, would be to store source files using utf-8. However, since for this issue we have to support source files already stored using other encodings, we need to either guess the encoding (and maybe we can do that with [chardet](https://pypi.python.org/pypi/chardet)) or stop supporting `--add-info`.

qnighy:
I see now and agree with the problem with encodings and ``--add-info``. Nevertheless I think #627 should be considered more.

#627 seems to deal with the problem by the following way:

- For ``--add-info`` cases, assume the sources are utf-8 encoded. This is the current behavior.
- For non-``--add-info`` cases, store the sources as-is. This is the behavior introduced in the patch.

If all sources are utf-8, then there is no problem. If there are non-utf-8 sources, non-``--add-info`` works safely. ``--add-info`` may fail, but this is necessary drawback. These behaviors seem to try to maximize both safety and convenience.

Or, for ``--add-info`` cases, we can use encoding detection as you say. Of course there may be errors in encoding detection, but this is unavoidable, unlike in non-``--add-info`` cases.

(By the way, there is another (not so attractive) choice: store the additional information in utf-8, and the sources as-is (just concatenating utf-8 strings and binaries). I see no advantages with it though.)

wil93:
I don't know if it's a good idea to save the output files in a different encoding depending on whether some additional information was added or not... it would be simpler to just assume we're *always* working with utf-8 instead of binary data (but we can't assume that, so we should try to *convert* everything to utf-8).

> there is another (not so attractive) choice: store the additional information in utf-8, and the sources as-is (just concatenating utf-8 strings and binaries).

The binary data is encoded in some way, so the outcome would be a concatenation of utf-8 and some other encoding. I don't know if the result would be anything valid...

wil93:
Oh, another reason for which we might want to enforce a specific encoding (utf-8) for all source files in the long run is this: sometimes we want to show the source code inside the browser without having to download it (e.g. AdminWebServer). That's not possible if we don't know the encoding of the file.

stefano-maggiolo:
I looked at this patch instead of #627 because it looked more complete, but I agree with @wafrelka that being a dumb pipe as much as possible is the preferred way (removing add-info would also be an option...).

@qnighy, @wafrelka, I think this patch is now ok, but I'd like your confirmation as we are much less likely to see encoding problems. Thanks!

---

Reviewed 2 of 4 files at r2, 2 of 2 files at r4.

---

*[cms/util.py, line 76 \[r4\]](https://reviewable.io:443/reviews/cms-dev/cms/628#-KLWcs8tKQtZmVQ1IlKF:-KLWcs8tKQtZmVQ1IlKG:460504132) ([raw file](https://github.com/cms-dev/cms/blob/fc11c870ab3697ddd7b73d8336ed66c017d988b6/cms/util.py#L76)):*
> ```Python
>     falling back to other encodings when possible (using chardet to guess).
> 
>     value (bytes): value to decode.
> ```

string in pydocchecker is an alias for basestring, which covers both bytes and unicode. Since this function accepts both, keep "string"

---

*[cmscontrib/ExportSubmissions.py, line 113 \[r4\]](https://reviewable.io:443/reviews/cms-dev/cms/628#-KLWdbdkHnTmQ9z1xvms:-KLWdbdkHnTmQ9z1xvmt:-746691383) ([raw file](https://github.com/cms-dev/cms/blob/fc11c870ab3697ddd7b73d8336ed66c017d988b6/cmscontrib/ExportSubmissions.py#L113)):*
> ```Python
>     parser.add_argument("--add-info", action="store_true",
>                         help="if set, information on the submission will"
>                              " be added in the first lines of each file")
> ```

Please add something like "this might create encoding problems"

---

*[cmscontrib/ExportSubmissions.py, line 138 \[r4\]](https://reviewable.io:443/reviews/cms-dev/cms/628#-KLWdiWK8DbA6aJZR2oY:-KLWdiWK8DbA6aJZR2oZ:-1065543403) ([raw file](https://github.com/cms-dev/cms/blob/fc11c870ab3697ddd7b73d8336ed66c017d988b6/cmscontrib/ExportSubmissions.py#L138)):*
> ```Python
>         logger.critical("If --add-info is specified, then --utf8 must be"
>                         " specified as well.")
>         sys.exit(1)
> ```

No need to block the user - just log something  like "--add-info will write additional information within the file, forcing a conversion of the whole file to UTF8. This might cause encoding problems", and flip args.utf8.

---

*[cmscontrib/ExportSubmissions.py, line 209 \[r4\]](https://reviewable.io:443/reviews/cms-dev/cms/628#-KLWeWFSmjHqw2YgY6tZ:-KLWeWFTFFhUyWnM4PVj:1596599206) ([raw file](https://github.com/cms-dev/cms/blob/fc11c870ab3697ddd7b73d8336ed66c017d988b6/cmscontrib/ExportSubmissions.py#L209)):*
> ```Python
>                     data = utf8_decoder(data)
> 
>                     if args.add_info:
> ```

I think it is more clear if you move this outside the if args.utf8, and repeat it at the end, closer to the else:

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/628#-:-KLWeifyBp0qSK4RpEQ1:-255869732)*

qnighy:
Now the patch seems good to me, in the sense of encodings. Thank you!

According to a comment on #627, @wafrelka too seems to be OK with this.

wafrelka:
Looks good to me.

wil93:

---

*[cms/util.py, line 76 \[r4\]](https://reviewable.io:443/reviews/cms-dev/cms/628#-KLWcs8tKQtZmVQ1IlKF:-KLXGByzzFx1reqIfULA:-499146009) ([raw file](https://github.com/cms-dev/cms/blob/fc11c870ab3697ddd7b73d8336ed66c017d988b6/cms/util.py#L76)):*
> string in pydocchecker is an alias for basestring, which covers both bytes and unicode. Since this function accepts both, keep "string"
</details>
Done.

---

*[cmscontrib/ExportSubmissions.py, line 113 \[r4\]](https://reviewable.io:443/reviews/cms-dev/cms/628#-KLWdbdkHnTmQ9z1xvms:-KLXGNskNGbx8N-yWwJl:994475001) ([raw file](https://github.com/cms-dev/cms/blob/fc11c870ab3697ddd7b73d8336ed66c017d988b6/cmscontrib/ExportSubmissions.py#L113)):*
> Please add something like "this might create encoding problems"
</details>
Mmm... I think it's not necessary, because if you try to run `cmsExportSubmission --add-info (..)` you will get an error saying that you need to also add `--utf8`, which already says "when possible" so the user is aware of the possibility of errors

---

*[cmscontrib/ExportSubmissions.py, line 138 \[r4\]](https://reviewable.io:443/reviews/cms-dev/cms/628#-KLWdiWK8DbA6aJZR2oY:-KLXH9HJAHERMzjL3V_S:-2102853973) ([raw file](https://github.com/cms-dev/cms/blob/fc11c870ab3697ddd7b73d8336ed66c017d988b6/cmscontrib/ExportSubmissions.py#L138)):*
> No need to block the user - just log something  like "--add-info will write additional information within the file, forcing a conversion of the whole file to UTF8. This might cause encoding problems", and flip args.utf8.
</details>
I'm not sure about this...

1. Explicit is better than implicit: I think it's better to explicitly force the user to know what's happening, and it's very easy to add a flag.
2. In general we may want to have a `--output-encoding=<encoding>` flag (it's not really needed since utf8 is enough, but maybe in the future) and, in that case, it's not reasonable to assume `--output-encoding=utf8` for `--add-info`; I think we should say "Error: you must specify some output encoding with `--output-encoding`".

---

*[cmscontrib/ExportSubmissions.py, line 209 \[r4\]](https://reviewable.io:443/reviews/cms-dev/cms/628#-KLWeWFSmjHqw2YgY6tZ:-KLXJA_4AInSiXv7MVTi:1738158546) ([raw file](https://github.com/cms-dev/cms/blob/fc11c870ab3697ddd7b73d8336ed66c017d988b6/cmscontrib/ExportSubmissions.py#L209)):*
> I think it is more clear if you move this outside the if args.utf8, and repeat it at the end, closer to the else:
</details>
I don't understand: outside the if args.utf8 means outside the else too, doesn't it?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/628)*

wafrelka:
Review status: 3 of 4 files reviewed at latest revision, 5 unresolved discussions.

---

*[cmscontrib/ExportSubmissions.py, line 207 \[r4\]](https://reviewable.io:443/reviews/cms-dev/cms/628#-KLaRvBK0X36BBOrZebh:-KLaRvBK0X36BBOrZebi:983759845) ([raw file](https://github.com/cms-dev/cms/blob/fc11c870ab3697ddd7b73d8336ed66c017d988b6/cmscontrib/ExportSubmissions.py#L207)):*
> ```Python
> 
>                 if args.utf8:
>                     data = utf8_decoder(data)
> ```

Maybe it's good to catch the `LookupError` exception and show a more user-friendly error message.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/628)*

wil93:
Review status: 3 of 4 files reviewed at latest revision, 5 unresolved discussions.

---

*[cmscontrib/ExportSubmissions.py, line 207 \[r4\]](https://reviewable.io:443/reviews/cms-dev/cms/628#-KLaRvBK0X36BBOrZebh:-KLaWba396Ex4PXUSR5m:-2119566704) ([raw file](https://github.com/cms-dev/cms/blob/fc11c870ab3697ddd7b73d8336ed66c017d988b6/cmscontrib/ExportSubmissions.py#L207)):*
<details><summary><i>Previously, wafrelka (wafrelka) wrote…</i></summary>
> Maybe it's good to catch the `LookupError` exception and show a more user-friendly error message.
</details>
Mmm when is that exception raised? Doesn't the `chardet.detect` method guarantee to always return some object with an `'encoding'` property? I thought it guaranteed that

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/628)*

wafrelka:
Review status: 3 of 4 files reviewed at latest revision, 5 unresolved discussions.

---

*[cmscontrib/ExportSubmissions.py, line 207 \[r4\]](https://reviewable.io:443/reviews/cms-dev/cms/628#-KLaRvBK0X36BBOrZebh:-KLaX7g8TqD2slHfx50-:-642625717) ([raw file](https://github.com/cms-dev/cms/blob/fc11c870ab3697ddd7b73d8336ed66c017d988b6/cmscontrib/ExportSubmissions.py#L207)):*
<details><summary><i>Previously, wil93 (William Di Luigi) wrote…</i></summary>
> Mmm when is that exception raised? Doesn't the `chardet.detect` method guarantee to always return some object with an `'encoding'` property? I thought it guaranteed that
</details>
`'encoding'` may become `"None"` and `value.decode("None")` will fail.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/628)*

wil93:
Review status: 3 of 4 files reviewed at latest revision, 5 unresolved discussions.

---

*[cmscontrib/ExportSubmissions.py, line 207 \[r4\]](https://reviewable.io:443/reviews/cms-dev/cms/628#-KLaRvBK0X36BBOrZebh:-KLaY3I33PQP7Fyf1_N2:1513717644) ([raw file](https://github.com/cms-dev/cms/blob/fc11c870ab3697ddd7b73d8336ed66c017d988b6/cmscontrib/ExportSubmissions.py#L207)):*
<details><summary><i>Previously, wafrelka (wafrelka) wrote…</i></summary>
> `'encoding'` may become `"None"` and `value.decode("None")` will fail.
</details>
Ok. But is seems to be `None`, not `"None"`... am I right? I tried with a PDF file and it gives encoding `None`. With a None value as encoding, the decode method raises a TypeError instead of a LookupError, it seems.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/628)*

wafrelka:
Review status: 3 of 4 files reviewed at latest revision, 5 unresolved discussions.

---

*[cmscontrib/ExportSubmissions.py, line 207 \[r4\]](https://reviewable.io:443/reviews/cms-dev/cms/628#-KLaRvBK0X36BBOrZebh:-KLanRRUowLgCfm9DC1_:2137045582) ([raw file](https://github.com/cms-dev/cms/blob/fc11c870ab3697ddd7b73d8336ed66c017d988b6/cmscontrib/ExportSubmissions.py#L207)):*
<details><summary><i>Previously, wil93 (William Di Luigi) wrote…</i></summary>
> Ok. But is seems to be `None`, not `"None"`... am I right? I tried with a PDF file and it gives encoding `None`. With a None value as encoding, the decode method raises a TypeError instead of a LookupError, it seems.
</details>
Sorry for the mistake. Actually it is `None`, not `"None"`. The actual error is `TypeError`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/628)*

wil93:
Review status: 2 of 4 files reviewed at latest revision, 5 unresolved discussions.

---

*[cmscontrib/ExportSubmissions.py, line 207 \[r4\]](https://reviewable.io:443/reviews/cms-dev/cms/628#-KLaRvBK0X36BBOrZebh:-KLayH2boWF1UESMTS81:-1003042541) ([raw file](https://github.com/cms-dev/cms/blob/fc11c870ab3697ddd7b73d8336ed66c017d988b6/cmscontrib/ExportSubmissions.py#L207)):*
<details><summary><i>Previously, wafrelka (wafrelka) wrote…</i></summary>
> Sorry for the mistake. Actually it is `None`, not `"None"`. The actual error is `TypeError`.
</details>
No problem, now TypeError exceptions should be caught

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/628)*

stefano-maggiolo:
Reviewed 1 of 1 files at r5, 1 of 1 files at r7.
Review status: 3 of 4 files reviewed at latest revision, 3 unresolved discussions.

---

*[cmscontrib/ExportSubmissions.py, line 138 \[r4\]](https://reviewable.io:443/reviews/cms-dev/cms/628#-KLWdiWK8DbA6aJZR2oY:-KMDjSBo_8AQ7VYDnDJ0:70939849) ([raw file](https://github.com/cms-dev/cms/blob/fc11c870ab3697ddd7b73d8336ed66c017d988b6/cmscontrib/ExportSubmissions.py#L138)):*
<details><summary><i>Previously, wil93 (William Di Luigi) wrote…</i></summary>
> I'm not sure about this...
> 
> 1. Explicit is better than implicit: I think it's better to explicitly force the user to know what's happening, and it's very easy to add a flag.
> 2. In general we may want to have a `--output-encoding=<encoding>` flag (it's not really needed since utf8 is enough, but maybe in the future) and, in that case, it's not reasonable to assume `--output-encoding=utf8` for `--add-info`; I think we should say "Error: you must specify some output encoding with `--output-encoding`".
</details>
I don't think we'll ever add other encoding as output - anyway, not too important, if you strongly prefer it like this.

---

*[cmscontrib/ExportSubmissions.py, line 209 \[r4\]](https://reviewable.io:443/reviews/cms-dev/cms/628#-KLWeWFSmjHqw2YgY6tZ:-KMDkJKmEKL6amW26O6S:1042856680) ([raw file](https://github.com/cms-dev/cms/blob/fc11c870ab3697ddd7b73d8336ed66c017d988b6/cmscontrib/ExportSubmissions.py#L209)):*
<details><summary><i>Previously, wil93 (William Di Luigi) wrote…</i></summary>
> I don't understand: outside the if args.utf8 means outside the else too, doesn't it?
</details>
More explicitly, I was suggesting this: 
```python
if utf8:
    # Convert data to utf8
if add_info:
    # Prepend info data
if utf8:
    # Write as utf8
else:
    # Write as bytes
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/628)*

stefano-maggiolo:
Merged!

PR #613, opened 2016-05-23
Add cmsImportDataset command

wil93:
Tired of uploading zips full of testcases via AdminWebServer? This pull request allows you to finally import new datasets from the command line!

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/613)

PR #629, opened 2016-06-30
Added script for cloning an existing task

akmohtashami:


wil93:
What is the use case for this? I recently proposed a `cmsImportDataset` command (#613) for the use case of needing to change task data (testcases, grader, ...) during a contest.

Also: why does this include so much code from DumpExporter? In order to clone a task it should be enough to just store a copy of: the Dataset, the statement file, and maybe some other things. It shouldn't be necessary to export anything, I think...

akmohtashami:
You may use this script when you want to use the same task in two or more contests. For example for training purposes or contest's analysis mode(Although hopefully I'll try to implement analysis mode separately soon). 

The export and import is actually the same as creating a copy of the task(though it might have one or two additional steps) but for the sake of simplicity of implementation I've used the code in DumpExporter and DumpImporter to create this script. Nothing is exported to a file though.

stefano-maggiolo:
I agree with @wil93, I wouldn't like having this code repeated for everything we want to clone...

I'll give another possibility: we already have a `Dataset.clone_from()` method, why don't we use that? It should be just a few lines to copy all the direct fields of task, and then a few more to clone the datasets.

PR #630, opened 2016-07-01
Add some SQL updaters

wil93:
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/630)

### DIFF
--- a/cmscontrib/updaters/update_10.sql
+++ b/cmscontrib/updaters/update_10.sql
@@ -1,0 +1,9 @@
+begin;
+
+update contests set start = '946684800' where start is null;
+update contests set stop = '4102444800' where stop is null;
+
+alter table contests alter start set not null;
+alter table contests alter stop set not null;
+
+rollback; -- change this to: commit;
--- a/cmscontrib/updaters/update_11.sql
+++ b/cmscontrib/updaters/update_11.sql
@@ -1,0 +1,5 @@
+begin;
+
+alter table contests add allowed_localizations not null default '';
+
+rollback; -- change this to: commit;
--- a/cmscontrib/updaters/update_12.sql
+++ b/cmscontrib/updaters/update_12.sql
@@ -1,0 +1,5 @@
+begin;
+
+alter table users add delay_time interval not null default '00:00:00'::interval;
+
+rollback; -- change this to: commit;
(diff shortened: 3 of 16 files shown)


stefano-maggiolo:
I'm sorry, as discussed we don't want to support two ways of updating DB schema version

wil93:
> I'm sorry, as discussed we don't want to support ~~two ways~~ **the correct way** of updating DB schema version

FTFY.

I want to also express how I feel that cms-dev is not really caring about the needs of one of its biggest users, if not the biggest, that is cms.di.unipi.it (I understand that CMS was not made for cms.di.unipi.it, but I don't thing one can reasonably argue that this proposal would be a huge change).

And again I will say that exporting + dropping + recreating + importing does not count as _updating_ the schema (just like buying a new car does not count as fixing it :stuck_out_tongue:)

stefano-maggiolo:
Two comments:

1. 
The key word is "some" (in "Add *some* SQL updaters"). Introducing them in the repo means that we support this way and thus we are going to continue doing so. "Some" is not a solution.

2.
If we recognize that this is something we do for cmsocial, then there is not much to lose if you host it in your repository. Especially since we said we don't have resources to keep them in sync with the official updaters.

Having said so, cmscontrib was (originally...) for unsupported resources possibly coming from outside, which would fit this definition. Nowadays it contains official parts, though.

So, if you really want to have those in cms-dev/cms, I prefer if we make it clear: send a PR to add them to cmscontrib/updaters/sql with a README stating that these are supported by you (or your GitHub organization) and not the cms-dev org, as it supports the dump updaters.

I like that CMS is the basis of derivative work that has substantially different goals, but I hope you understand why we try to maintain a line between our goals and those of our forks.

wil93:
> So, if you really want to have those in cms-dev/cms

Not really, I just wanted to be able to do in-place update schema updating, so if the `.sql` files are here but the new PRs won't contribute new `.sql` files that wouldn't change much.

> If we recognize that this is something we do for cmsocial, then there is not much to lose if you host it in your repository

We have CMS instances as well (e.g. for the team olympiads), it's not just a issue for cmsocial. Also, I don't think we're the only ones that would like to do in-place schema updating, IIRC, this problem was mentioned by others in the gitter chat.

> we said we don't have resources to keep them in sync with the official updaters.

This is the part I don't get :confused: from a new contributor perspective, I think it would be much easier to write a `.sql` updater (or `.py` updater which uses sqlalchemy) than writing a custom updater for some not-really-documented export format.

stefano-maggiolo:
Merged

PR #631, opened 2016-07-02
Admin panel bootstrap 3.3.6

valiro21:
Added boostrap 3.3.6 to admin panel for better aspect for multiple resolutions.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/631)


wil93:
I think one might want to use https://necolas.github.io/normalize.css instead, which is included in Bootstrap.

stefano-maggiolo:
All in favour of making aws more usable, but I tried this and the only things it seems to change is that it slightly break the left hand panel. Can you give an example where this patch helps?

stefano-maggiolo:
Closing for the moment, if you want to add more info please reopen!

valiro21:
Sorry I didn't respond earlier. It seemed to fix text overlapping on smaller screens for me. I will come back once I can reproduce the problem.

PR #632, opened 2016-07-06
Skip directory when handle writing permission in Sandbox.py

myungwoo:
When using TwoSteps task with comparator, error occurs.
Since Sandbox(probably "isolate") makes mapping directory with owner "root", writing_none method in IsolateSandbox tries to chmod for directory which its owner is root, thus "[Errno 1] Operation not permitted" occurs.

I'm not sure it's the best solution to handle this. However, considering the role of writing_none method, I think it is not a bad idea to solve this.


stefano-maggiolo:
Hello Myungwoo, do you know which directories create problems? I don't see how the checker can make a difference, it should not be a problem without the checker and it still should not be a problem with the checker. Also a test task showing this problem would be nice if you have one at hand. Thanks!

PR #636, opened 2016-07-29
Suggest problem's name for title in Announcement and Question

PJeBeK:
We can choose problem's name from drop down or just write a title.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/636)


stefano-maggiolo:
Thanks! I will just change the id from "problems" since we don't use that word to mean "tasks".

stefano-maggiolo:
Merged

Issue #637, opened 2016-08-12
Bug with custom checker


asdacap:
Hi there. I think I found a bug. When importing using the Polygon format, and the task uses a custom checker which is not named as `check.cpp`, the Polygon task loader will not find the custom checker. This seems to be fixed by checking for `check.cpp` in the main directory instead of in `files/check.cpp`, as for some reason the Polygon builder will copy the checker anyway. Modifying these two line seems to fix the issue.

https://github.com/cms-dev/cms/blob/master/cmscontrib/loaders/polygon.py#L189
https://github.com/cms-dev/cms/blob/master/cmscontrib/loaders/polygon.py#L192

wil93:
Could this is be a recent change in the polygon format? Maybe the best thing to do is to look in "files/check.cpp" and then (if the file is not there) in the main folder.

P.S. if you already have a patch in mind feel free to open a pull request, there are tutorials out there on how to do it if you're not familiar with them, e.g. [this one](https://yangsu.github.io/pull-request-tutorial/)

asdacap:
I've made a pull request. Unrelated to that, I suggest that the checker is compiler with C++11 flag enabled. The custom checker that I tested this with uses some C++11 feature, which causes warning... for now.

PR #638, opened 2016-08-13
Fixed issue with polygon task importer and custom checker.

asdacap:
- Previously it won't import the checker if the name of the checker's
  source file is not "check.cpp"

This is related to this [issue](https://github.com/cms-dev/cms/issues/637)

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/638)

### DIFF
--- a/cmscontrib/loaders/polygon.py
+++ b/cmscontrib/loaders/polygon.py
@@ -187,9 +187,13 @@ class PolygonTaskLoader(TaskLoader):
             outfile_param = judging.attrib['output-file']
 
             checker_src = os.path.join(self.path, "files", "check.cpp")
+            if not os.path.exists(checker_src):
+                checker_src = os.path.join(self.path, "check.cpp")
+            checker_dir = os.path.dirname(self.path)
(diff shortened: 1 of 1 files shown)


cmscontrib/loaders/polygon.py
```diff
@@ -187,9 +187,13 @@ def get_task(self, get_statement=True):
             outfile_param = judging.attrib['output-file']
 
             checker_src = os.path.join(self.path, "files", "check.cpp")
+            if not os.path.exists(checker_src):
+                checker_src = os.path.join(self.path, "check.cpp")
+            checker_dir = os.path.dirname(self.path)
```
stefano-maggiolo: Shouldn't this be `checker_dir = os.path.dirname(checker_src)` (or putting dirname() directly where you use it later)?

asdacap: Ow, yea. Did not realize that...

stefano-maggiolo:
Merged!

PR #640, opened 2016-08-15
Added analysis mode

akmohtashami:
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/640)

### DIFF
--- a/cms/db/contest.py
+++ b/cms/db/contest.py
@@ -8,6 +8,7 @@
 # Copyright © 2012-2014 Luca Wehrstedt <luca.wehrstedt@gmail.com>
 # Copyright © 2013 Bernard Blackham <bernard@largestprime.net>
 # Copyright © 2016 Myungwoo Chun <mc.tamaki@gmail.com>
+# Copyright © 2016 Amir Keivan Mohtashami <akmohtashami97@gmail.com>
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as
@@ -51,6 +52,8 @@ class Contest(Base):
     __tablename__ = 'contests'
     __table_args__ = (
         CheckConstraint("start <= stop"),
+        CheckConstraint("stop <= analysis_start"),
+        CheckConstraint("analysis_start <= analysis_stop"),
         CheckConstraint("token_gen_initial <= token_gen_max"),
     )
 
@@ -189,6 +192,21 @@ class Contest(Base):
         nullable=False,
         default=datetime(2000, 1, 1))
     stop = Column(
+        DateTime,
+        nullable=False,
+        default=datetime(2100, 1, 1))
+
+    # Beginning and ending of the contest anaylsis mode.
+    analysis_enabled = Column(
+        Boolean,
+        nullable=False,
+        default=False,
+    )
+    analysis_start = Column(
+        DateTime,
+        nullable=False,
+        default=datetime(2100, 1, 1))
+    analysis_stop = Column(
         DateTime,
         nullable=False,
         default=datetime(2100, 1, 1))
@@ -378,7 +396,11 @@ class Contest(Base):
     def phase(self, timestamp):
         """Return: -1 if contest isn't started yet at time timestamp,
                     0 if the contest is active at time timestamp,
-                    1 if the contest has ended.
+                    1 if the contest has ended but analysis mode
+                      hasn't started yet
+                    2 if the contest has eneded and analysis mode is active
+                    3 if the contest has ended and analysis mode is disabled or
+                      has ended
 
         timestamp (datetime): the time we are iterested in.
         return (int): contest phase as above.
@@ -388,7 +410,13 @@ class Contest(Base):
             return -1
         if self.stop is None or self.stop > timestamp:
             return 0
-        return 1
+        if self.analysis_enabled:
+            if self.analysis_start is not None and \
+                    self.analysis_start > timestamp:
+                return 1
+            elif self.analysis_stop is None or self.analysis_stop > timestamp:
+                return 2
+        return 3
 
     @staticmethod
     def _tokens_available(token_timestamps, token_mode,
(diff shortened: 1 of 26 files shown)


wil93:
Why does this PR include the feature of ignoring submissions? It would probably be better to open 2 separate PRs, one for ignoring submissions and another for analysis mode.

qnighy:
> Why does this PR include the feature of ignoring submissions? It would probably be better to open 2 separate PRs, one for ignoring submissions and another for analysis mode.

I think this is not problematic because the main intention of the "ignoring submissions" commit is the analysis mode feature (although it might have more use-cases). These commits are reasonably tied together to be included in one PR.

wil93:
OK, I probably misunderstood the meaning of ignoring submissions.

qnighy:
Review status: 0 of 22 files reviewed at latest revision, 2 unresolved discussions.

---

*[cms/server/contest/templates/base.html, line 138 \[r2\]](https://reviewable.io:443/reviews/cms-dev/cms/640#-KPJnL5SuwyavdyssqJI:-KPJnL5SuwyavdyssqJJ:b-mpkut7) ([raw file](https://github.com/cms-dev/cms/blob/bf6cc94a4c0ad82e7be37513c7662acac4eecb9d/cms/server/contest/templates/base.html#L138)):*
> ```HTML
>         </div>
>         <div style="display: none" id="translation_until_analysis_starts">
>             {{ _("Until analysis starts:") }}
> ```

I think `cms.pot` should be regenerated in accordance with http://cms.readthedocs.io/en/latest/Localization.html when we add new texts.

---

*[cms/server/contest/templates/submission_row.html, line 83 \[r2\]](https://reviewable.io:443/reviews/cms-dev/cms/640#-KPJrPzFFiRtQId_xUGU:-KPJrPzFFiRtQId_xUGV:b1fh6u4) ([raw file](https://github.com/cms-dev/cms/blob/bf6cc94a4c0ad82e7be37513c7662acac4eecb9d/cms/server/contest/templates/submission_row.html#L83)):*
> ```HTML
>     <td class="considered_in_score">
>       {% if s.ignored is False %}
>       Yes
> ```

Is there any chance to make them translatable? (And if so, possibly "Yes"/"No" are too general for translating this specific meaning.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/640)*

qnighy:
Review status: 0 of 22 files reviewed at latest revision, 3 unresolved discussions.

---

*[cms/grading/__init__.py, line 961 \[r2\]](https://reviewable.io:443/reviews/cms-dev/cms/640#-KPJsYBJp91XMVtEgfJM:-KPJsYBKcDb-RBbylYgy:bbjm7v7) ([raw file](https://github.com/cms-dev/cms/blob/bf6cc94a4c0ad82e7be37513c7662acac4eecb9d/cms/grading/__init__.py#L961)):*
> ```Python
> 
>     submissions = [s for s in participation.submissions if s.task is task and
>                    s.ignored]
> ```

Do you mean `not s.ignored`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/640)*

qnighy:
If I'm correct, the patch currently doesn't seem to work around `RankingWebServer`/`ProxyService`. Am I right?

akmohtashami:
I almost fixed all the problems mentioned(except one which I will talk about below) and some others too. Also, changed the code so that in analysis mode all submissions total score(tokened or not) are shown to the contestant.

However, regarding the problem you mentioned about Yes/No and translating them, I think currently it would suffice to make them translatable and there is no need for more specific words in any language. However I am there are many languages I am not familiar with, so if you think this is a serious issue, please give me some suggestions as well. 

Additionally, currently I am still having doubts regarding the name "ignored". I am looking for something better and I would be very happy to hear your recommendations about it.

akmohtashami:
Review status: 0 of 26 files reviewed at latest revision, 3 unresolved discussions.

---

*[cms/grading/__init__.py, line 961 \[r2\]](https://reviewable.io:443/reviews/cms-dev/cms/640#-KPJsYBJp91XMVtEgfJM:-KPizD1i-mdeggRSefAl:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/bf6cc94a4c0ad82e7be37513c7662acac4eecb9d/cms/grading/__init__.py#L961)):*
<details><summary><i>Previously, qnighy (Masaki Hara) wrote…</i></summary>
> Do you mean `not s.ignored`?
</details>
Done.

---

*[cms/server/contest/templates/base.html, line 138 \[r2\]](https://reviewable.io:443/reviews/cms-dev/cms/640#-KPJnL5SuwyavdyssqJI:-KPizExXG4uW-UgVddLz:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/bf6cc94a4c0ad82e7be37513c7662acac4eecb9d/cms/server/contest/templates/base.html#L138)):*
<details><summary><i>Previously, qnighy (Masaki Hara) wrote…</i></summary>
> I think `cms.pot` should be regenerated in accordance with http://cms.readthedocs.io/en/latest/Localization.html when we add new texts.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/640)*

cms/db/contest.py
```diff
@@ -193,6 +196,21 @@ class Contest(Base):
...
+        default=False,
+    )
+    analysis_start = Column(
+        DateTime,
+        nullable=False,
+        default=datetime(2100, 1, 1))
+    analysis_stop = Column(
+        DateTime,
+        nullable=False,
+        default=datetime(2100, 1, 1))
```
wil93: Why not just "analysis_enabled"? Is there a good reason to have "analysis_start" and "analysis_stop"? I would just have CWS show analysis mode (i.e. the possibility to download submissions and to test new submission but without affecting the score) *if and only if* "analysis_enabled" is true, regardless of the date.

stefano-maggiolo:
I'm rebasing this and will send a new pr to continue the discussion (similarly to what was done with multi-cws).

cms/db/contest.py
stefano-maggiolo: I like having times - for online contests it could be nice to show "X minutes to start of analysis".

stefano-maggiolo: (or if you meant enable analysis as soon as the contest ended, you don't want that for several reasons on onsite contests)

stefano-maggiolo:
Continues in #746

PR #641, opened 2016-08-16
Fix buggy equal count

ecnerwala:
The count of users with equal score is incorrect if another equal user's score is updated without changing it. This is especially obvious at the start of a contest when there are many submits which all receive 0.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/641)

### DIFF
--- a/cmsranking/static/HistoryStore.js
+++ b/cmsranking/static/HistoryStore.js
@@ -173,10 +173,10 @@ var HistoryStore = new function () {
                         above -= 1;
                         changed = true;
                     }
-                    if (d[user] == d[user_id]) {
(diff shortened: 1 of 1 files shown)


ecnerwala:
Fixes #642

lerks:
A history entry always updates the score, i.e. every line in the output given by the `/history` endpoint means a change in the score for that user/task. (For your reference, the code that ensures it is [1].) Hence the checks you are adding would always evaluate to true. I'm closing this PR.

[1] https://github.com/cms-dev/cms/blob/044ff246c4d13537690bc371353b76aeba1dd77d/cmsranking/Scoring.py#L128

PR #644, opened 2016-08-20
Add Haskell as a supported language

mark03:
@veluca93 and I added Haskell to the supported languages of cms

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/644)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ sudo: required
 
 install:
   - sudo apt-get update -qq
-  - sudo apt-get install -y build-essential fpc php5-cli gcj-jdk ghc
+  - sudo apt-get install -y build-essential fpc php5-cli gcj-jdk haskell-platform
   - sudo apt-get install -y postgresql-client gettext python2.7 iso-codes shared-mime-info stl-manual cgroup-lite
   - sudo apt-get install -y python-dev libpq-dev libcups2-dev libyaml-dev
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ sudo: required
 
 install:
   - sudo apt-get update -qq
-  - sudo apt-get install -y build-essential fpc php5-cli gcj-jdk
+  - sudo apt-get install -y build-essential fpc php5-cli gcj-jdk ghc
   - sudo apt-get install -y postgresql-client gettext python2.7 iso-codes shared-mime-info stl-manual cgroup-lite
   - sudo apt-get install -y python-dev libpq-dev libcups2-dev libyaml-dev
 
--- a/cms/__init__.py
+++ b/cms/__init__.py
@@ -35,7 +35,7 @@ import cms.log
 
 __all__ = [
     "__version__",
-    "LANG_C", "LANG_CPP", "LANG_PASCAL", "LANG_PYTHON", "LANG_PHP",
+    "LANG_C", "LANG_CPP", "LANG_PASCAL", "LANG_PYTHON", "LANG_PHP", "LANG_HS",
     "LANGUAGE_NAMES", "LANGUAGES", "DEFAULT_LANGUAGES",
     "SOURCE_EXT_TO_LANGUAGE_MAP", "filename_to_language",
     "LANGUAGE_TO_SOURCE_EXT_MAP", "LANGUAGE_TO_HEADER_EXT_MAP",
@@ -78,7 +78,8 @@ LANGUAGE_NAMES = {
     LANG_HS: "Haskell",
 }
 
-LANGUAGES = [LANG_C, LANG_CPP, LANG_PASCAL, LANG_PYTHON, LANG_PHP, LANG_JAVA, LANG_HS]
+LANGUAGES = [LANG_C, LANG_CPP, LANG_PASCAL, LANG_PYTHON, LANG_PHP, LANG_JAVA, 
+             LANG_HS]
 DEFAULT_LANGUAGES = [LANG_C, LANG_CPP, LANG_PASCAL]
 
 # A reference for extension-based automatic language detection.
--- a/cms/__init__.py
+++ b/cms/__init__.py
@@ -66,6 +66,7 @@ LANG_PASCAL = "pas"
 LANG_PYTHON = "py"
 LANG_PHP = "php"
 LANG_JAVA = "java"
+LANG_HS = "hs"
 
 LANGUAGE_NAMES = {
     LANG_C: "C",
@@ -74,9 +75,10 @@ LANGUAGE_NAMES = {
     LANG_PYTHON: "Python",
     LANG_PHP: "PHP",
     LANG_JAVA: "Java",
+    LANG_HS: "Haskell",
 }
 
-LANGUAGES = [LANG_C, LANG_CPP, LANG_PASCAL, LANG_PYTHON, LANG_PHP, LANG_JAVA]
+LANGUAGES = [LANG_C, LANG_CPP, LANG_PASCAL, LANG_PYTHON, LANG_PHP, LANG_JAVA, LANG_HS]
 DEFAULT_LANGUAGES = [LANG_C, LANG_CPP, LANG_PASCAL]
 
 # A reference for extension-based automatic language detection.
@@ -92,6 +94,7 @@ SOURCE_EXT_TO_LANGUAGE_MAP = {
     ".py": LANG_PYTHON,
     ".php": LANG_PHP,
     ".java": LANG_JAVA,
+    ".hs": LANG_HS
 }
 
 # Our preferred source file and header file extension for each language.
@@ -102,6 +105,7 @@ LANGUAGE_TO_SOURCE_EXT_MAP = {
     LANG_PYTHON: ".py",
     LANG_PHP: ".php",
     LANG_JAVA: ".java",
+    LANG_HS: ".hs",
 }
 LANGUAGE_TO_HEADER_EXT_MAP = {
     LANG_C: ".h",
@@ -112,6 +116,7 @@ LANGUAGE_TO_OBJ_EXT_MAP = {
     LANG_C: ".o",
     LANG_CPP: ".o",
     LANG_PASCAL: ".o",
+    LANG_HS: ".o",
 }
 
 
--- a/cms/__init__.py
+++ b/cms/__init__.py
@@ -78,8 +78,16 @@ LANGUAGE_NAMES = {
     LANG_HS: "Haskell",
 }
 
-LANGUAGES = [LANG_C, LANG_CPP, LANG_PASCAL, LANG_PYTHON, LANG_PHP, LANG_JAVA, 
-             LANG_HS]
+LANGUAGES = [
+    LANG_C,
+    LANG_CPP,
+    LANG_PASCAL,
+    LANG_PYTHON,
+    LANG_PHP, 
+    LANG_JAVA, 
+    LANG_HS
+]
+
 DEFAULT_LANGUAGES = [LANG_C, LANG_CPP, LANG_PASCAL]
 
 # A reference for extension-based automatic language detection.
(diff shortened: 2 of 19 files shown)


stefano-maggiolo:
Hi, thanks for the patch. Could you add at least one functional test? We also need to give the dependency for ghc somewhere, look for the places where we say something about free pascal.

---

Reviewed 3 of 3 files at r1.

---

*[cms/__init__.py, line 81 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/644#-KPbjjZ0PiqRNlJpWPDV-r1-81:-KPm4TUxmp68N8BCoo5A:b-rll82r) ([raw file](https://github.com/cms-dev/cms/blob/332e980716b58b26ecc39e7e93d5cda42a295a43/cms/__init__.py#L81)):*
<details><summary><i>Previously, codacy-bot (Codacy Bot) wrote…</i></summary>
> Codacy Issue found: [Line too long (86/80)](https://www.codacy.com/app/s-maggiolo/cms/file/3445283919/issues/source?bid=3167167&fileBranchId=3614239#l81)
</details>
Please fix!

---

*[cms/grading/__init__.py, line 270 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/644#-KPm4Vz1CVaJ6ua3vtKq:-KPm4Vz1CVaJ6ua3vtKr:bimfb9e) ([raw file](https://github.com/cms-dev/cms/blob/332e980716b58b26ecc39e7e93d5cda42a295a43/cms/grading/__init__.py#L270)):*
> ```Python
>     elif language == LANG_HS:
>         # Haskell module names are capitalized, so we change
>         # the source file names (except for the first one)
> ```

Why not the first one?

---

*[cmstaskenv/cmsMake.py, line 234 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/644#-KPm541sJGARD_U2y3Kx:-KPm541sJGARD_U2y3Ky:b-bl694o) ([raw file](https://github.com/cms-dev/cms/blob/332e980716b58b26ecc39e7e93d5cda42a295a43/cmstaskenv/cmsMake.py#L234)):*
> ```Python
>                 new_srcs.append(source_name)
>                 # Libraries are needed/used only for C/C++ and Pascal
>                 lib_template = "%slib.pas" if lang == LANG_PASCAL else "%s.h"
> ```

I think you can use something in cms to get these names too

---

*[cmstaskenv/cmsMake.py, line 239 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/644#-KPm5EdMXSkUsf_8-J7P:-KPm5EdMXSkUsf_8-J7Q:bh1fhea) ([raw file](https://github.com/cms-dev/cms/blob/332e980716b58b26ecc39e7e93d5cda42a295a43/cmstaskenv/cmsMake.py#L239)):*
> ```Python
>                 if os.path.exists(lib_path):
>                     shutil.copyfile(lib_path, os.path.join(tempdir,
>                                                            lib_filename))
> ```

Nit: wrapping at higher level is usually better (in this case before os.path.join())

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/644#-:-KPm4DaqoOzxGVe2JU5o:bquezdj)*

mark03:
Review status: 0 of 19 files reviewed at latest revision, 4 unresolved discussions.

---

*[cms/__init__.py, line 81 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/644#-KPbjjZ0PiqRNlJpWPDV-r1-81:-KPmHqBxGLvA7gBft9pZ:b32w1t) ([raw file](https://github.com/cms-dev/cms/blob/332e980716b58b26ecc39e7e93d5cda42a295a43/cms/__init__.py#L81)):*
> Please fix!
</details>
Ok

---

*[cms/grading/__init__.py, line 270 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/644#-KPm4Vz1CVaJ6ua3vtKq:-KPmEjLKW73skEOwswY6:b-cgwliz) ([raw file](https://github.com/cms-dev/cms/blob/332e980716b58b26ecc39e7e93d5cda42a295a43/cms/grading/__init__.py#L270)):*
> Why not the first one?
</details>
In Haskell, to import a module, the file must match the module's name.  The first source file is, instead, the grader or the standalone source file; they won't be imported in any other source file, so there is no need to capitalize them.

---

*[cmstaskenv/cmsMake.py, line 234 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/644#-KPm541sJGARD_U2y3Kx:-KPmHW2lQRJNKOjZl63h:bb6un1j) ([raw file](https://github.com/cms-dev/cms/blob/332e980716b58b26ecc39e7e93d5cda42a295a43/cmstaskenv/cmsMake.py#L234)):*
> I think you can use something in cms to get these names too
</details>
Done

---

*[cmstaskenv/cmsMake.py, line 239 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/644#-KPm5EdMXSkUsf_8-J7P:-KPmHkvclbK2NnHuQlBc:bb6un1j) ([raw file](https://github.com/cms-dev/cms/blob/332e980716b58b26ecc39e7e93d5cda42a295a43/cmstaskenv/cmsMake.py#L239)):*
> Nit: wrapping at higher level is usually better (in this case before os.path.join())
</details>
Done

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/644)*

stefano-maggiolo:
Thanks for all the tests!

---

Review status: 0 of 19 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[cms/__init__.py, line 81 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/644#-KPbjjZ0PiqRNlJpWPDV-r1-81:-KPnc_geQwAfEn1pN64h:bw3ttc6) ([raw file](https://github.com/cms-dev/cms/blob/332e980716b58b26ecc39e7e93d5cda42a295a43/cms/__init__.py#L81)):*
<details><summary><i>Previously, mark03 (Marco Donadoni) wrote…</i></summary>
> Ok
</details>
Nit: similarly to other discussion, you can wrap right after the '['

---

*[cms/grading/__init__.py, line 270 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/644#-KPm4Vz1CVaJ6ua3vtKq:-KPnbp8GuSMMjWvXgERs:b32jdyf) ([raw file](https://github.com/cms-dev/cms/blob/332e980716b58b26ecc39e7e93d5cda42a295a43/cms/grading/__init__.py#L270)):*
<details><summary><i>Previously, mark03 (Marco Donadoni) wrote…</i></summary>
> In Haskell, to import a module, the file must match the module's name.  The first source file is, instead, the grader or the standalone source file; they won't be imported in any other source file, so there is no need to capitalize them.
</details>
Thanks! Now copy and paste this into the comment :)

---

*[cms/grading/__init__.py, line 311 \[r2\]](https://reviewable.io:443/reviews/cms-dev/cms/644#-KPnc-a2GAW3WjnozLEK:-KPnc-a2GAW3WjnozLEL:b-zaoh75) ([raw file](https://github.com/cms-dev/cms/blob/0a4ae0a67ea6add08a10ffde9c931fbac5221d2c/cms/grading/__init__.py#L311)):*
> ```Python
>         commands.append(command)
>     elif language == LANG_HS:
>         command = [os.path.join(".", executable_filename), "+RTS", "-K8G",
> ```

Looking at docs, it seems -K0 is more elegant. But also, is this really necessary? Does it get confused within isolate for some reason? (I read the default is 80% physical RAM)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/644#-:-KPncgkCW7jLgtrCF0Fu:b-6290qx)*

mark03:
Review status: 0 of 19 files reviewed at latest revision, 3 unresolved discussions.

---

*[cms/__init__.py, line 81 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/644#-KPbjjZ0PiqRNlJpWPDV-r1-81:-KPnu8VYkgCLwPkVoPzX:b-iem4it) ([raw file](https://github.com/cms-dev/cms/blob/332e980716b58b26ecc39e7e93d5cda42a295a43/cms/__init__.py#L81)):*
> Nit: similarly to other discussion, you can wrap right after the '['
</details>
Should be better now

---

*[cms/grading/__init__.py, line 270 \[r1\]](https://reviewable.io:443/reviews/cms-dev/cms/644#-KPm4Vz1CVaJ6ua3vtKq:-KPnuB76RC5Zb9_gEo3T:b-896fja) ([raw file](https://github.com/cms-dev/cms/blob/332e980716b58b26ecc39e7e93d5cda42a295a43/cms/grading/__init__.py#L270)):*
> Thanks! Now copy and paste this into the comment :)
</details>
Done!

---

*[cms/grading/__init__.py, line 311 \[r2\]](https://reviewable.io:443/reviews/cms-dev/cms/644#-KPnc-a2GAW3WjnozLEK:-KPnuDOgLEerlqEzjouA:b-f67jc9) ([raw file](https://github.com/cms-dev/cms/blob/0a4ae0a67ea6add08a10ffde9c931fbac5221d2c/cms/grading/__init__.py#L311)):*
> Looking at docs, it seems -K0 is more elegant. But also, is this really necessary? Does it get confused within isolate for some reason? (I read the default is 80% physical RAM)
</details>
There seems to be differences between different versions of ghc: in some of them (the older ones) the stack is limited to 8M, in newer ones the stack is limited to 80% of physical ram. Also, the option "-K0" doesn't seem to work in any of the ghc version I could try and even the "-K8G" doesn't seem to work well in some cases. So, for now, is probably better if we remove this option.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/644)*

stefano-maggiolo:
:lgtm:

---

Reviewed 2 of 2 files at r5.
Review status: 2 of 19 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[cms/grading/__init__.py, line 311 \[r2\]](https://reviewable.io:443/reviews/cms-dev/cms/644#-KPnc-a2GAW3WjnozLEK:-KPpy1d29vqz53-4l5Na:b-scm62x) ([raw file](https://github.com/cms-dev/cms/blob/0a4ae0a67ea6add08a10ffde9c931fbac5221d2c/cms/grading/__init__.py#L311)):*
<details><summary><i>Previously, mark03 (Marco Donadoni) wrote…</i></summary>
> There seems to be differences between different versions of ghc: in some of them (the older ones) the stack is limited to 8M, in newer ones the stack is limited to 80% of physical ram. Also, the option "-K0" doesn't seem to work in any of the ghc version I could try and even the "-K8G" doesn't seem to work well in some cases. So, for now, is probably better if we remove this option.
</details>
Oh I see. Thanks for explanation.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/644#-:-KPpy82XoO1nOOMBS4pc:bnfp4nl)*

stefano-maggiolo:
I added haskell-platform to the ubuntu installation packages, and ghc to the arch ones (apparently). Still, functional tests on my ubuntu with haskell-platform installed are failing. ghc cannot compile and prints

```ghc: can't find a package database at /usr/lib/ghc/package.conf.d```

any idea?

mark03:
GHC is probably looking for the database of haskell's package (haskell has, indeed, a package manager called "cabal"), which he cannot access when it compiles in isolate.

mark03:
Ok, turns out `/usr/lib/ghc/package.conf.d` is just a symlink to `/var/lib/ghc/package.conf.d` (at least on Ubuntu), which wasn't accessible during compilation due to the restrictions of the sandbox.

stefano-maggiolo:
Not sure why these things are in /var... 
Do you know if there is a way to tell the compiler not to look there? If not, let's add the minimum needed (/var/lib/ghc, I assume). Also please add a comment saying why we need it. Thanks!

veluca93:
On a (relatively) unrelated note: shouldn't the list of allowed directories at compilation (or run) time be defined by the language of the submission?

stefano-maggiolo:
We like to allow the same ability to cheat regardless of the language

mark03:
I think everything should be ok now

stefano-maggiolo:
FYI if ghc wasn't installed no submissions could be compiled :)
I added a check, submitting this as soon as it passes on travis

mark03:
Perfect!

PR #594, opened 2016-03-26
Make CWS multi-contest

wil93:
With this patch, ContestWebServer does not require a `-c` flag anymore. When you start CWS, it will expose a list of all available contests on `/`, and you can click on the contest that you're interested in.

The former cookie behavior was to save 2 cookies: `login` and `unread_count`. The new cookie behavior is to save 2*N cookies where N is the number of contests. Each cookie will either be `<contest-name>_login` or `<contest-name>_unread_count`.

This PR depends on #545

---

This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/594)

Issue #73, opened 2012-11-27
Allow services to handle many contests at the same time


lw:
At the moment many services require a contest ID as a command line argument. I think that their use of this ID is very limited (except in some cases like CWS). I think that they could manage multiple contest simultaneously. For example AWS already does that and, in fact, it doesn't require that command line argument. That would mean, for example, that Workers could accept job requests for different contests, that ES should manage submissions of many contests (all in the same queue?), and the same for SS.

I therefore suggest to check exactly how services use the contest ID and check if it's possible to remove it, making them able to handle data from all contests at the same time.

What do you think?

PS: the reason for this is that, for example, for training purposes it seems convenient to run many contest simultaneously on the same machine, and it seems excessive to require dedicated services for each of them (actually, I'm not even sure if, for example, many ESs on the same machine would work correctly...)

PR #646, opened 2016-08-23
Support open participation in contests + hidden contests

wil93:
This is useful if you have some users already in the database, and you want them to be able to automatically sign up to one or more contests without having to create a participation for them.

This PR depends on #594

---

UPDATE: I also added a commit which introduces Contest.hidden, a way to hide contests from the list (useful for contests which aren't ready yet, e.g. the "day2" of a 2-day contest, or for special contests which are not meant to be consumed by CWS, as in my use case).

I didn't create another PR because:
1. that PR would also need to be based on the multi_cws branch so, when that is merged, I would need to rebase two PRs instead of just one
2. it's a quite small change, after all

---

This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/646)

### DIFF
--- a/cms/db/contest.py
+++ b/cms/db/contest.py
@@ -127,6 +127,13 @@ class Contest(Base):
         nullable=False,
         default=False)
 
+    # Whether to automatically create a participation for users when they log in
+    # to this contest. Useful for multi-contest setups.
+    open_participation = Column(
+        Boolean,
+        nullable=False,
+        default=False)
+
     # The parameters that control contest-tokens follow. Note that
     # their effect during the contest depends on the interaction with
     # the parameters that control task-tokens, defined on each Task.
--- a/cms/db/contest.py
+++ b/cms/db/contest.py
@@ -134,6 +134,14 @@ class Contest(Base):
         nullable=False,
         default=False)
 
+    # Whether this contest is hidden from the contest list. Useful for hiding
+    # contests which shouldn't be accessed yet, or contests which aren't meant
+    # to be participated on (e.g. a contest which is not processed by CWS).
+    hidden = Column(
+        Boolean,
+        nullable=False,
+        default=False)
+
     # The parameters that control contest-tokens follow. Note that
     # their effect during the contest depends on the interaction with
     # the parameters that control task-tokens, defined on each Task.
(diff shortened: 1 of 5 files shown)


wil93:
@cms-dev now that multicontest is in, can we merge also this (tiny :pray:) feature? 😄

lerks:
Please wait a second on this.

I am planning to make PS support multiple contest soon. (It's the last service that needs to be ported). After that is done there will be at least two scenarios where we'd want to disable contest, and they don't necessarily overlap: the ones we present to contestants in CWS and the ones we present to the public in RWS (and thus PS). (There's also the ones we evaluate, but since datasets can already be disabled this is sort-of already doable).

See also https://github.com/cms-dev/cms/issues/73#issuecomment-12644421

stefano-maggiolo:
I haven't read this thread, but I believe active datasets cannot be disabled for judging.

lerks:
Oh yes I think you're right. Well, in that case there's three possibly
independent locations where we'd want to disable a contest.

wil93:
What if I remove the third commit and leave just the "open participation" stuff?

wil93:
I just realized that maybe the open_participation thing should be improved.

Right now, when enabled, this feature will:

1. _silently_ create a participation for users
2. create it even if the contest has ended

For point 1, I think it's best to inform the user with an intermediate step, a page that says: "You are registering for this contests, do you want to proceed?", maybe with a textarea explaining the contest rules (we could add a nullable `Contest.rules` text field in the database model) like it's done on Codeforces.

For point 2, I don't have a strong opinion, but maybe it's better to avoid registering users when it's "too late". We could use the "intermediate step" page to inform the user that they can't register.

totorigolo:
Are there any updates regarding this PR? I would enjoy to help if needed :)

wil93:
FWIW, an "update" on this is that we're about to merge #1067.

That PR will make the "open participation contests" pretty much unnecessary, since now you can just enable "allow_registration" for the contests that you want to open up to the public and that will let people create new accounts (thus circumventing the original problem).

The only problem is that the same user might need to create multiple accounts (necessarily with different usernames) if they want to participate in multiple contests.

PR #654, opened 2016-09-11
Php and sandboxes

stefano-maggiolo:
Update to ubuntu 16.04 (in particular update PHP version) and keep sandboxes around if the job in them was not successful.

---

This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/654)


stefano-maggiolo:
Merged the keep sandboxes on unsuccessful job part, the rest are obsolete.

PR #655, opened 2016-09-19
Parallel startup of processes in functional tests

stefano-maggiolo:
In my machine, this reduces the startup time (till first submission)
from 21s to 9s.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/655)

### DIFF
--- a/cmstestsuite/programstarter.py
+++ b/cmstestsuite/programstarter.py
@@ -1,0 +1,280 @@
+#!/usr/bin/env python2
+# -*- coding: utf-8 -*-
+
+# Contest Management System - http://cms-dev.github.io/
+# Copyright © 2012 Bernard Blackham <bernard@largestprime.net>
+# Copyright © 2013-2016 Stefano Maggiolo <s.maggiolo@gmail.com>
+# Copyright © 2013-2014 Luca Wehrstedt <luca.wehrstedt@gmail.com>
+# Copyright © 2014 Luca Versari <veluca93@gmail.com>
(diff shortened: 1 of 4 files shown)


cmstestsuite/programstarter.py
```diff
@@ -0,0 +1,280 @@
...
+        t.start()
+
+    @property
+    def running(self):
+        """Return whether the program is live."""
+        self._check()
+        return self.healthy
+
+    def stop(self):
+        """Quit gracefully. Or not."""
```
veluca93: Maybe this docstring could be a bit more explicit (why "or not"?)

cmstestsuite/programstarter.py
```diff
@@ -0,0 +1,280 @@
...
+from urlparse import urlsplit
+
+from cmstestsuite import CONFIG, FrameworkException, get_cms_config
+
+
+logger = logging.getLogger(__name__)
+
+
+# Maximum number of attempts to check if a service becomes healthy.
+_MAX_ATTEMPTS = 20
```
veluca93: Shouldn't this belong to a class?

stefano-maggiolo: Maybe, but I'm using it in two classes and you can argue is a constant of the module.

cmstestsuite/programstarter.py
stefano-maggiolo: Done.

PR #656, opened 2016-09-21
Make documentation for DB configuration a bit more clear

wil93:
I believe that by using the `-U` argument instead of directly logging in as the postgres user, the DB configuration steps described in the documentation become more intuitive. Moreover, I found that using the expanded version of each flag (`--username` instead of `-U`) makes it even more clear to newcomers. It's always a good idea to have a vision of what's happening that is as clear as possible. I also changed the `database` and `password` defaults, which may look like keywords.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/656)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,14 +24,15 @@ install:
   - sudo pip install -r dev-requirements.txt
   - sudo python setup.py install
 
-  - sudo -u postgres psql -c "CREATE USER cmsuser WITH PASSWORD 'password';"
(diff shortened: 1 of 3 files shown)


stefano-maggiolo:
Ok but there is some problem with the test (assume you'll have to change name of db and password in the script too)

wil93:
Done

stefano-maggiolo:
Merged.

PR #657, opened 2016-09-26
Some performance improvements to CWS and the submissions page in particular

stefano-maggiolo:
Test setup:
- 1 contest (coming from functional test)
- 5 tasks
- 1 dataset/task
- 50 submissions/task
- 150 participations
- repeatedly querying the submissions page with 1 or 5 threads 

```
Master:                 4.05 qps (1 thread) /  5.64 qps (5 threads)
Joinedload SR:         11.90                / 13.51
Passing participation: 12.34                / 15.67
Skip user query:       13.33                / 19.08
```

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/657)

### DIFF
--- a/cms/db/contest.py
+++ b/cms/db/contest.py
@@ -284,24 +284,6 @@ class Contest(Base):
                 return idx
         raise KeyError("Task not found")
 
-    # FIXME - Use SQL syntax
-    def get_participation(self, username):
-        """Return the first participation in the contest with the given
-        username.
-
-        username (string): the name of the user we are interested in.
-
-        return (Participation): the corresponding participation object.
-
-        raise (KeyError): if no users with the given name participate.
-
-        """
-
-        for participation in self.participations:
-            if participation.user.username == username:
-                return participation
-        raise KeyError("Participation not found")
-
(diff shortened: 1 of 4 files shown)


cms/db/contest.py
```diff
@@ -542,7 +524,7 @@ def tokens_available(self, username, task_name, timestamp=None):
         future. Also, if r[0] == 0 and r[1] is None, then r[2] should
         be ignored.
 
-        username (string): the username of the user.
+        participation (Participaiton): the participation.
```
veluca93: Participaiton should be Participation

cms/db/contest.py
```diff
@@ -542,7 +524,7 @@ def tokens_available(self, username, task_name, timestamp=None):
...
         task_name (string): the name of the task.
```
veluca93: Since we pass a Participation object, shouldn't we pass a Task object instead of a task_name string?

stefano-maggiolo: Good point.

cms/db/contest.py
stefano-maggiolo: Done

PR #668, opened 2016-10-21
Fix install error: Rename requirements.txt to REQUIREMENTS.txt

woooook:
Hello.

I've got following error while installing cms.

```
root@localhost:~# sudo pip install -r REQUIREMENTS.txt  
Could not open requirements file: [Errno 2] No such file or directory: 'REQUIREMENTS.txt'  
Storing debug log for failure in /root/.pip/pip.log  
```
  
According to this [manual](http://cms.readthedocs.io/en/v1.2/Installation.html#dependencies), it saids "sudo pip install -r REQUIREMENTS.txt" but the actual file name was requirements.txt, so I renamed file like so.  

Thank you.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/668)


wil93:
Version 1.2 of CMS uses `REQUIREMENTS.txt`, as in the manual you linked.

However, version 1.3+ of CMS uses `requirements.txt` as in this manual: http://cms.readthedocs.io/en/latest/Installation.html#dependencies

So, if you're using the most recent version of CMS (and you probably are :smile:) then you just need to use the *new* manual (with `latest` instead of `v1.2` in the URL).

woooook:
Oh, I think I've just confused master branch was the stable version.

Thank you @wil93 :)

PR #675, opened 2016-11-26
Allow checker to fork processes

wil93:
We needed this patch when developing a task

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/675)

### DIFF
--- a/cms/grading/tasktypes/Batch.py
+++ b/cms/grading/tasktypes/Batch.py
@@ -349,6 +349,12 @@ class Batch(TaskType):
                             sandbox.create_file_from_storage(
                                 input_filename,
                                 job.input)
+
+                            # Allow using any number of processes (because e.g.
+                            # one may want to write a bash checker who calls
+                            # other processes)
+                            sandbox.max_processes = 1000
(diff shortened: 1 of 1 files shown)


akmohtashami:
You can set the max_processes to None to allow unlimited number of processes. However this allows the possibility of unintentionally fork-bombing the workers. So I think it might be reasonable to write a comment explaining why this limit is being set to a big number rather than None.

wil93:
`None` would work for my use case, I just didn't remember that it was a possibility :smile: 

Anyway I can see why one would like to avoid being unintentionally forkbombed by their checker, so it's probably good to just have a big number there...

stefano-maggiolo:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/675#-:-KY47eL4kDnYf8VTyYtI:bnfp4nl)*

stefano-maggiolo:
Just added the justification to "1000" in the comment. Merged.

PR #680, opened 2016-12-10
Fix HTML validation

andreyv:
With these changes, CWS and AWS pages (as served by the servers and without JS DOM modifications) now pass validator checks at https://validator.w3.org/.

RWS is partially fixed, but still does not pass validation because of disallowed empty elements that are used by JS to generate content.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/680)

### DIFF
--- a/cms/server/admin/templates/announcements.html
+++ b/cms/server/admin/templates/announcements.html
@@ -20,6 +20,7 @@
               <option value="{{ task.name }}">
               {% end %}
             </datalist>
+          </div>
           <div class="notification_text">
             <label for="new_text">Text</label>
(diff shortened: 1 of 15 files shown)


stefano-maggiolo:
LGTM!

---

Reviewed 7 of 7 files at r1, 5 of 6 files at r2, 1 of 1 files at r3, 1 of 2 files at r4, 1 of 1 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/680#-:-KZ2TSwhxHmjJNSUmrmT:b-85i648)*

stefano-maggiolo:
Merged, thanks!

PR #687, opened 2016-12-21
[cmsMake] Automatically respond 'n' after a 5 seconds timeout

wil93:
During automated "cmsMake -a" testing, one can sometimes get stuck waiting
for a yes or no answer.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/687)

### DIFF
--- a/cmstaskenv/Test.py
+++ b/cmstaskenv/Test.py
@@ -26,9 +26,10 @@ from __future__ import unicode_literals
 
 import atexit
 import json
-import os
-import sys
 import logging
+import os
+import select
+import sys
 
 import cmscontrib.loaders
 from cms.db import Executable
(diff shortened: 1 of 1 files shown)


karliss:
Wouldn't it be better to have a flag for non interactive mode instead of adding timeout? Also sys.stdin.isatty() can be used to make a guess if no flag is specified.

wil93:
There are flags already (`-y` and `-n`) but sometimes I find myself hitting Enter when testing is in progress, because I forgot to add `-n`. It's not a lot of code and it may be useful (ideally, cmsMake could be rewritten as a ncurses application, so I think it's reasonable to add this kind of features).

The `sys.stdin.isatty()` could be used, but I'm not sure about how, because it seems to return `True` both when calling the script directly and when calling it in a for loop...

stefano-maggiolo:
Forgot to close, I had already merged it.

PR #689, opened 2016-12-22
Select code language

PJeBeK:
it lets us have C++11 language beside C++

This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/689)

### DIFF
--- a/cms/__init__.py
+++ b/cms/__init__.py
@@ -134,8 +134,8 @@ def filename_to_language(filename):
 
     filename (string): the file to test.
 
-    return (list[string]|None): available languages of the extension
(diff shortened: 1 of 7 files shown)


stefano-maggiolo:
Sorry for not communicating enough, but I was already working on a bigger refactoring of how CMS handles programming languages (including the ability of having different versions of the same language available to contestants).

stefano-maggiolo:
For rference, here's the change: https://github.com/stefano-maggiolo/cms/commit/71d23f1e8c68170fafaafff5403a4221e32dd92b

PR #690, opened 2016-12-27
Improve CSRF protection

### LINKED ISSUE #453: Vulnerability of CSRF
Implements in some GET method handlers are used for non-readonly requests, which are vulnerability of CSRF. (uses POST method or validates referer)

e.g.
https://github.com/cms-dev/cms/blob/41ff834621001e8dae9abe4097f291c30d3a3f98/cms/server/admin/templates/task.html#L225

karliss:
### Changes
* Convert remaining state changing GET requests to POST or DELETE.
* Use Tornado CSRF protection to prevent CSRF attacks on POST requests as making requests POST by itself doesn't provide sufficient protection.

Fixes #453 . I couldn't find more state changing GET requests. There is unread message count, but I don't think it's worth the effort protecting that.
I checked all the forms in html templates,  aws_utils.js and tests runned by Travis. Those should be fine, but I am slightly worried that some more unusual requests might have slipped by. Are there any tests not executed by Travis or other js code making requests?

### Important information for all CMS devleopers
After merging these changes all forms making POST requests now need to contain hidden input field containing CSRF token. It can be done by adding `{% module xsrf_form_html() %}` to the form in  HTML template. See this pull request for examples or read http://www.tornadoweb.org/en/stable/guide/security.html#cross-site-request-forgery-protection for more details. All non GET requests without CSRF token will be automatically rejected. Requests done using `CMS.AWSUtils.ajax_delete` or admin/static/web_rpc.js don't need additional changes.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/690)

### DIFF
--- a/cms/server/admin/static/web_rpc.js
+++ b/cms/server/admin/static/web_rpc.js
@@ -21,7 +21,7 @@
 
 function get_cookie(name) {
     var r = document.cookie.match("(^|;)\\s*" + name + "=([^;]*)(;|$)");
-    return r ? r[2] : undefined;
+    return r ? r[2] : void(0);
 }
 
 /**
@@ -42,8 +42,8 @@ function cmsrpc_request(url_root, service, shard, method, args, callback) {
         type: "POST",
         url: url,
         data: JSON.stringify(args),
-        contentType: 'application/json',
-        dataType: 'json',
+        contentType: "application/json",
+        dataType: "json",
         headers: {"X-XSRFToken": get_cookie("_xsrf")}
     });
 
@@ -69,4 +69,4 @@ function cmsrpc_request(url_root, service, shard, method, args, callback) {
     });
 
     return jqxhr;
-};
+}
--- a/cms/server/admin/static/web_rpc.js
+++ b/cms/server/admin/static/web_rpc.js
@@ -18,6 +18,12 @@
 
 "use strict";
 
+
+function get_cookie(name) {
+    var r = document.cookie.match("(^|;)\\s*" + name + "=([^;]*)(;|$)");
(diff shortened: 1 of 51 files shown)


cms/server/admin/static/web_rpc.js
```diff
@@ -18,6 +18,12 @@
 
 "use strict";
 
+
+function get_cookie(name) {
```
karliss: It seems to match the style of file. Is there any JS style guide for CMS?

cms/server/admin/static/web_rpc.js
```diff
@@ -37,7 +43,8 @@ function cmsrpc_request(url_root, service, shard, method, args, callback) {
         url: url,
         data: JSON.stringify(args),
         contentType: 'application/json',
-        dataType: 'json'
+        dataType: 'json',
```
karliss: The same as with camelCase above.

cms/server/admin/static/web_rpc.js
stefano-maggiolo: No, we don't have a JS style (and it shows...)

karliss: In that case I propose #692 .

stefano-maggiolo:
Great PR!

---

Reviewed 33 of 39 files at r1, 4 of 4 files at r2, 6 of 8 files at r3, 6 of 6 files at r4, 2 of 2 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/690#-:-Kb6N8_54HeKdj6i4uIn:bi0afqb)*

PR #693, opened 2017-01-07
Finally here, plugin system for programming language

stefano-maggiolo:
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/693)

### DIFF
--- a/cms/grading/Sandbox.py
+++ b/cms/grading/Sandbox.py
@@ -213,6 +213,7 @@ class SandboxBase(object):
         self.cmd_file = "commands.log"
 
         # These are not necessarily used, but are here for API compatibility
+        # TODO: move all other common properties here.
         self.box_id = 0
         self.fsize = None
         self.cgroup = False
--- a/cms/grading/Sandbox.py
+++ b/cms/grading/Sandbox.py
@@ -3,7 +3,7 @@
 
 # Contest Management System - http://cms-dev.github.io/
 # Copyright © 2010-2015 Giovanni Mascellani <mascellani@poisson.phc.unipi.it>
-# Copyright © 2010-2016 Stefano Maggiolo <s.maggiolo@gmail.com>
+# Copyright © 2010-2017 Stefano Maggiolo <s.maggiolo@gmail.com>
 # Copyright © 2010-2012 Matteo Boscariol <boscarim@hotmail.com>
 # Copyright © 2014 Luca Wehrstedt <luca.wehrstedt@gmail.com>
 #
@@ -195,14 +195,35 @@ class SandboxBase(object):
     EXIT_SYSCALL = 'syscall'
     EXIT_NONZERO_RETURN = 'nonzero return'
 
-    def __init__(self, file_cacher):
+    def __init__(self, file_cacher, temp_dir=None):
         """Initialization.
 
         file_cacher (FileCacher): an instance of the FileCacher class
             (to interact with FS), if the sandbox needs it.
+        temp_dir (unicode|None): temporary directory to use; if None, use the
+            default temporary directory specified in the configuration.
 
         """
         self.file_cacher = file_cacher
+        self.temp_dir = temp_dir if temp_dir is not None else config.temp_dir
+
+        self.cmd_file = "commands.log"
+
+        # These are not necessarily used, but are here for API compatibility
+        self.box_id = 0
+        self.fsize = None
+        self.cgroup = False
+        self.dirs = []
+        self.preserve_env = False
+        self.inherit_env = []
+        self.set_env = {}
+        self.max_processes = 1
+        self.verbosity = 0
+
+        # Set common environment variables.
+        # Specifically needed by Python, that searches the home for
+        # packages.
+        self.set_env["HOME"] = "./"
 
     def get_stats(self):
         """Return a human-readable string representing execution time
@@ -429,20 +450,17 @@ class StupidSandbox(SandboxBase):
 
     """
 
-    def __init__(self, file_cacher=None, temp_dir=None):
+    def __init__(self, file_cacher, temp_dir=None):
         """Initialization.
 
         For arguments documentation, see SandboxBase.__init__.
 
         """
-        SandboxBase.__init__(self, file_cacher)
+        SandboxBase.__init__(self, file_cacher, temp_dir)
 
         # Make box directory
-        if temp_dir is None:
-            temp_dir = config.temp_dir
-        self.path = tempfile.mkdtemp(dir=temp_dir)
+        self.path = tempfile.mkdtemp(dir=self.temp_dir)
 
-        self.cmd_file = "commands.log"
         self.exec_num = -1
         self.popen = None
         self.popen_time = None
@@ -460,23 +478,6 @@ class StupidSandbox(SandboxBase):
         self.timeout = None
         self.wallclock_timeout = None
         self.extra_timeout = None
-
-        # These parameters are not going to be used, but are here for
-        # API compatibility
-        self.box_id = 0
-        self.fsize = None
-        self.cgroup = False
-        self.dirs = []
-        self.preserve_env = False
-        self.inherit_env = []
-        self.set_env = {}
-        self.max_processes = 1
-        self.verbosity = 0
-
-        # Set common environment variables.
-        # Specifically needed by Python, that searches the home for
-        # packages.
-        self.set_env["HOME"] = "./"
 
     # TODO - It returns wall clock time, because I have no way to
     # check CPU time (libev doesn't have wait4() support)
@@ -753,13 +754,13 @@ class IsolateSandbox(SandboxBase):
     """
     next_id = 0
 
-    def __init__(self, file_cacher=None, temp_dir=None):
+    def __init__(self, file_cacher, temp_dir=None):
         """Initialization.
 
         For arguments documentation, see SandboxBase.__init__.
 
         """
-        SandboxBase.__init__(self, file_cacher)
+        SandboxBase.__init__(self, file_cacher, temp_dir)
 
         # Isolate only accepts ids between 0 and 99. We assign the
         # range [(shard+1)*10, (shard+2)*10) to each Worker and keep
@@ -782,9 +783,7 @@ class IsolateSandbox(SandboxBase):
         # system to, which is why the outer directory exists with no read
         # permissions.
         self.inner_temp_dir = "/tmp"
-        if temp_dir is None:
-            temp_dir = config.temp_dir
-        self.outer_temp_dir = tempfile.mkdtemp(dir=temp_dir)
+        self.outer_temp_dir = tempfile.mkdtemp(dir=self.temp_dir)
         # Don't use os.path.join here, because the absoluteness of /tmp will
         # bite you.
         self.path = self.outer_temp_dir + self.inner_temp_dir
@@ -794,7 +793,6 @@ class IsolateSandbox(SandboxBase):
         self.exec_name = 'isolate'
         self.box_exec = self.detect_box_executable()
         self.info_basename = "run.log"   # Used for -M
-        self.cmd_file = "commands.log"
         self.log = None
         self.exec_num = -1
         logger.debug("Sandbox in `%s' created, using box `%s'.",
--- a/cms/grading/Sandbox.py
+++ b/cms/grading/Sandbox.py
@@ -195,15 +195,18 @@ class SandboxBase(object):
     EXIT_SYSCALL = 'syscall'
     EXIT_NONZERO_RETURN = 'nonzero return'
 
-    def __init__(self, file_cacher, temp_dir=None):
+    def __init__(self, multithreaded, file_cacher, temp_dir=None):
         """Initialization.
 
+        multithreaded (boolean): whether the sandbox should allow
+            multithreading.
         file_cacher (FileCacher): an instance of the FileCacher class
             (to interact with FS), if the sandbox needs it.
         temp_dir (unicode|None): temporary directory to use; if None, use the
             default temporary directory specified in the configuration.
 
         """
+        self.multithreaded = multithreaded
         self.file_cacher = file_cacher
         self.temp_dir = temp_dir if temp_dir is not None else config.temp_dir
 
@@ -217,8 +220,12 @@ class SandboxBase(object):
         self.preserve_env = False
         self.inherit_env = []
         self.set_env = {}
-        self.max_processes = 1
         self.verbosity = 0
+
+        self.max_processes = 1
+        if multithreaded:
+            # Max processes is set to 1000 to limit the effect of fork bombs.
+            self.max_processes = 1000
 
         # Set common environment variables.
         # Specifically needed by Python, that searches the home for
@@ -450,13 +457,13 @@ class StupidSandbox(SandboxBase):
 
     """
 
-    def __init__(self, file_cacher, temp_dir=None):
+    def __init__(self, multithreaded, file_cacher, temp_dir=None):
         """Initialization.
 
         For arguments documentation, see SandboxBase.__init__.
 
         """
-        SandboxBase.__init__(self, file_cacher, temp_dir)
+        SandboxBase.__init__(self, multithreaded, file_cacher, temp_dir)
 
         # Make box directory
         self.path = tempfile.mkdtemp(dir=self.temp_dir)
@@ -754,13 +761,13 @@ class IsolateSandbox(SandboxBase):
     """
     next_id = 0
 
-    def __init__(self, file_cacher, temp_dir=None):
+    def __init__(self, multithreaded, file_cacher, temp_dir=None):
         """Initialization.
 
         For arguments documentation, see SandboxBase.__init__.
 
         """
-        SandboxBase.__init__(self, file_cacher, temp_dir)
+        SandboxBase.__init__(self, multithreaded, file_cacher, temp_dir)
 
         # Isolate only accepts ids between 0 and 99. We assign the
         # range [(shard+1)*10, (shard+2)*10) to each Worker and keep
@@ -812,7 +819,6 @@ class IsolateSandbox(SandboxBase):
         self.stack_space = None        # -k
         self.address_space = None      # -m
         self.stdout_file = None        # -o
-        self.max_processes = 1         # -p
         self.stderr_file = None        # -r
         self.timeout = None            # -t
         self.verbosity = 0             # -v
--- a/cms/grading/Sandbox.py
+++ b/cms/grading/Sandbox.py
@@ -761,6 +761,11 @@ class IsolateSandbox(SandboxBase):
     """
     next_id = 0
 
+    # If the command line starts with this command name, we are just
+    # going to execute it without sandboxing, and with all permissions
+    # on the current directory.
+    SECURE_COMMANDS = ["/bin/mv", "/usr/bin/zip", "/usr/bin/unzip"]
+
     def __init__(self, multithreaded, file_cacher, temp_dir=None):
         """Initialization.
 
@@ -958,7 +963,7 @@ class IsolateSandbox(SandboxBase):
         if self.stack_space is not None:
             res += ["--stack=%d" % self.stack_space]
         if self.address_space is not None:
-            res += ["--mem=%d" % self.address_space]
+            res += ["--cg-mem=%d" % self.address_space]
         if self.stdout_file is not None:
             res += ["--stdout=%s" % self.inner_absolute_path(self.stdout_file)]
         if self.max_processes is not None:
@@ -1179,29 +1184,6 @@ class IsolateSandbox(SandboxBase):
         """
         return os.path.join(self.inner_temp_dir, path)
 
-    # XXX - Temporarily disabled (i.e., set as private), in order to
-    # ease interface while implementing other sandboxes, since it's
-    # not used; anyway, it should probably factored through _popen in
-    # order to avoid code duplication
-    def _execute(self, command):
-        """Execute the given command in the sandbox.
-
-        command (list): executable filename and arguments of the
-                        command.
-
-        return (bool): True if the sandbox didn't report errors
-                       (caused by the sandbox itself), False otherwise
-
-        """
-        self.exec_num += 1
-        self.log = None
-        args = [self.box_exec] + self.build_box_options() + ["--"] + command
-        logger.debug("Executing program in sandbox with command: `%s'.",
-                     pretty_print_cmdline(args))
-        with io.open(self.relative_path(self.cmd_file), 'at') as commands:
-            commands.write("%s\n" % (pretty_print_cmdline(args)))
-        return self.translate_box_exitcode(subprocess.call(args))
-
     def _popen(self, command,
                stdin=None, stdout=None, stderr=None,
                close_fds=True):
@@ -1219,8 +1201,38 @@ class IsolateSandbox(SandboxBase):
         return (Popen): popen object.
 
         """
-        self.exec_num += 1
         self.log = None
+        self.exec_num += 1
+
+        # We run a selection of commands without isolate, as they need
+        # to create new files. This is safe because these commands do
+        # not depend on the user input.
+        if command[0] in IsolateSandbox.SECURE_COMMANDS:
+            logger.debug("Executing non-securely: %s at %s",
+                         pretty_print_cmdline(command), self.path)
+            try:
+                prev_permissions = stat.S_IMODE(os.stat(self.path).st_mode)
+                os.chmod(self.path, 0777)
+                with io.open(self.relative_path(self.cmd_file), 'at') as cmds:
+                    cmds.write("%s\n" % (pretty_print_cmdline(command)))
+                p = subprocess.Popen(command, cwd=self.path,
+                                     stdin=stdin, stdout=stdout, stderr=stderr,
+                                     close_fds=close_fds)
+                os.chmod(self.path, prev_permissions)
+                # For secure commands, we clear the output so that it
+                # is not forwarded to the contestants. Secure commands
+                # are "setup" commands, which should not fail or
+                # provide information for the contestants.
+                open(os.path.join(self.path, self.stdout_file), "w").close()
+                open(os.path.join(self.path, self.stderr_file), "w").close()
+                self._write_empty_run_log(self.exec_num)
+            except OSError:
+                logger.critical(
+                    "Failed to execute program in sandbox with command: %s",
+                    pretty_print_cmdline(command), exc_info=True)
+                raise
+            return p
+
         args = [self.box_exec] + self.build_box_options() + ["--"] + command
         logger.debug("Executing program in sandbox with command: `%s'.",
                      pretty_print_cmdline(args))
@@ -1241,6 +1253,14 @@ class IsolateSandbox(SandboxBase):
             raise
 
         return p
+
+    def _write_empty_run_log(self, index):
+        """Write a fake run.log file with no information."""
+        with open(os.path.join(self.path, "run.log.%s" % index), "w") as f:
+            f.write("time:0.000\n")
+            f.write("time-wall:0.000\n")
+            f.write("max-rss:0\n")
+            f.write("cg-mem:0\n")
 
     def execute_without_std(self, command, wait=False):
         """Execute the given command in the sandbox using
--- a/cms/grading/Sandbox.py
+++ b/cms/grading/Sandbox.py
@@ -764,7 +764,7 @@ class IsolateSandbox(SandboxBase):
     # If the command line starts with this command name, we are just
     # going to execute it without sandboxing, and with all permissions
     # on the current directory.
-    SECURE_COMMANDS = ["/bin/mv", "/usr/bin/zip", "/usr/bin/unzip"]
+    SECURE_COMMANDS = ["/bin/cp", "/bin/mv", "/usr/bin/zip", "/usr/bin/unzip"]
 
     def __init__(self, multithreaded, file_cacher, temp_dir=None):
         """Initialization.
--- a/cms/grading/Sandbox.py
+++ b/cms/grading/Sandbox.py
@@ -1212,7 +1212,7 @@ class IsolateSandbox(SandboxBase):
                          pretty_print_cmdline(command), self.path)
             try:
                 prev_permissions = stat.S_IMODE(os.stat(self.path).st_mode)
-                os.chmod(self.path, 0777)
+                os.chmod(self.path, 0700)
                 with io.open(self.relative_path(self.cmd_file), 'at') as cmds:
                     cmds.write("%s\n" % (pretty_print_cmdline(command)))
                 p = subprocess.Popen(command, cwd=self.path,
@@ -1238,7 +1238,7 @@ class IsolateSandbox(SandboxBase):
                      pretty_print_cmdline(args))
         # Temporarily allow writing new files.
         prev_permissions = stat.S_IMODE(os.stat(self.path).st_mode)
-        os.chmod(self.path, 0777)
+        os.chmod(self.path, 0770)
         with io.open(self.relative_path(self.cmd_file), 'at') as commands:
             commands.write("%s\n" % (pretty_print_cmdline(args)))
         os.chmod(self.path, prev_permissions)
--- a/cms/server/contest/handlers/tasksubmission.py
+++ b/cms/server/contest/handlers/tasksubmission.py
@@ -48,7 +48,7 @@ from sqlalchemy.orm import joinedload
 
 from cms import config
 from cms.db import File, Submission, SubmissionResult, Task, Token
-from cms.grading import LANGUAGE_MANAGER
+from cms.grading.languagemanager import get_language
 from cms.grading.scoretypes import get_score_type
 from cms.grading.tasktypes import get_task_type
 from cms.server import actual_phase_required
@@ -552,8 +552,7 @@ class SubmissionFileHandler(FileHandler):
         # decode it to 'foo.%l'.
         stored_filename = filename
         if submission.language is not None:
-            extension = LANGUAGE_MANAGER.get_language(submission.language)\
-                .source_extension
+            extension = get_language(submission.language).source_extension
             stored_filename = re.sub(r'%s$' % extension, '.%l', filename)
 
         if stored_filename not in submission.files:
--- a/cms/server/contest/handlers/tasksubmission.py
+++ b/cms/server/contest/handlers/tasksubmission.py
@@ -3,7 +3,7 @@
 
 # Contest Management System - http://cms-dev.github.io/
 # Copyright © 2010-2014 Giovanni Mascellani <mascellani@poisson.phc.unipi.it>
-# Copyright © 2010-2016 Stefano Maggiolo <s.maggiolo@gmail.com>
+# Copyright © 2010-2017 Stefano Maggiolo <s.maggiolo@gmail.com>
 # Copyright © 2010-2012 Matteo Boscariol <boscarim@hotmail.com>
 # Copyright © 2012-2016 Luca Wehrstedt <luca.wehrstedt@gmail.com>
 # Copyright © 2013 Bernard Blackham <bernard@largestprime.net>
@@ -546,30 +546,27 @@ class SubmissionFileHandler(FileHandler):
         # submissions, yet, if the submission_format changes during the
         # competition, this may not hold anymore for old submissions.
 
-        # filename follows our convention (e.g. 'foo.%l'), real_filename
-        # follows the one we present to the user (e.g. 'foo.c').
-        real_filename = filename
+        # filename is the name used by the browser, hence is something
+        # like 'foo.c' (and the extension is CMS's preferred extension
+        # for the language). To retrieve the right file, we need to
+        # decode it to 'foo.%l'.
+        stored_filename = filename
         if submission.language is not None:
             extension = LANGUAGE_MANAGER.get_language(submission.language)\
                 .source_extension
-            if filename in submission.files:
-                real_filename = filename.replace(".%l", extension)
-            else:
-                # We don't recognize this filename. Let's try to 'undo'
-                # the '%l' -> 'c|cpp|pas' replacement before giving up.
-                filename = re.sub(r'%s$' % extension, '.%l', filename)
+            stored_filename = re.sub(r'%s$' % extension, '.%l', filename)
 
-        if filename not in submission.files:
+        if stored_filename not in submission.files:
             raise tornado.web.HTTPError(404)
 
-        digest = submission.files[filename].digest
+        digest = submission.files[stored_filename].digest
         self.sql_session.close()
 
-        mimetype = get_type_for_file_name(real_filename)
+        mimetype = get_type_for_file_name(filename)
         if mimetype is None:
             mimetype = 'application/octet-stream'
 
-        self.fetch(digest, mimetype, real_filename)
+        self.fetch(digest, mimetype, filename)
 
 
 class UseTokenHandler(BaseHandler):
--- a/cms/server/contest/handlers/tasksubmission.py
+++ b/cms/server/contest/handlers/tasksubmission.py
@@ -46,8 +46,9 @@ import tornado.web
 from sqlalchemy import func
 from sqlalchemy.orm import joinedload
 
-from cms import config, filename_to_language
+from cms import config
 from cms.db import File, Submission, SubmissionResult, Task, Token
+from cms.grading import LANGUAGE_MANAGER
 from cms.grading.scoretypes import get_score_type
 from cms.grading.tasktypes import get_task_type
 from cms.server import actual_phase_required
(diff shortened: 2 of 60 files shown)


stefano-maggiolo:
I'd be glad if any of you volunteers would take a look at this, and possibly try with actual Java submissions, thanks!

For example @karliss @veluca93 @akmohtashami but not only :)

cms/grading/Sandbox.py
```diff
@@ -1215,8 +1201,38 @@ def _popen(self, command,
...
+
+        # We run a selection of commands without isolate, as they need
+        # to create new files. This is safe because these commands do
+        # not depend on the user input.
+        if command[0] in IsolateSandbox.SECURE_COMMANDS:
+            logger.debug("Executing non-securely: %s at %s",
+                         pretty_print_cmdline(command), self.path)
+            try:
+                prev_permissions = stat.S_IMODE(os.stat(self.path).st_mode)
+                os.chmod(self.path, 0777)
```
andreyv: Could this allow any other user on the system to freely modify the directory during this time?

karliss:
Solution download links in CWS are incorrect. They look something like `http://192.168.56.3:8888/tasks/t1/submissions/1/files/t1.C%2B%2B11%20%2F%20g%2B%2B` which decodes to `http://192.168.56.3:8888/tasks/t1/submissions/1/files/t1.C++11 / g++`

karliss:
Custom test source download links also look suspicious `./tasks/t1/tests/1/files/t1.%25l` , but at least they work.

cms/grading/Sandbox.py
stefano-maggiolo: Not really, everything is inside the mkdtemp which is created as 700 (you're also free to set a root dir for temporary files which has restrictions, instead of /tmp). Anyhow, since it actually works with less permission, I'll change to something less scary.

stefano-maggiolo:
Thanks @karliss, fixed the links too.

karliss:
User test download now results in 404.

stefano-maggiolo:
I see, the UserTest and Submission file handlers are behaving differently... <fixing...>

---

Review status: 0 of 58 files reviewed at latest revision, 5 unresolved discussions.

---

*[cms/db/contest.py, line 85 at r9](https://reviewable.io:443/reviews/cms-dev/cms/693#-K_v7CSme-eGYu_lo1MJ-r9-85:-K_zLoxN_CMgdF5W8YDV:b-5wgcc0) ([raw file](https://github.com/cms-dev/cms/blob/2e7821fde164366cfe12c027e54273de7b8e09ad/cms/db/contest.py#L85)):*
<details><summary><i>Previously, veluca93 (Luca Versari) wrote…</i></summary><blockquote>

Any particular reason to have this decided at contest-level instead of at language-level or at task-level? Also, is there any reason not to enable multithreading by default on all compilations?
As I see it, one would need this field in the following two situations:
- there is an allowed language that requires multithreading at runtime for evaluations, in which case enabling multithreading (should/could) be done automatically - possibly with a warning somewhere, or maybe we could have a field that decides if multithreading should be completely disabled (disabling languages that require it), or enabled for languages that require it, or enabled for everything if a language requires it, or enabled globally [that's just an idea]
- A task requires multithreading, in which case it should be enabled at task level (or maybe in the TaskType)

</blockquote></details>

For reference, at the last 2 IOIs it was enabled at the contest level, even if not needed by any task (and obviously only required by Java).

1. Not at the language level because I don't like too much the idea of changing the environment based on the language you use (even if that change is supposedly useless).

2. Task level... I thought about it, then I said to myself, you don't have much time for CMS, it's time to do the essential and not overcomplicate things that don't need to be :) I don't see coming tasks for which multithreaded solutions give an an advantage, so I saw fit having only a switch at the contest level for the technical reasons.

3. Auto enabling in case Java/JDK is a language (and also for compilations): this is a good point, given 1 and 2 it seems it would be the right thing to do (plan: adding a property to the language stating if it requires multithreading).

If nobody objects, I'll go ahead and do that, and also enable for all compilations regardless of language.

---

*[cms/grading/__init__.py, line 100 at r9](https://reviewable.io:443/reviews/cms-dev/cms/693#-K_v7CSn70hHSkvNdd3y-r9-100:-K_zSRZIFCnhX77OUqs_:b-1733qh) ([raw file](https://github.com/cms-dev/cms/blob/2e7821fde164366cfe12c027e54273de7b8e09ad/cms/grading/__init__.py#L100)):*
<details><summary><i>Previously, veluca93 (Luca Versari) wrote…</i></summary><blockquote>

Wouldn't it be better to return all the languages that match the given filename in a list, and then letting the users of LanguageManager choose the one they prefer (or doing something else unrelated), instead of answering something random? Another option could be returning a pair (one_match, True/False) that allows a client to know if the match was unique or not.
If you prefer this version, I would at least guarantee that the language is chosen deterministically across multiple runs.
</blockquote></details>

Yes, I didn't care too much because this is only used in tests and scripts, but I'll change it to be deterministic.

---

*[cms/grading/__init__.py, line 125 at r9](https://reviewable.io:443/reviews/cms-dev/cms/693#-K_v7CSn70hHSkvNdd3y-r9-125:-K_zSvzVRIK6Vo5YcQGx:begzh35) ([raw file](https://github.com/cms-dev/cms/blob/2e7821fde164366cfe12c027e54273de7b8e09ad/cms/grading/__init__.py#L125)):*
<details><summary><i>Previously, veluca93 (Luca Versari) wrote…</i></summary><blockquote>

Why is this method here? It seems to me that it does not make a lot of sense - I see it's only used in the Batch tasktype, and it seems kind of forced to have it here
</blockquote></details>

Makes sense. Changing.

---

*[cms/grading/Job.py, line 59 at r9](https://reviewable.io:443/reviews/cms-dev/cms/693#-K_v7CSme-eGYu_lo1MH-r9-59:-K_zZ39ACoexEq5fARMA:b7u03hh) ([raw file](https://github.com/cms-dev/cms/blob/2e7821fde164366cfe12c027e54273de7b8e09ad/cms/grading/Job.py#L59)):*
<details><summary><i>Previously, veluca93 (Luca Versari) wrote…</i></summary><blockquote>

I could be wrong, but it seems that this has been done :)
</blockquote></details>

Correct

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/693#-:-K_zemJAaiVRw1i8V54B:b-oqz2js)*

lerks:
:lgtm:

---

Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, 12 of 12 files at r4, 47 of 47 files at r5, 9 of 9 files at r6, 1 of 1 files at r7, 2 of 2 files at r8, 2 of 2 files at r9, 1 of 1 files at r10, 2 of 2 files at r11.
Review status: all files reviewed at latest revision, 21 unresolved discussions.

---

*[cms/plugin.py, line 131 at r5](https://reviewable.io:443/reviews/cms-dev/cms/693#-Ka-AwZzn5AcGiSRAQNm:-Ka-Aw_-QS6F6-yGmqca:b-4gdvj9) ([raw file](https://github.com/cms-dev/cms/blob/f60850fb2c02ec346c64e99e46ebefcd253ec2ee/cms/plugin.py#L131)):*
> ```Python
>     classes = []
>     for module in modules:
>         if module.__dict__.get(module.__name__) is not None:
> ```

Can you use `hasattr(module, module.__name__)` here? Same for getattr in the line below.
Also, I would prefer to user foo[...] instead of foo.get(...) once you have established that the attribute exists.

---

*[cms/db/contest.py, line 91 at r4](https://reviewable.io:443/reviews/cms-dev/cms/693#-K_xspq6v7vtzvMJmG3m:-K_xspq6v7vtzvMJmG3n:b1p1yc6) ([raw file](https://github.com/cms-dev/cms/blob/cfa287ead6415ecaa5a033ee1cfd0cd9a6b83728/cms/db/contest.py#L91)):*
> ```Python
>         Boolean,
>         nullable=False,
>         default=True)
> ```

Why on contest instead of dataset?

Need to update the DB schema version number and provide an upgrader.

---

*[cms/db/contest.py, line 85 at r9](https://reviewable.io:443/reviews/cms-dev/cms/693#-K_v7CSme-eGYu_lo1MJ-r9-85:-Ka-C7zM-AWEC4sOfOA0:b-jhc0s5) ([raw file](https://github.com/cms-dev/cms/blob/2e7821fde164366cfe12c027e54273de7b8e09ad/cms/db/contest.py#L85)):*

For reference, at the last 2 IOIs it was enabled at the contest level, even if not needed by any task (and obviously only required by Java).

1. Not at the language level because I don't like too much the idea of changing the environment based on the language you use (even if that change is supposedly useless).

2. Task level... I thought about it, then I said to myself, you don't have much time for CMS, it's time to do the essential and not overcomplicate things that don't need to be :) I don't see coming tasks for which multithreaded solutions give an an advantage, so I saw fit having only a switch at the contest level for the technical reasons.

3. Auto enabling in case Java/JDK is a language (and also for compilations): this is a good point, given 1 and 2 it seems it would be the right thing to do (plan: adding a property to the language stating if it requires multithreading).

If nobody objects, I'll go ahead and do that, and also enable for all compilations regardless of language.
</blockquote></details>

I would prefer 1. The main reason I have against 3 is that then a task is evaluated differently based on what other tasks are in the contest (if I understood correctly). But I could live with that.

---

*[cms/grading/__init__.py, line 85 at r5](https://reviewable.io:443/reviews/cms-dev/cms/693#-Ka-CscqB2l-1Vy3JqCR:-Ka-CscqB2l-1Vy3JqCS:b3rwgcp) ([raw file](https://github.com/cms-dev/cms/blob/f60850fb2c02ec346c64e99e46ebefcd253ec2ee/cms/grading/__init__.py#L85)):*
> ```Python
>             self.header_exts |= set(lang.header_extensions)
>             self.source_exts |= set(lang.source_extensions)
>             self.object_exts |= set(lang.object_extensions)
> ```

Can't all these be class attributes? I don't see any need to have them computed at object creation instead of at class definition...

And actually, it seems this class is essentially empty and used only as a namespace. Wouldn't a module be more suited for that?

---

*[cms/grading/Job.py, line 356 at r2](https://reviewable.io:443/reviews/cms-dev/cms/693#-K_xrQzxlA_BZtHSYNo-:-K_xrQzyu9zVDWAH4FD5:b-zdiptm) ([raw file](https://github.com/cms-dev/cms/blob/4c69c7b85bc0072bd908ae0cdebf297563bf3679/cms/grading/Job.py#L356)):*
> ```Python
> 
>         # dict() is required to detach the dictionary that gets added
>         # to the Job from the control of SQLAlchemy
> ```

This comment applies to the above `manager = dict(...)` line as well. Move it or duplicate it?

---

*[cms/grading/Job.py, line 519 at r2](https://reviewable.io:443/reviews/cms-dev/cms/693#-K_xrwuEEywqdrdpMyWi:-K_xrwuF_hAFJZOizrzU:b9p9njo) ([raw file](https://github.com/cms-dev/cms/blob/4c69c7b85bc0072bd908ae0cdebf297563bf3679/cms/grading/Job.py#L519)):*
> ```Python
>             input=testcase.input,
>             output=testcase.output,
>             info=info
> ```

Why don't you inline the definition? (for consistency with the other cases)

---

*[cms/grading/Job.py, line 583 at r2](https://reviewable.io:443/reviews/cms-dev/cms/693#-K_xsDX3ZKWnGOdXqRIM:-K_xsDX4hrbY5lnIziu-:b-yg2you) ([raw file](https://github.com/cms-dev/cms/blob/4c69c7b85bc0072bd908ae0cdebf297563bf3679/cms/grading/Job.py#L583)):*
> ```Python
> 
>         # dict() is required to detach the dictionary that gets added
>         # to the Job from the control of SQLAlchemy
> ```

Same here (comment is needed earlier).

---

*[cms/grading/Job.py, line 299 at r4](https://reviewable.io:443/reviews/cms-dev/cms/693#-K_xuH2g2_IReDS8R9gP:-K_xuH2g2_IReDS8R9gQ:b8zofos) ([raw file](https://github.com/cms-dev/cms/blob/cfa287ead6415ecaa5a033ee1cfd0cd9a6b83728/cms/grading/Job.py#L299)):*
> ```Python
>             task_type_parameters=dataset.task_type_parameters,
>             language=submission.language,
>             multithreaded_sandbox=multithreaded,
> ```

Why not inline it? Line length concerns? (This comment applies also to the 3 other instances below)

---

*[cms/grading/language.py, line 5 at r5](https://reviewable.io:443/reviews/cms-dev/cms/693#-Ka-Dh9RlKE1EQGT2XIG:-Ka-Dh9RlKE1EQGT2XIH:boy805c) ([raw file](https://github.com/cms-dev/cms/blob/f60850fb2c02ec346c64e99e46ebefcd253ec2ee/cms/grading/language.py#L5)):*
> ```Python
> 
> # Contest Management System - http://cms-dev.github.io/
> # Copyright © 2016 Stefano Maggiolo <s.maggiolo@gmail.com>
> ```

2017 ;)

Also in other files

---

*[cms/grading/language.py, line 126 at r6](https://reviewable.io:443/reviews/cms-dev/cms/693#-Ka-QHaG4JnTDaXE_7wd:-Ka-QHaG4JnTDaXE_7we:b-jum0ev) ([raw file](https://github.com/cms-dev/cms/blob/5690236ce5c88eca495571c78df02d50b171dab4/cms/grading/language.py#L126)):*
> ```Python
>             need to be executable per se).
>         main (string|None): The name of the main file, or none to use
>             executable_filename (this is required by Java).
> ```

args isn't explained

---

*[cms/grading/Sandbox.py, line 480 at r1](https://reviewable.io:443/reviews/cms-dev/cms/693#-K_xplDdjS-7qgEr0h9Z:-K_xplDen6FWooNqOwmb:b-m86abr) ([raw file](https://github.com/cms-dev/cms/blob/f6ab1b4e1a3478e5a7cd93c374ff5d4a767f8425/cms/grading/Sandbox.py#L480)):*
> ```Python
>         self.timeout = None
>         self.wallclock_timeout = None
>         self.extra_timeout = None
> ```

All these parameters are used by both StupidSandbox and IsolateSandbox as well. Any reason you didn't move them to the superclass?

---

*[cms/grading/Sandbox.py, line 767 at r6](https://reviewable.io:443/reviews/cms-dev/cms/693#-Ka-QR2uF7Joaa5LXbbD:-Ka-QR2uF7Joaa5LXbbE:b-8p9xwb) ([raw file](https://github.com/cms-dev/cms/blob/5690236ce5c88eca495571c78df02d50b171dab4/cms/grading/Sandbox.py#L767)):*
> ```Python
>     # going to execute it without sandboxing, and with all permissions
>     # on the current directory.
>     SECURE_COMMANDS = ["/bin/mv", "/usr/bin/zip", "/usr/bin/unzip"]
> ```

Add `/bin/cp`? (used by PHP "compilation")

---

*[cms/grading/languages/haskell_ghc.py, line 80 at r5](https://reviewable.io:443/reviews/cms-dev/cms/693#-Ka-Ex0lfFbVzySjtnsS:-Ka-Ex0mZmFCubYhm1Yo:b-xq7dz7) ([raw file](https://github.com/cms-dev/cms/blob/f60850fb2c02ec346c64e99e46ebefcd253ec2ee/cms/grading/languages/haskell_ghc.py#L80)):*
> ```Python
>     def _capitalize(string):
>         dirname, basename = os.path.split(string)
>         return os.path.join(dirname, basename[0].upper() + basename[1:])
> ```

Would `.capitalize()` work here? (The two are equivalent iff `basename[1:].islower()`)

---

*[cms/grading/languages/python2.py, line 44 at r5](https://reviewable.io:443/reviews/cms-dev/cms/693#-Ka-FT5Zh7fZ0zw7Z9oq:-Ka-FT5Zh7fZ0zw7Z9or:b-ivlcys) ([raw file](https://github.com/cms-dev/cms/blob/f60850fb2c02ec346c64e99e46ebefcd253ec2ee/cms/grading/languages/python2.py#L44)):*
> ```Python
>     def name(self):
>         """See Language.name."""
>         return "Python 2"
> ```

Nit: "Python 2 / CPython", in case we ever add PyPy & friends?

---

*[cms/grading/tasktypes/Batch.py, line 36 at r5](https://reviewable.io:443/reviews/cms-dev/cms/693#-Ka-FkgqBLby5dveCrAm:-Ka-FkgqBLby5dveCrAn:bzehdka) ([raw file](https://github.com/cms-dev/cms/blob/f60850fb2c02ec346c64e99e46ebefcd253ec2ee/cms/grading/tasktypes/Batch.py#L36)):*
> ```Python
> from cms.grading.TaskType import TaskType, \
>     create_sandbox, delete_sandbox
> from cms.grading import LANGUAGE_MANAGER
> ```

Merge with above import?

---

*[cms/grading/tasktypes/Communication.py, line 94 at r5](https://reviewable.io:443/reviews/cms-dev/cms/693#-Ka-KjPJd-MPkQFqA2KB:-Ka-KjPJd-MPkQFqA2KC:b-qpgsmh) ([raw file](https://github.com/cms-dev/cms/blob/f60850fb2c02ec346c64e99e46ebefcd253ec2ee/cms/grading/tasktypes/Communication.py#L94)):*
> ```Python
>                 source_filenames, executable_filename)
>             res[language.name] = commands
>         return res
> ```

I believe this can be turned into a LANGUAGE_MANAGER.get_compilation_commands call

---

*[cmscontrib/loaders/italy_yaml.py, line 461 at r5](https://reviewable.io:443/reviews/cms-dev/cms/693#-Ka-PU-vtwgQsKgPDvZA:-Ka-PU-vtwgQsKgPDvZB:b-cd00q5) ([raw file](https://github.com/cms-dev/cms/blob/f60850fb2c02ec346c64e99e46ebefcd253ec2ee/cmscontrib/loaders/italy_yaml.py#L461)):*
> ```Python
>             # Read grader for each language
>             for lang in LANGUAGE_MANAGER.languages:
>                 extension = lang.source_extensions[0]
> ```

Isn't this just `.source_extension`? (Occurs a few more times below)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/693#-:-Ka-SZ64PxPLjAP_mMOP:bnfp4nl)*

stefano-maggiolo:
I think the only open question now is how to deal with multithreaded_sandbox. I'm still with @veluca93 on making it implicit when the contest, but as a compromise I'm ok to keep it as a toggle as it is now.

---

Review status: 28 of 59 files reviewed at latest revision, 21 unresolved discussions.

---

*[cms/plugin.py, line 131 at r5](https://reviewable.io:443/reviews/cms-dev/cms/693#-Ka-AwZzn5AcGiSRAQNm:-Ka1Cg0ZB_P1mgrhIrFW:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/f60850fb2c02ec346c64e99e46ebefcd253ec2ee/cms/plugin.py#L131)):*

Can you use `hasattr(module, module.__name__)` here? Same for getattr in the line below.
Also, I would prefer to user foo[...] instead of foo.get(...) once you have established that the attribute exists.
</blockquote></details>

Done.

---

*[cms/db/contest.py, line 91 at r4](https://reviewable.io:443/reviews/cms-dev/cms/693#-K_xspq6v7vtzvMJmG3m:-Ka1e-ItOU3VVD9Jwsqj:bdnbm6d) ([raw file](https://github.com/cms-dev/cms/blob/cfa287ead6415ecaa5a033ee1cfd0cd9a6b83728/cms/db/contest.py#L91)):*

Why on contest instead of dataset?

Need to update the DB schema version number and provide an upgrader.
</blockquote></details>

Ack for updater (if we don't remove this), for the rest the discussion is in the comment above

---

*[cms/db/contest.py, line 85 at r9](https://reviewable.io:443/reviews/cms-dev/cms/693#-K_v7CSme-eGYu_lo1MJ-r9-85:-Ka1jXkue5Ifsf3MWWx1:b-rr5ity) ([raw file](https://github.com/cms-dev/cms/blob/2e7821fde164366cfe12c027e54273de7b8e09ad/cms/db/contest.py#L85)):*

I would prefer 1. The main reason I have against 3 is that then a task is evaluated differently based on what other tasks are in the contest (if I understood correctly). But I could live with that.
</blockquote></details>

1 would be allowing it only for Java submissions? I think it was @giomasce originally pushing strongly the idea of same env for all languages, and we would also need to patch the code for the IOI. 

---

*[cms/grading/__init__.py, line 85 at r5](https://reviewable.io:443/reviews/cms-dev/cms/693#-Ka-CscqB2l-1Vy3JqCR:-Ka1zAXO3Pam6crP7bWH:b-yow9eb) ([raw file](https://github.com/cms-dev/cms/blob/f60850fb2c02ec346c64e99e46ebefcd253ec2ee/cms/grading/__init__.py#L85)):*

Can't all these be class attributes? I don't see any need to have them computed at object creation instead of at class definition...

And actually, it seems this class is essentially empty and used only as a namespace. Wouldn't a module be more suited for that?
</blockquote></details>

Make sense. Changed.

---

*[cms/grading/Job.py, line 356 at r2](https://reviewable.io:443/reviews/cms-dev/cms/693#-K_xrQzxlA_BZtHSYNo-:-Ka1eQBGFTTjWenJvxd6:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/4c69c7b85bc0072bd908ae0cdebf297563bf3679/cms/grading/Job.py#L356)):*

This comment applies to the above `manager = dict(...)` line as well. Move it or duplicate it?
</blockquote></details>

Done.

---

*[cms/grading/Job.py, line 519 at r2](https://reviewable.io:443/reviews/cms-dev/cms/693#-K_xrwuEEywqdrdpMyWi:-Ka1eqK_p3SLnyU_xQc6:b-6xhkfx) ([raw file](https://github.com/cms-dev/cms/blob/4c69c7b85bc0072bd908ae0cdebf297563bf3679/cms/grading/Job.py#L519)):*

Why don't you inline the definition? (for consistency with the other cases)
</blockquote></details>

Length.

---

*[cms/grading/Job.py, line 583 at r2](https://reviewable.io:443/reviews/cms-dev/cms/693#-K_xsDX3ZKWnGOdXqRIM:-Ka1ety4JH7NlD9yXy_O:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/4c69c7b85bc0072bd908ae0cdebf297563bf3679/cms/grading/Job.py#L583)):*

Same here (comment is needed earlier).
</blockquote></details>

Done.

---

*[cms/grading/Job.py, line 299 at r4](https://reviewable.io:443/reviews/cms-dev/cms/693#-K_xuH2g2_IReDS8R9gP:-Ka1eNgcv76-0tKtV4tw:b-smaq09) ([raw file](https://github.com/cms-dev/cms/blob/cfa287ead6415ecaa5a033ee1cfd0cd9a6b83728/cms/grading/Job.py#L299)):*

Why not inline it? Line length concerns? (This comment applies also to the 3 other instances below)
</blockquote></details>

Yes, line length concerns.

---

*[cms/grading/language.py, line 5 at r5](https://reviewable.io:443/reviews/cms-dev/cms/693#-Ka-Dh9RlKE1EQGT2XIG:-Ka1evPBi7LWiUhTPM3M:b-jvav09) ([raw file](https://github.com/cms-dev/cms/blob/f60850fb2c02ec346c64e99e46ebefcd253ec2ee/cms/grading/language.py#L5)):*

2017 ;)

Also in other files
</blockquote></details>

You don't know when I worked on it!

---

*[cms/grading/language.py, line 126 at r6](https://reviewable.io:443/reviews/cms-dev/cms/693#-Ka-QHaG4JnTDaXE_7wd:-Ka1f9LAilxi88BvH5Xb:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/5690236ce5c88eca495571c78df02d50b171dab4/cms/grading/language.py#L126)):*

args isn't explained
</blockquote></details>

Done.

---

*[cms/grading/Sandbox.py, line 480 at r1](https://reviewable.io:443/reviews/cms-dev/cms/693#-K_xplDdjS-7qgEr0h9Z:-Ka1fCAlzhjmhXHWAgMw:b-26l6kk) ([raw file](https://github.com/cms-dev/cms/blob/f6ab1b4e1a3478e5a7cd93c374ff5d4a767f8425/cms/grading/Sandbox.py#L480)):*

All these parameters are used by both StupidSandbox and IsolateSandbox as well. Any reason you didn't move them to the superclass?
</blockquote></details>

No specific reasons, I thought the comment below was an "only these are needed for api compatibility". Patches welcome, I'll not complicate this PR more.

---

*[cms/grading/Sandbox.py, line 767 at r6](https://reviewable.io:443/reviews/cms-dev/cms/693#-Ka-QR2uF7Joaa5LXbbD:-Ka1fpVFklJC3vUaTGd-:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/5690236ce5c88eca495571c78df02d50b171dab4/cms/grading/Sandbox.py#L767)):*

Add `/bin/cp`? (used by PHP "compilation")
</blockquote></details>

Done.

---

*[cms/grading/languages/haskell_ghc.py, line 80 at r5](https://reviewable.io:443/reviews/cms-dev/cms/693#-Ka-Ex0lfFbVzySjtnsS:-Ka1fzicKeXogvDVaXLy:b36gfe8) ([raw file](https://github.com/cms-dev/cms/blob/f60850fb2c02ec346c64e99e46ebefcd253ec2ee/cms/grading/languages/haskell_ghc.py#L80)):*

Would `.capitalize()` work here? (The two are equivalent iff `basename[1:].islower()`)
</blockquote></details>

Not relevant to this PR

---

*[cms/grading/tasktypes/Batch.py, line 36 at r5](https://reviewable.io:443/reviews/cms-dev/cms/693#-Ka-FkgqBLby5dveCrAm:-Ka1j2U2JxZ7_jomIHkN:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/f60850fb2c02ec346c64e99e46ebefcd253ec2ee/cms/grading/tasktypes/Batch.py#L36)):*

Merge with above import?
</blockquote></details>

Done.

---

*[cms/grading/tasktypes/Communication.py, line 94 at r5](https://reviewable.io:443/reviews/cms-dev/cms/693#-Ka-KjPJd-MPkQFqA2KB:-Ka1j4_UzSU1iPICH-2Y:b-xf09wz) ([raw file](https://github.com/cms-dev/cms/blob/f60850fb2c02ec346c64e99e46ebefcd253ec2ee/cms/grading/tasktypes/Communication.py#L94)):*

I believe this can be turned into a LANGUAGE_MANAGER.get_compilation_commands call
</blockquote></details>

Following another comment, I removed that method.

---

*[cmscontrib/loaders/italy_yaml.py, line 461 at r5](https://reviewable.io:443/reviews/cms-dev/cms/693#-Ka-PU-vtwgQsKgPDvZA:-Ka1j9YlPi2CWCVoYqkd:b-qvp5ue) ([raw file](https://github.com/cms-dev/cms/blob/f60850fb2c02ec346c64e99e46ebefcd253ec2ee/cmscontrib/loaders/italy_yaml.py#L461)):*

Isn't this just `.source_extension`? (Occurs a few more times below)
</blockquote></details>

Yep, forgot to change everything when I added it. Done.

---

*[cms/grading/languages/python2.py, line 44 at r5](https://reviewable.io:443/reviews/cms-dev/cms/693#-Ka-FT5Zh7fZ0zw7Z9oq:-Ka1ga0UX_-lSGx0mIuW:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/f60850fb2c02ec346c64e99e46ebefcd253ec2ee/cms/grading/languages/python2.py#L44)):*

Nit: "Python 2 / CPython", in case we ever add PyPy & friends?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/693#-:-Ka2-XurRf7hC4boNhAK:b5yjhg)*

veluca93:
Review status: 28 of 59 files reviewed at latest revision, 21 unresolved discussions.

---

*[cms/db/contest.py, line 85 at r9](https://reviewable.io:443/reviews/cms-dev/cms/693#-K_v7CSme-eGYu_lo1MJ-r9-85:-Ka2Vg5HtbJQxRw6uihJ:bpwssya) ([raw file](https://github.com/cms-dev/cms/blob/2e7821fde164366cfe12c027e54273de7b8e09ad/cms/db/contest.py#L85)):*

1 would be allowing it only for Java submissions? I think it was @giomasce originally pushing strongly the idea of same env for all languages, and we would also need to patch the code for the IOI. 
</blockquote></details>

No, in 3 a task would be evaluated differently depending on what languages are enabled for that contest.

Another possible solution would be to have multithreaded_sandbox nullable, where null means "Figure out a sensible value depending on the languages enabled", but I don't think it would make much sense to be allowed to disable threads if an enabled language requires them (while it could make sense the other way round).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/693)*

cms/server/contest/handlers/tasksubmission.py
```diff
@@ -232,46 +234,38 @@ def post(self, task_name):
...
 
-        # If we allow partial submissions, implicitly we recover the
-        # non-submitted files from the previous submission. And put them
-        # in file_digests (i.e. like they have already been sent to FS).
-        submission_lang = None
+        # Read the submission language provided in the request; we
+        # integrate it with the language fetched from the previous
+        # submission (if we use it) and later make sure it is
+        # recognized and allowed.
+        submission_lang = self.get_argument("language", None)
```
wil93: I think the `None` here is redundant, but maybe it's there for readability?

cms/server/contest/handlers/taskusertest.py
```diff
@@ -299,41 +302,37 @@ def post(self, task_name):
         for uploaded, data in self.request.files.iteritems():
             files[uploaded] = (data[0]["filename"], data[0]["body"])
 
+        # Read the submission language provided in the request; we
+        # integrate it with the language fetched from the previous
+        # submission (if we use it) and later make sure it is
+        # recognized and allowed.
+        submission_lang = self.get_argument("language", None)
```
wil93: Also here

lerks:
Reviewed 2 of 2 files at r12, 1 of 1 files at r13, 1 of 1 files at r14, 2 of 2 files at r15, 1 of 1 files at r16, 1 of 1 files at r17, 1 of 1 files at r18, 1 of 1 files at r19, 1 of 1 files at r20, 3 of 3 files at r21, 1 of 1 files at r22, 1 of 1 files at r23, 25 of 25 files at r24.
Review status: all files reviewed at latest revision, 10 unresolved discussions.

---

*[cms/db/contest.py, line 85 at r9](https://reviewable.io:443/reviews/cms-dev/cms/693#-K_v7CSme-eGYu_lo1MJ-r9-85:-Ka3RKbz8GBhNWnmzOFh:bndxng7) ([raw file](https://github.com/cms-dev/cms/blob/2e7821fde164366cfe12c027e54273de7b8e09ad/cms/db/contest.py#L85)):*
<details><summary><i>Previously, veluca93 (Luca Versari) wrote…</i></summary><blockquote>

No, in 3 a task would be evaluated differently depending on what languages are enabled for that contest.

Another possible solution would be to have multithreaded_sandbox nullable, where null means "Figure out a sensible value depending on the languages enabled", but I don't think it would make much sense to be allowed to disable threads if an enabled language requires them (while it could make sense the other way round).
</blockquote></details>

Now I understand option 3 better. It is OK for me, even though I still prefer option 1 (which is to only enable it for Java submissions).

---

*[cms/grading/Sandbox.py, line 480 at r1](https://reviewable.io:443/reviews/cms-dev/cms/693#-K_xplDdjS-7qgEr0h9Z:-Ka3QpBtKp8MqkYxwNdW:bhr1em1) ([raw file](https://github.com/cms-dev/cms/blob/f6ab1b4e1a3478e5a7cd93c374ff5d4a767f8425/cms/grading/Sandbox.py#L480)):*

No specific reasons, I thought the comment below was an "only these are needed for api compatibility". Patches welcome, I'll not complicate this PR more.
</blockquote></details>

OK. Could you add a TODO, though?

---

*[cmscontrib/loaders/italy_yaml.py, line 41 at r24](https://reviewable.io:443/reviews/cms-dev/cms/693#-Ka3Q0cEppfV_9SNimtV:-Ka3Q0cFSZjeBHzfh6V_:b-2nfv0m) ([raw file](https://github.com/cms-dev/cms/blob/e9eb954eea242d665014be0fdc7c821f22a31d5a/cmscontrib/loaders/italy_yaml.py#L41)):*
> ```Python
> from cms.db import Contest, User, Task, Statement, Attachment, \
>     Team, SubmissionFormatElement, Dataset, Manager, Testcase
> from cms.grading import LANGUAGES, HEADER_EXTS
> ```

`from cms.grading.languagemanager`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/693)*

stefano-maggiolo:
Review status: all files reviewed at latest revision, 10 unresolved discussions.

---

*[cms/db/contest.py, line 85 at r9](https://reviewable.io:443/reviews/cms-dev/cms/693#-K_v7CSme-eGYu_lo1MJ-r9-85:-Ka3UFqB6w6AQcCrW6Pq:b-2vtppt) ([raw file](https://github.com/cms-dev/cms/blob/2e7821fde164366cfe12c027e54273de7b8e09ad/cms/db/contest.py#L85)):*

Now I understand option 3 better. It is OK for me, even though I still prefer option 1 (which is to only enable it for Java submissions).
</blockquote></details>

Let's go with 3 then. But I'm not adding a commit right now as it's going to make a mess, let me do it before pushing.

---

*[cms/grading/Sandbox.py, line 480 at r1](https://reviewable.io:443/reviews/cms-dev/cms/693#-K_xplDdjS-7qgEr0h9Z:-Ka3TqRjRgXdutTZMZ81:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/f6ab1b4e1a3478e5a7cd93c374ff5d4a767f8425/cms/grading/Sandbox.py#L480)):*

OK. Could you add a TODO, though?
</blockquote></details>

Done.

---

*[cms/server/contest/handlers/tasksubmission.py, line 241 at r24](https://reviewable.io:443/reviews/cms-dev/cms/693#-K_v7CSn70hHSkvNdd46-r24-241:-Ka3U0v1xKmo9pkVtW6B:bncsdxk) ([raw file](https://github.com/cms-dev/cms/blob/e9eb954eea242d665014be0fdc7c821f22a31d5a/cms/server/contest/handlers/tasksubmission.py#L241)):*

I think the `None` here is redundant, but maybe it's there for readability?
</blockquote></details>

Let's say it's for readability :)

---

*[cms/server/contest/handlers/taskusertest.py, line 309 at r24](https://reviewable.io:443/reviews/cms-dev/cms/693#-K_v7CSn70hHSkvNdd47-r24-309:-Ka3URpt2VHIfRvJO-d2:b2nhb2c) ([raw file](https://github.com/cms-dev/cms/blob/e9eb954eea242d665014be0fdc7c821f22a31d5a/cms/server/contest/handlers/taskusertest.py#L309)):*

Also here
</blockquote></details>

Ack

---

*[cmscontrib/loaders/italy_yaml.py, line 41 at r24](https://reviewable.io:443/reviews/cms-dev/cms/693#-Ka3Q0cEppfV_9SNimtV:-Ka3U4Gx3vS46KCmCSyb:b-hkoz9a) ([raw file](https://github.com/cms-dev/cms/blob/e9eb954eea242d665014be0fdc7c821f22a31d5a/cmscontrib/loaders/italy_yaml.py#L41)):*

`from cms.grading.languagemanager`?
</blockquote></details>

Right... why flake8 doesn't catch this?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/693)*

lerks:
Reviewed 1 of 1 files at r25, 1 of 1 files at r26.
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---

*[cms/db/contest.py, line 91 at r4](https://reviewable.io:443/reviews/cms-dev/cms/693#-K_xspq6v7vtzvMJmG3m:-Ka3VDvLnQlNrW1C5KCP:b510p7c) ([raw file](https://github.com/cms-dev/cms/blob/cfa287ead6415ecaa5a033ee1cfd0cd9a6b83728/cms/db/contest.py#L91)):*

Ack for updater (if we don't remove this), for the rest the discussion is in the comment above
</blockquote></details>

OK, marking this resolved since we're going with option 3 and thus the new field won't happen anymore.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/693)*

stefano-maggiolo:
I just changed the compilation command to collect all *.class (thanks @wil93 for noticing that otherwise inner classes are not functioning properly). He also noted that in arch we might need some more tuning for isolate and java, but we can do that in a follow up. I think this is now ready to be submitted (modulo removing the option for multithreaded sandbox).

Thanks all a lot for the careful review!

---

Review status: 57 of 60 files reviewed at latest revision, 7 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/693#-:-KaESc_1VTP2Iu0D7Zjg:b-k0vkld)*

docs/Installation.rst
```diff
@@ -53,8 +55,9 @@ On Ubuntu 14.10, one will need to run the following script to satisfy all depend
 
 .. sourcecode:: bash
 
-    sudo apt-get install build-essential fpc postgresql postgresql-client \
-         gettext python2.7 iso-codes shared-mime-info stl-manual cgroup-lite
+    sudo apt-get install build-essential openjdk-8-jre openjdk-8-jdk fpc \
```
artikz: Should this also include instructions for Oracle JDK?

wil93: For what it's worth:

> OpenJDK is the official reference implementation of Java SE since version 7.

So I would not include it... In some cases it may be useful to include different implementations (e.g. pypy) but in this case I don't see the point... Also, if I'm not mistaken, it's quite hard to install Oracle JDK >=7 on Ubuntu (you may need to use an external PPA)

artikz: I didn't do any tests, but a lot of contestants believe that Oracle's implementation is more efficient.
Installing it is not hard (https://www.digitalocean.com/community/tutorials/how-to-install-java-with-apt-get-on-ubuntu-16-04).

artikz:
Did I miss it, or there are no instruction on how java solutions should look like? I mean the fact that the class name must be the same as problem name.

docs/Installation.rst
wil93: It requires an external PPA, which is often frowned upon... I think it's best not to require external PPAs in order to install CMS...

wil93: However, I think it's OK to put it with a clear "optional" notice, so that the majority of users (who may not even need Java support) don't end up installing a PPA without needing it

stefano-maggiolo:
Artem, do you mean in CWS for contestants, or in the documentation for admins?

---

Review status: 56 of 60 files reviewed at latest revision, 8 unresolved discussions.

---

*[docs/Installation.rst, line 58 at r9](https://reviewable.io:443/reviews/cms-dev/cms/693#-K_v7CSsiF2Rxzf1XS2R-r9-58:-KaNL7WYXVDvaFsms-jJ:b-lmb0ct) ([raw file](https://github.com/cms-dev/cms/blob/2e7821fde164366cfe12c027e54273de7b8e09ad/docs/Installation.rst#L58)):*

However, I think it's OK to put it with a clear "optional" notice, so that the majority of users (who may not even need Java support) don't end up installing a PPA without needing it
</blockquote></details>

I will add a note that any other sane JDK is fine, and any savvy admin can install their preferred flavour.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/693#-:-KaNL-6ENt_HRrfig3MH:b-cqp0sv)*

lerks:
For other "interface" specifications (e.g., the name of the functions the user has to implement or can call, when managers/stubs are used) we rely on them being explained in the problem statement. This case is slightly different (since it's a constraint enforced by the system, on which the admin has no control) but I think we could solve it the same way. I do agree that it would be good to add it to the doc for admins, though.

---

Review status: 55 of 60 files reviewed at latest revision, 8 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/693#-:-KaPPcCxhFJBr_yt7YFH:b8rqni)*

artikz:
Stefano, the comment about class name was for contestants.

artikz:
We just had a two day contest with 120+ participants with this patch, everything worked perfectly! One thing I noticed though is that cmsSpoolExporter is broken now:

  File "/home/ubuntu/.local/lib/python2.7/site-packages/cmscontrib/SpoolExporter.py", line 104, in do_export
    self.export_submissions()
  File "/home/ubuntu/.local/lib/python2.7/site-packages/cmscontrib/SpoolExporter.py", line 133, in export_submissions
    os.mkdir(submission_dir)
OSError: [Errno 2] No such file or directory: '.../money.1484363551.C++11 / g++'

artikz:
This patch seems to fix the above: https://gist.github.com/artikz/00afdcb4fb9b10e8a864c292d15ebfc2

artikz:
cmsExportSubmissions is also broken the same way

stefano-maggiolo:
Thanks, fixed both (I'm reorganizing for pushing everything, so not going to add them here).

---

Review status: 54 of 60 files reviewed at latest revision, 8 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/693#-:-KaYDEaLKkb2dPo2Fy-8:bu2cj5)*

stefano-maggiolo:
It's over! Thanks all for the contributions!

PR #695, opened 2017-01-18
Update links broken by remove user/contest/task changes

karliss:
Some of the links in recent remove user  49c61f18d3097807c5003513a4071826cf1050f8 commit weren't properly updated resulting in links that lead to 404 .  `ParticipationHandler` was only one with significant changes in path, so assume that similar bug shouldn't affect other links.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/695)


stefano-maggiolo:
Thanks!

PR #702, opened 2017-01-28
Strip executables and use pipes when compiling C/C++

andreyv:
This pull request adds two well-known compiler options to the C/C++ compiler command line: `-s` and `-pipe`.

`-s` tells the compiler to strip symbols from executables. This will significantly reduce the size of compiled executables, which should help with database dumps and network communication. In my tests, a static C++ Hello World program is 2.0 MB unstripped and 1.6 MB stripped (20% size reduction).

`-pipe` tells the compiler to pipe assembly input directly to the assembler instead of using a temporary file. On my computer this gives about 1% compilation speed increase for a C++ Hello World.

I also rearranged the options to follow the usual `<CPPFLAGS> <CFLAGS> <LDFLAGS>` pattern.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/702)

### DIFF
--- a/cms/grading/languages/c++11_g++.py
+++ b/cms/grading/languages/c++11_g++.py
@@ -62,6 +62,6 @@ class Cpp11Gpp(CompiledLanguage):
         command = ["/usr/bin/g++"]
         if for_evaluation:
             command += ["-DEVAL"]
-        command += ["-static", "-O2", "-std=c++11", "-o", executable_filename]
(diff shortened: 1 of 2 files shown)


stefano-maggiolo:
Pipe seems a bit overkill for the kind of work this has to do, but since it shouldn't hurt... :lgtm: 

---

Reviewed 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/702#-:-KbaPrHdMPrYIb7AbDN4:b-wzw185)*

PR #703, opened 2017-01-28
Update Latvian translation

andreyv:
Updated from the new `.pot` and added some additional fixes.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/703)


stefano-maggiolo:
Thanks!

PR #706, opened 2017-01-29
French translation for new strings

smadbe:
French translation for new strings

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/706)


stefano-maggiolo:
Thanks Damien!

PR #708, opened 2017-02-02
Simplified Chinese

vanshady:
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/708)


stefano-maggiolo:
Thanks!

PR #712, opened 2017-02-08
Update Japanese translation.

qnighy:
This PR includes both improvements to the existing Japanese translation and addition of translations of new texts.

- Fix wrapping so that we can easily compare two versions of translations.
- More consistent formatting:
  - Periods: we use `。` if and only if there is a period in the original text.
  - Punctuation symbols: we uniformly use ASCII `!` instead of `！`. We use `:` consistently with the original texts.
  - Spacing: we adjusted translations so that there will be white-spaces between Japanese texts and European texts.
  - Long vowels: there are known ambiguities regarding when to omit `ー`, which indicates a long vowel, at the end of the word. This includes エラー, コンパイラー, サーバー, バッファオーバーフロー, メモリー, ユーザー, ライブラリー. We investigated how these words are used in existing judge systems like AtCoder and AOJ and changed our translations according to the result.
- Terminology and translation:
  - We found free translations were used even if literal translations suffice. We eliminated some of these.
  - More appropriate terminology in context: for example we use `全般的` instead of `一般的`. Another example: we found the translation for `more than %s` didn't fit the actual usage, although this translation would be correct in general contexts. We fixed it to the appropriate one.
  - More brief words for results: our translations for compilation/execution results sounded somewhat verbose. Since these messages are meant to be a brief explanation, we changed our translations from longer sentences to shorter ones or even noun phrases.
- Translations for new texts.

Thanks @wafrelka for reviewing. I hope these changes are helpful to Japanese-speaking users.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/712)


stefano-maggiolo:
Thanks :)

PR #713, opened 2017-02-12
Check r_params != None in case of error.

qnighy:
When one (possibly accidentally) encounters CSRF token mismatch, it is supposed to show a 403 error page. In fact it doesn't. It only shows a blank page.

This seems to be because `r_param` is set to `None` when CSRF token mismatch occurs. In this patch, CMS falls back to the simple error page when `r_param == None`, in addition to the case `r_param` is not defined.

In fact we actually encountered CSRF token mismatch in the beginning of the contest. The cause is still under investigation, but is supposed to be related to caching.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/713)

### DIFF
--- a/cms/server/contest/handlers/base.py
+++ b/cms/server/contest/handlers/base.py
@@ -416,7 +416,7 @@ class BaseHandler(CommonRequestHandler):
         # the data we need to display a basic template with the error
         # information. If r_params is not defined (i.e. something went
         # *really* bad) we simply return a basic textual error notice.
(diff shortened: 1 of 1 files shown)


stefano-maggiolo:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/713#-:-KcxXS8oBYetv9uWHUXx:bnfp4nl)*

PR #722, opened 2017-03-04
Added sign up interface

xdizzaster:
Hello everybody. This PR adds a functionality that I felt like was needed in CMS, namely a way to sign up for a contest without an administrator's help.

When organizing the Romanian Master of Informatics 2016 contest last autumn, we wanted to have an open round, where people can sign up and compete alongside the official contest, using the same interface we use for the contest, but we bumped into a problem. CMS didn't have an official way of letting people sign up. At that time, we hacked together something that would fit the bill, but now the need is coming up again and I've decided to implement the functionality in CMS. 

This PR does just that. It adds a boolean field in the contest table of the DB, storing whether or not people are allowed to sign up, changes the AWS to be able to modify that field and adds a page to the CWS containing the sign up form.

This is my first contribution to CMS, so if I messed something up, I'll try to fix it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ cms.egg-info
 cache/
 lib/
 log/
+.idea/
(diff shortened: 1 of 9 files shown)


stefano-maggiolo:
Definitely appreciated, thanks! It has a few more chunky PR in front of it for inclusion though, so you might need to wait a bit for a merge... sorry about that.

xdizzaster:
No problem about waiting. Thanks for your time!

PR #723, opened 2017-03-10
Detect virtualenv

wil93:
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/723)

### DIFF
--- a/cms/conf.py
+++ b/cms/conf.py
@@ -123,10 +123,12 @@ class Config(object):
         self.pdf_printing_allowed = False
 
         # Installed or from source?
-        self.installed = sys.argv[0].startswith("/usr/") and \
-            sys.argv[0] != '/usr/bin/ipython' and \
-            sys.argv[0] != '/usr/bin/python2' and \
-            sys.argv[0] != '/usr/bin/python'
+        self.installed = (sys.argv[0].startswith("/usr/") and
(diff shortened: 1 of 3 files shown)


stefano-maggiolo:
This still doesn't work with non-installed virtualenv. I'll see if there is a way to save it...

stefano-maggiolo:
What about 
```py
bin_path = os.path.join(os.getcwd(), sys.argv[0])
bin_name = os.path.basename(ran_path)
bin_is_python = bin_name in ["ipython", "python", "python2", "python3"]
bin_in_installed_path = bin_path.startswith(sys.prefix) or (
    hasattr(sys, 'real_prefix') and bin_path.startswith(sys.real_prefix))
self.installed = bin_in_installed_path and not bin_is_python
```

wil93:
> non-installed virtualenv

What is that?

stefano-maggiolo:
All libraries are installed in a virtualenv, CMS is ran from the sources.

(There was a "ran_path" which should have been a "bin_path", changing it in the code above.)

wil93:
> All libraries are installed in a virtualenv, CMS is ran from the sources

Oh, OK... even if I can't really see why one would want to do that :smile:

> What about [...]

LGTM

PR #724, opened 2017-03-11
Added 'Download submissions' button

xdizzaster:
While running the Romanian Master of Informatics 2016 contest last autumn, we wanted to make all the sources submitted public after the contest had ended. That way, everyone can learn and better themselves. Or something along those lines. Now, there was no easy way of doing that, so we eventually dropped the idea. This patch adds that functionality in case anyone ever needs it.

This patch adds:

- A link on the task submission page to a zip file containing the source files in the format `username_X.%l`, where X is the number of the submission, beginning with 1
- The handler code required to gather the sources and create the zip file

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/724)

### DIFF
--- a/cms/server/admin/handlers/task.py
+++ b/cms/server/admin/handlers/task.py
@@ -34,14 +34,20 @@ from __future__ import unicode_literals
 import json
 import logging
 import traceback
+import os
+import shutil
+import tempfile
+import zipfile
 
 import tornado.web
 
+from cms import config
 from cms.db import Attachment, Dataset, Session, Statement, Submission, \
     SubmissionFormatElement, Task
 from cmscommon.datetime import make_datetime
+from cms.grading.languagemanager import get_language
 
-from .base import BaseHandler, SimpleHandler, require_permission
+from .base import BaseHandler, SimpleHandler, FileHandler, require_permission
 
 
 logger = logging.getLogger(__name__)
@@ -502,3 +508,43 @@ class RemoveTaskHandler(BaseHandler):
 
         # Maybe they'll want to do this again (for another task)
         self.write("../../tasks")
+
+
+class DownloadSubmissionsHandler(FileHandler):
+    """Download submissions for a dataset in a zip file.
+    """
+    @require_permission(BaseHandler.AUTHENTICATED)
+    def get(self, task_id):
+        task = self.safe_get_item(Task, task_id)
+
+        zip_filename = "%s.zip" % task.name
+        source_template = "%s_%s%s"
+
+        tempdir = tempfile.mkdtemp(dir=config.temp_dir)
+        zip_path = os.path.join(tempdir, zip_filename)
+
+        users = {}
+
+        with zipfile.ZipFile(zip_path, "w") as zip_file:
+            for submission in task.submissions:
+                user = submission.participation.user.username
+                if user not in users:
+                    users[user] = 1
+                else:
+                    users[user] += 1
+                for _, sub_file in submission.files.items():
+                    source_path = self.application.service.file_cacher.\
+                        get_file(sub_file.digest).name
+                    zip_file.write(
+                        source_path,
+                        source_template % (
+                            submission.participation.user.username,
+                            users[user],
+                            get_language(submission.language).source_extension
+                        )
+                    )
+            zip_file.close()
+
+        self.fetch_from_filesystem(zip_path, "application/zip", zip_filename)
+
+        shutil.rmtree(tempdir)
--- a/cms/server/admin/handlers/__init__.py
+++ b/cms/server/admin/handlers/__init__.py
@@ -69,7 +69,8 @@ from .task import \
     AddAttachmentHandler, \
     AttachmentHandler, \
     TaskListHandler, \
-    RemoveTaskHandler
+    RemoveTaskHandler, \
+    DownloadSubmissionsHandler
 from .dataset import \
     DatasetSubmissionsHandler, \
     CloneDatasetHandler, \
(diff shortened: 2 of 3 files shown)


cms/server/admin/handlers/task.py
```diff
@@ -502,3 +508,43 @@ def delete(self, task_id):
...
+        users = {}
+
+        with zipfile.ZipFile(zip_path, "w") as zip_file:
+            for submission in task.submissions:
+                user = submission.participation.user.username
+                if user not in users:
+                    users[user] = 1
+                else:
+                    users[user] += 1
+                for filename, sub_file in submission.files.items():
```
xdizzaster: Whoops, didn't see that one. I'm not really sure what to replace it with, but the internet says for unused variables, using `_` is ok, so I'll update the PR with that.

cms/server/admin/handlers/task.py
```diff
@@ -502,3 +508,43 @@ def delete(self, task_id):
...
+                    users[user] += 1
+                for _, sub_file in submission.files.items():
+                    source_path = self.application.service.file_cacher.\
+                        get_file(sub_file.digest).name
+                    zip_file.write(
+                        source_path,
+                        source_template % (
+                            submission.participation.user.username,
+                            users[user],
+                            get_language(submission.language).source_extension
```
akmohtashami: I didn't manage to test this yet, but to the best of my knowledge submission.language is None for certain submissions such as submissions to an OutputOnly problem which will make this line to raise a ValueError.

cms/server/admin/handlers/task.py
```diff
@@ -502,3 +508,43 @@ def delete(self, task_id):
...
+
+class DownloadSubmissionsHandler(FileHandler):
+    """Download submissions for a dataset in a zip file.
+    """
+    @require_permission(BaseHandler.AUTHENTICATED)
+    def get(self, task_id):
+        task = self.safe_get_item(Task, task_id)
+
+        zip_filename = "%s.zip" % task.name
+        source_template = "%s_%s%s"
```
akmohtashami: Just wanted to make a suggestion to use a . before the extension so the file actually has that extension (e.g. akm_1.cpp).

xdizzaster: There's no need to. `submission.language` is the file extension of the language, so it gets added anyway.

cms/server/admin/handlers/task.py
xdizzaster: I tested it with an OutputOnly task and it worked fine.

lerks:
I only just looked at this PR, sorry, and I was wondering if you had considered adding a script (like the ones found in cmscontrib/) to implement this functionality rather than adding it in AWS. In my opinion it would work better for the use case you described, as outputting files would be easier (no need to zip them!) and it's easier to add parameters to fine-tune the behavior. Moreover, I _personally_ consider AWS to be the main tool to manage the contest while it is underway, and the scripts to be the best tools to put data in CMS before the contest and take data out after the contest.

akmohtashami:
Actually, in my opinion, it would be nicer to have all functionalities available in the AWS as well. Having scripts is a great thing specially when you are using bash to run some custom actions. However for normal usage it's much easier to do something using a web interface.

xdizzaster:
I agree with @akmohtashami. The AWS is, IMHO, the obvious place to add such functionality, as that is the place where you browse contestant's submissions. That does not mean however that there couldn't be a script to replicate the functionality in the command line for more complex scenarios (I'm imagining something like extracting all the sources written in a specific language, or submitted after a certain time, or with a score above a certain threshold, or a combination of the above and more). I'll look into implementing such a script if people think it could be useful.

lerks:
I think that AWS is the primary tool to monitor a contest while it is underway and to modify it in emergency situations (reevaluating a submission, changing a broken testcase, changing a contestant's IP address because they moved, ...). Adding functionalities that aren't indispensable in such a setting would add clutter and make handling an emergency more complicated (one needs to find the right section, the right button, mistakes could happen, ...). I'm not against adding this feature to AWS if it stays clear of these critical paths, but I feel it would be safer to put it in a script.

wil93:
I am not sure I understood the use case. Shouldn't it be more sensible to
put an optional download button in RankingWebServer?

Il mar 28 mar 2017, 15:34 Luca Wehrstedt <notifications@github.com> ha
scritto:

xdizzaster:
@wil93 I thought it fits better in AWS as it gives the contest administrators more flexibility. Maybe they don't want to give away the sources. Maybe they want to download the sources in bulk and run them through a system like [Moss](https://theory.stanford.edu/~aiken/moss/) to scan for plagiarism in the case of online contests.

@lerks I understand the concern about cluttering the AWS with too many useless buttons and links. Right now, the patch adds a link just above the sources list of a task, as seen in the screenshot, highlighted with red

admin - google chrome_001

It could be moved below the list if folks think that's not a good place.

wil93:
Oh OK now I got it. I thought the use case was to give away the sources (e.g. google code jam).

I tend to agree with @lerks then... For this use case there is `cmsExportSubmissions` already (and I actually use it to export all subs and them scan them with a plagiarism detection system), so I don't think it's worth to basically create a web interface upon an existing script.

wil93:
By the way, from the original post of this PR it actually seems that my interpretation (a button on every submission, in RankingWebServer) would fit the desired behavior better 😕

lerks:
Oh, sorry, I wasn't aware we already had `cmsExportSubmissions`, my bad. @xdizzaster do you think that would be enough for the use case you originally proposed? If so, we should probably avoid implementing redundant functionalities and close this. If not, would it be possible to improve `cmsExportSubmissions` rather than duplicating parts of it?

xdizzaster:
I wasn't aware of that script either (sorry). I guess that makes this PR redundant, so I'm just gonna close it. Thanks to everyone for their time.

PR #725, opened 2017-03-18
Fix outputonly submission regression caused by e16ad23.

qnighy:
As the title shows.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/725)


stefano-maggiolo:
Thanks, merged also in v1.3

PR #727, opened 2017-03-18
Upgrade python deps

stefano-maggiolo:
Not sure that there are no side effects, but many of these are 2y old and when would be a better time than right at the start of a release?

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/727)

### DIFF
--- a/cms/db/session.py
+++ b/cms/db/session.py
@@ -31,6 +31,7 @@ from __future__ import absolute_import
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import logging
 import psycopg2
 
 from sqlalchemy.orm import sessionmaker, scoped_session
@@ -39,6 +40,9 @@ from sqlalchemy.engine.url import make_url
 from cms import config
 
 from . import engine
+
+
+logger = logging.getLogger(__name__)
 
 
 Session = sessionmaker(engine, twophase=config.twophase_commit)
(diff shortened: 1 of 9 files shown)


lerks:
Fix travis and it'll :lgtm:

---

Reviewed 9 of 9 files at r1.

---

*[requirements.txt, line 14 at r1](https://reviewable.io:443/reviews/cms-dev/cms/727#-KfYsZMK-ZtrrcEuR4Ej:-KfYsZMK-ZtrrcEuR4Ek:b5sas1j) ([raw file](https://github.com/cms-dev/cms/blob/11b2ad22c6c5d3f0c4c9d3e698f64bc58264cde5/requirements.txt#L14)):*
> ```
> patool==1.12  # https://github.com/wummel/patool/blob/master/doc/changelog.txt
> bcrypt==3.1.3  # https://github.com/pyca/bcrypt/
> chardet==2.3.0  # https://pypi.python.org/pypi/chardet
> ```

Great idea to put references to the changelogs!

One remark: why pin down the micro version so specifically? We will not be regularly checking for updates on these packages and modify this file to keep up. I suggest to use a format like `>=4.4,<4.5` (say, for tornado), like we already had for alchemy. This way we ensure that we don't get to a higher major or minor version (which may introduce API differences) but at least we benefit from bug-fixing micro updates.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/727#-:-KfYt_9IT86xOTfY3n9J:b-6d7r5g)*

PR #728, opened 2017-03-18
RS refactoring

stefano-maggiolo:
Mostly to get better performance when some services are not running, but I think the structure is also nicer.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/728)

### DIFF
--- a/cms/service/ResourceService.py
+++ b/cms/service/ResourceService.py
@@ -3,7 +3,7 @@
 
 # Contest Management System - http://cms-dev.github.io/
 # Copyright © 2010-2014 Giovanni Mascellani <mascellani@poisson.phc.unipi.it>
-# Copyright © 2010-2014 Stefano Maggiolo <s.maggiolo@gmail.com>
+# Copyright © 2010-2017 Stefano Maggiolo <s.maggiolo@gmail.com>
 # Copyright © 2010-2012 Matteo Boscariol <boscarim@hotmail.com>
 # Copyright © 2013-2016 Luca Wehrstedt <luca.wehrstedt@gmail.com>
 # Copyright © 2014 William Di Luigi <williamdiluigi@gmail.com>
@@ -33,10 +33,12 @@ from __future__ import unicode_literals
 import bisect
 import logging
 import os
+import re
 import time
 
 import psutil
 
+from collections import defaultdict
 from gevent import subprocess
 try:
     from subprocess import DEVNULL  # py3k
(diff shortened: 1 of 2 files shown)


lerks:
:lgtm:

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[cms/service/ResourceService.py, line 84 at r1](https://reviewable.io:443/reviews/cms-dev/cms/728#-KfaqEYeAUx7-VRq2MKY:-KfaqEYft-HIlidFyytY:b-hi4omz) ([raw file](https://github.com/cms-dev/cms/blob/1717ce722f4bd4f94603eb30f5afbf9412ff3b28/cms/service/ResourceService.py#L84)):*
> ```Python
> 
>         procs ({string: {int|None}: object): maps tuples (argument
>             with service name, shard) to psutil proc_info objects.
> ```

procs isn't an argument, is this a leftover? also, you should explain cpu_times.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/728#-:-Kfasw8gjsaIQNUr4rHC:bnfp4nl)*

lerks:
:lgtm_cancel: 

I'd like to try a totally different approach for RS (using lockfiles instead of parsing command lines) and I'm putting this PR on hold until that's done so that we can compare the two approaches and push the one we prefer.

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/728#-:-KfjX_J0ZYDEsnTQDQHJ:b-3w3xvl)*

stefano-maggiolo:
I disagree, we should merge this improvement at zero cost and if it becomes a need try different approaches. What you say might be better, but it's significantly different and creates some work that by merging this we won't have.

lerks:
:lgtm:

You're right, agreed (and sorry). LGTM again, go ahead! :)

---

Reviewed 1 of 2 files at r1.
Review status: 1 of 2 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/728#-:-KfkOtEJvqOz2n2AIpUm:b-9u1j2)*

stefano-maggiolo:
Review status: 1 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[cms/service/ResourceService.py, line 84 at r1](https://reviewable.io:443/reviews/cms-dev/cms/728#-KfaqEYeAUx7-VRq2MKY:-Kfs4jfAxLO9jVkO_Lbs:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/1717ce722f4bd4f94603eb30f5afbf9412ff3b28/cms/service/ResourceService.py#L84)):*

procs isn't an argument, is this a leftover? also, you should explain cpu_times.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/728)*

PR #729, opened 2017-03-21
Fix misuse of gevent's StreamServer

lerks:
StreamServer calls the handler we provided to it for every new incoming connection, giving it the socket for that connection. That call is performed in its own greenlet. Once that handler terminates, StreamServer considers the socket as dealt with, and therefore ready for closure and elimination. Yet, in gevent up to 1.0, the socket was left untouched by StreamServer. Starting in gevent 1.1 StreamServer closes the socket once the handler returns.

Our handler used to spawn another greenlet, inside of which it was running an infinite loop that read from the socket and took care of executing the requests. Once it spawned that greenlet, the handler would return. This worked correctly in gevent 1.0 but breaks now, as the socket will be closed immediately after the reading loop starts.

This causes the client to get into a loop of connection, immediate disconnection and reattempt.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/729)

### DIFF
--- a/cms/io/rpc.py
+++ b/cms/io/rpc.py
@@ -5,7 +5,7 @@
 # Copyright © 2010-2013 Giovanni Mascellani <mascellani@poisson.phc.unipi.it>
 # Copyright © 2010-2015 Stefano Maggiolo <s.maggiolo@gmail.com>
 # Copyright © 2010-2012 Matteo Boscariol <boscarim@hotmail.com>
-# Copyright © 2013 Luca Wehrstedt <luca.wehrstedt@gmail.com>
+# Copyright © 2013-2017 Luca Wehrstedt <luca.wehrstedt@gmail.com>
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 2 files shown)


stefano-maggiolo:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---

*[cmstestsuite/unit_tests/io/rpc_test.py, line 120 at r1](https://reviewable.io:443/reviews/cms-dev/cms/729#-KfiCLlYTo4iqI30doMy:-KfiCLlYTo4iqI30doMz:bju2g3i) ([raw file](https://github.com/cms-dev/cms/blob/0d39c949d4c443370264db369620a9c5e40ad5e8/cmstestsuite/unit_tests/io/rpc_test.py#L120)):*
> ```Python
>         socket_ (socket): the socket to use
>         address (tuple): the (ip address, port) of the remote part
>         return (RemoteServiceServer): a server
> ```

Remove this (and maybe rename the method?)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/729#-:-KfiCROts1G31vRwOqtP:bnfp4nl)*

lerks:
Review status: 1 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[cmstestsuite/unit_tests/io/rpc_test.py, line 120 at r1](https://reviewable.io:443/reviews/cms-dev/cms/729#-KfiCLlYTo4iqI30doMy:-KfiD_az2uFs9EYyB21d:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/0d39c949d4c443370264db369620a9c5e40ad5e8/cmstestsuite/unit_tests/io/rpc_test.py#L120)):*

Remove this (and maybe rename the method?)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/729)*

PR #731, opened 2017-03-25
Add Rust as a supported language

dariost:
This PR adds support for the Rust programming language

Note that the Rust compiler is available in Ubuntu since version 16.04 and TravisCI use Ubuntu 14.04, so I don't know how to make TravisCI pass the tests for Rust

I already tried installing the Rust compiler using [rustup](https://github.com/rust-lang-nursery/rustup.rs), but it gets installed in the user home, so it becomes a mess to whitelist it with isolate

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/731)


dariost:
I've tried running the test suite on Ubuntu 16.04 with the stock Ubuntu 16.04 Rust compiler and I can confirm that everything works

stefano-maggiolo:
Thanks Dario, but I cannot help but wonder where is the line for the languages we want to include in the repo, given that adding them is now quite easy. (You could argue that we have PHP, so the bar is pretty low  right now...).

I'd be inclined to add this without functional tests - also from a test  runtime perspective, if not only because of the complex travis integration.

PS: no need to update the updater.

dariost:
I removed the commit to update the updaters

As for the language(s), I'm planning to add support for some of the languages supported by [Codeforces](http://codeforces.com/) (Rust being one of them), so this is where I'm drawing the line (if you think otherwise let me know)

As for the functional tests I think maybe it's better to add them and add some kind of blacklist for `cmsRunTests` to prevent them from running in the TravisCI build (but keep them in cms because someone might want to run them to see if the Rust environment is working correctly)

dariost:
I managed to resolve the issue with TravisCI, now this PR passes all checks

lerks:
Another option would be to add the language to the repo, with all the tests, but then change the testsuite to test by default (i.e., when no explicit `-l` parameter is passed) only on a small subset of the available languages. This subset, IMO, should be the languages used at the IOI. What do you think?

dariost:
@lerks proposal sounds good to me, but I think this should be done in another PR since it's not strictly related to Rust support.

stefano-maggiolo:
Uh, I expected downloading and installing to be much slower, but it's actually ok. I might get around in implementing a switch so that by default it only uses tests for a small set of languages (e.g., only C++), but I might also leave it as it is. Will merge with minor editing.

stefano-maggiolo:
Other nit: commit comments are usually in imperative form :)

---

Reviewed 13 of 13 files at r11, 1 of 1 files at r12, 1 of 1 files at r13, 1 of 1 files at r14, 1 of 1 files at r15.

---

*[cms/grading/languages/rust.py, line 52 at r9](https://reviewable.io:443/reviews/cms-dev/cms/731#-KrcCHclU4rMMSyN9SSO:-KrcCHclU4rMMSyN9SSP:b-45cz0a) ([raw file](https://github.com/cms-dev/cms/blob/61de5f14660ea9a9936694579f0b8f42f6472e6d/cms/grading/languages/rust.py#L52)):*
> ```Python
>     @property
>     def header_extensions(self):
>         """See Language.source_extensions."""
> ```

Nit: here and later comments are wrong.

---

*[cms/grading/languages/rust.py, line 69 at r9](https://reviewable.io:443/reviews/cms-dev/cms/731#-KrcCpgohI3teVPKyHx1:-KrcCpgohI3teVPKyHx2:b-w7t2d2) ([raw file](https://github.com/cms-dev/cms/blob/61de5f14660ea9a9936694579f0b8f42f6472e6d/cms/grading/languages/rust.py#L69)):*
> ```Python
>         command.append(["/usr/bin/rustc", "-O", "-o",
>                          executable_filename, source_filenames[0]])
>         return command
> ```

Can be simplified heavily.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/731#-:-KrcDLS4Z0zrEIEo44T8:b-sayv08)*

stefano-maggiolo:
Merged

PR #732, opened 2017-03-26
Update Lithuanian translation

vytisb:
I've updated the Lithuanian translation.
Hopefully not too late for v1.3.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/732)


stefano-maggiolo:
Definitely not :> thanks!

Issue #736, opened 2017-03-31
Possible bug in smartmappedcollection


wil93:
@lerks any ideas?

In the code below, `t` is an instance of Task.

SQLAlchemy version used is >1.0,<1.1 as required by CMS.

```
....
  File "/home/lversari/cmsocial/cmsocial/server/pws.py", line 1057, in task_handler
    dict([(l, s.digest) for l, s in t.statements.iteritems()])
  File "/usr/lib/python2.7/dist-packages/sqlalchemy/orm/attributes.py", line 237, in __get__
    return self.impl.get(instance_state(instance), dict_)
  File "/usr/lib/python2.7/dist-packages/sqlalchemy/orm/attributes.py", line 599, in get
    return self.set_committed_value(state, dict_, value)
  File "/usr/lib/python2.7/dist-packages/sqlalchemy/orm/attributes.py", line 1074, in set_committed_value
    collection, user_data = self._initialize_collection(state)
  File "/usr/lib/python2.7/dist-packages/sqlalchemy/orm/attributes.py", line 986, in _initialize_collection
    self.dispatch.init_collection(state, collection, adapter)
  File "/usr/lib/python2.7/dist-packages/sqlalchemy/event/attr.py", line 256, in __call__
    fn(*args, **kw)
  File "/usr/lib/python2.7/dist-packages/sqlalchemy/orm/events.py", line 1606, in wrap
    fn(target, value, *arg)
  File "/home/lversari/cms/cms/db/smartmappedcollection.py", line 46, in cb_on_init
    coll.on_init()
  File "/home/lversari/cms/cms/db/smartmappedcollection.py", line 146, in on_init
    assert not self.linked
AssertionError
```

lerks:
No ideas from just looking at that.

Is it deterministic? Does it happen every time you run that code? Did you do any change that started triggering it? Did you update the micro version of SA?

I think that even though SMC does completely legal things with the SA API our use case isn't really what the devs had in mind and thus it's always a mess. I could start debugging this again, or we could get rid of SMC entirely. It was never really necessary, it was mostly to prevent some pitfalls when using SA collections (and it was a nice excuse to learn more about the internals of SA). I believe nothing would break if we were to replace SMC with a regular collection.

What's the general opinion around this?

wil93:
> Is it deterministic? Does it happen every time you run that code? Did you do any change that started triggering it? Did you update the micro version of SA?

I will let @veluca93 reply

[...]

Issue #726, opened 2017-03-18
Submissions not moving past Compiling


rcabreu:
Sample output from LogService:
```
19:00:08,569 - INFO [Evaluation,0] Operations `compile on 15 against dataset 2' and 0 more concluded successfully.
2017-03-18 19:00:08,569 - INFO [Worker,2] Starting job group.
2017-03-18 19:00:08,570 - INFO [Worker,2] [compile submission 15] Starting job.
2017-03-18 19:00:09,740 - INFO [Worker,2] [compile submission 15] Finished job.
2017-03-18 19:00:09,741 - INFO [Worker,2] Finished job group.
2017-03-18 19:00:09,741 - INFO [Worker,2] Executed in 1.172 after free for 5.903; busyness is 54.0%; avg free time is 2.962 avg busy time is 3.477
2017-03-18 19:00:09,742 - INFO [Evaluation,0] `compile on 15 against dataset 2' completed. Success: True.
2017-03-18 19:00:56,607 - INFO [Proxy,0] Operations `sending data of type 0 to ranking' and 3 more concluded successfully.
```

As we can see, submissions are being judged. However, contestants just see them i "Compiling" stage. 
HELP!! Using the newly released cms-1.3.rc0

stefano-maggiolo:
@rcabreu what do you see in AWS? From the logs it looks like it's a compilation failure, do you know if the submitted code compiles?

rcabreu:
@stefano-maggiolo the code we sent compiles. All submissions are experiencing this.
Could we chat (Hangout)? rcabreu@gmail.com

stefano-maggiolo:
Sent invitation

ABeltramo:
Same problem here!
```
log_1  | 2017-03-22 11:11:20,488 - INFO [Contest,0] 302 POST /tasks/baht/submit (172.19.0.1) 134.41ms
log_1  | 2017-03-22 11:11:20,522 - INFO [Evaluation,0] Asking worker 0 to `compile on 1 against dataset 1'.
log_1  | 2017-03-22 11:11:20,553 - INFO [Evaluation,0] Operations `compile on 1 against dataset 1' and 0 more concluded successfully.
log_1  | 2017-03-22 11:11:20,554 - INFO [Worker,0] Starting job group.
log_1  | 2017-03-22 11:11:20,554 - INFO [Worker,0] [compile submission 1] Starting job.
[...]

Issue #666, opened 2016-10-19
Error while adding a test case to a task


shivam312:
While adding the test case to the task i get an error. After submitting it shows error 500. And any page related to the task shows the same.
And the following is printed on the command line  : 

```
`2016/10/19 03:57:49 - ERROR [AdminWebServer,0] Uncaught exception GET /task/7 (172.16.193.1)
HTTPServerRequest(protocol='http', host='172.16.193.161:8889', method='GET', uri='/task/7', version='HTTP/1.1', remote_ip='172.16.193.1', headers={'Accept-Language': 'en-US,en;q=0.8', 'Accept-Encoding': 'gzip, deflate, sdch', 'If-None-Match': '"89a26cee227ab04e8c6bd7e32ccc408b97203e7f"', 'Connection': 'keep-alive', 'Accept': 'text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,_/_;q=0.8', 'Upgrade-Insecure-Requests': '1', 'Host': '172.16.193.161:8889', 'Referer': 'http://172.16.193.161:8889/add_testcase/7', 'Cache-Control': 'max-age=0', 'Cookie': 'login="2|1:0|10:1476730150|5:login|52:KFZBQQpwMApWQUEKcDEKRjE0NzY3MzAxNTAuODgwNDE0CnRwMgou|6085a23c95cba3a28993585825af5c9143d23616d61347cb703e8ce20274fab5"; unread_count="2|1:0|10:1476730286|12:unread_count|4:MA==|d5bb94c258684df0cbf4e153f8466a7d3d998f336514d7170a0a7cb27b1c4f2d"', 'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/53.0.2785.116 Safari/537.36'})
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/tornado-4.1-py2.7-linux-x86_64.egg/tornado/web.py", line 1346, in _execute
 result = method(_self.path_args, *_self.path_kwargs)
  File "/usr/local/lib/python2.7/dist-packages/cms-1.2.0-py2.7.egg/cms/server/AdminWebServer.py", line 1506, in get
 self.render("task.html", **self.r_params)
  File "/usr/local/lib/python2.7/dist-packages/tornado-4.1-py2.7-linux-x86_64.egg/tornado/web.py", line 664, in render
[...]

PR #737, opened 2017-04-02
Get rid of SmartMappedCollection

lerks:
The troubles it caused far exceeded its benefits.

Addresses #736 and possibly #726 and #666.

(Negative diff is best diff.)

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/737)


lerks:
PS: I didn't do any manual testing; just ran the test suite.

stefano-maggiolo:
Looks fine, IIUC the reason why we have so few changed lines is that we never move objects between different parents, and that to add children we usually use the pattern of creating the parent, then creating the children objects directly (since we don't use them in the rest of the session anyway). Correct? 

---

Review status: 0 of 6 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/737#-:-KgmqnON42fhSQixXYu1:bz4mi16)*

lerks:
Yes, quite. Also, the main deficiency of SA's mapped_collection that SMC tried to address is the inability to respond to child objects changing their value on the column that acts as key. Apparently, though, we never did that or it never bit us.

---

Review status: 0 of 6 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/737#-:-KgoBAgzGzZAZ8X1eNmJ:b5hhcsm)*

stefano-maggiolo:
One more thing, can we do these with column_mapped_collection?

---

Review status: 0 of 6 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/737#-:-Kgo_0HzKiWHklfezzwL:b87pasq)*

lerks:
Apparently not. I tried and Alchemy fails, saying that it expected a `Column`-based argument but got a `unicode` instead. In fact, the documentation specifies that "`column_mapped_collection` [...] is almost like `attribute_mapped_collection` except given the `Column` object directly". We cannot give the `Column` object, as we haven't built it yet. Unless you have other ideas on how to do this I'm sticking to `attribute_mapped_collection`.

PS: I'm still waiting for your review and LGTM...

---

Review status: 0 of 6 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/737#-:-KgrGn3Nf21bj8XTmPiS:b-6rum50)*

stefano-maggiolo:
:lgtm:

---

Review status: 0 of 6 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/737#-:-KgrJI17k3mtSnTp1iO9:bnfp4nl)*

Issue #157, opened 2013-06-01
Add IPv6 support


giomasce:
At the moment CMS supports only IPv4, but the Internet is (very slowly) moving towards a new protocol: IPv6! Do we really want to stay behind?

lerks:
The only outstanding issue for IPv6 support that appears here is being addressed in #738. I believe we can close this after that PR is merged.

edomora97:
IPv6 doesn't really work in a network IPv6 only. If the services are accessible only via IPv6 there are many errors in the RPC.
The problem we noticed is caused by https://github.com/cms-dev/cms/blob/bfb359ba21935785b47ecb21b86ff4fcdb1f3522/cms/io/rpc.py#L481

We patched cms to make it work during the night, maybe a better fix should be implemented soon, for now here it's the diff from https://github.com/cms-dev/cms/commit/279c0fd9acfe16bf004267b7d89ceb9c3a0190cd

```diff
diff --git a/cms/io/rpc.py b/cms/io/rpc.py
index 9ec358b0..a8d42217 100644
--- a/cms/io/rpc.py
+++ b/cms/io/rpc.py
@@ -20,6 +20,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import functools
+import ipaddress
 import json
 import logging
 import socket
@@ -478,7 +479,9 @@ class RemoteServiceClient(RemoteServiceBase):
 
         """
         try:
-            sock = gevent.socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            info = socket.getaddrinfo(self.remote_address.ip, self.remote_address.port)
+            info = info[0]
+            sock = gevent.socket.socket(info[0], socket.SOCK_STREAM)
             sock.connect(self.remote_address)
         except OSError as error:
             logger.debug("Couldn't connect to %s: %s.",
diff --git a/cms/io/service.py b/cms/io/service.py
index 942319e3..63be02d3 100644
--- a/cms/io/service.py
+++ b/cms/io/service.py
@@ -26,6 +26,7 @@ using gevent and JSON encoding.
 
 import errno
 import functools
+import ipaddress
 import logging
 import os
 import pwd
@@ -97,7 +98,8 @@ class Service:
                               "Is it specified in core_services in cms.conf?" %
                               (self._my_coord,))
 
-        self.rpc_server = StreamServer(address, self._connection_handler)
+        info = socket.getaddrinfo(address.ip, address.port)
[...]

PR #738, opened 2017-04-02
Add in-database check constraints to validate some fields

### LINKED ISSUE #582: Downloading tests with / in codename fails

lerks:
Make PostgreSQL perform some constraint validation on some fields.

Fixes #582 and closes #584.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/738)

### DIFF
--- a/cms/db/validation.py
+++ b/cms/db/validation.py
@@ -96,15 +96,3 @@ class DigestConstraint(ValidationConstraint):
     def get_condition(column):
         return column.op("~")("^([0-9a-f]{40}|%s)$"
                               % FileCacher.TOMBSTONE_DIGEST)
-
-
-class IPv4Constraint(ValidationConstraint):
-    """Check that the column is a valid IPv4 address (plus subnet?)."""
-    RE255 = "[0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5]"
-    RE32 = "[0-9]|[1-2][0-9]|3[0-2]"
-
-    @staticmethod
-    def get_condition(column):
-        return column.op("~")(
-            "^(%(re255)s)\.(%(re255)s)\.(%(re255)s)\.(%(re255)s)(/(%(re32)s))?$"
-            % {"re255": IPv4Constraint.RE255, "re32": IPv4Constraint.RE32})
--- a/cms/db/validation.py
+++ b/cms/db/validation.py
@@ -26,6 +26,8 @@ from sqlalchemy.event import listens_for
 from sqlalchemy.orm import mapper
 from sqlalchemy.schema import CheckConstraint
 
+from cms.db.filecacher import FileCacher
+
 
 class ValidationConstraint(object):
     """Constraint adding server-side validation to a SQLAlchemy Column.
@@ -92,7 +94,8 @@ class DigestConstraint(ValidationConstraint):
     """Check that the column is a valid SHA1 hex digest."""
     @staticmethod
     def get_condition(column):
-        return column.op("~")("^[0-9a-f]{40}$")
+        return column.op("~")("^([0-9a-f]{40}|%s)$"
+                              % FileCacher.TOMBSTONE_DIGEST)
 
 
 class IPv4Constraint(ValidationConstraint):
--- a/cms/db/validation.py
+++ b/cms/db/validation.py
@@ -1,0 +1,107 @@
+#!/usr/bin/env python2
+# -*- coding: utf-8 -*-
+
+# Contest Management System - http://cms-dev.github.io/
+# Copyright © 2017 Luca Wehrstedt <luca.wehrstedt@gmail.com>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import absolute_import
+from __future__ import print_function
+from __future__ import unicode_literals
+
+from sqlalchemy import or_, not_
+from sqlalchemy.event import listens_for
+from sqlalchemy.orm import mapper
+from sqlalchemy.schema import CheckConstraint
+
+
+class ValidationConstraint(object):
+    """Constraint adding server-side validation to a SQLAlchemy Column.
+
+    An instance of (a subclass of) this object has to be provided as an
+    argument to the constructor of the Column object to which it has to
+    apply. It results in a CheckConstraint being added to the table of
+    the class said column is defined on.
+
+    """
+    # In order to create a CheckConstraint we need to know the name of
+    # the column, but the Column object doesn't know this itself until
+    # it gets eventually assigned to a class-level attribute and the
+    # metaclass magic of SQLAlchemy kicks in. We could ask the user to
+    # provide the name as an argument, but that would lead to ugly
+    # redundancy. The solution we adopt stores the Column objects we
+    # want to add a constraint to, listens to an SQLAlchemy that gets
+    # fired when all the mappers are ready and only then attaches the
+    # CheckConstraint to the table. This is all mostly legal, with the
+    # exception of this class exposing a method that mimics an
+    # undocumented internal SQLAlchemy interface.
+    columns = list()
+
+    @staticmethod
+    def get_condition(column):
+        """Build an expression checking that the given Column is valid."""
+        raise NotImplementedError()
+
+    @staticmethod
+    @listens_for(mapper, "mapper_configured")
+    def _mapper_configured_handler(class_mapper, _):
+        """Add the constraints to the table of the given mapper."""
+        table = class_mapper.mapped_table
+        for obj, column in ValidationConstraint.columns:
+            if table.columns.contains_column(column):
+                table.append_constraint(
+                    CheckConstraint(obj.get_condition(column)))
+
+    # This is the method called by the Column object's constructor to
+    # let us bind to it. We sort of abuse this.
+    def _set_parent_with_dispatch(self, column):
+        self.columns.append((self, column))
+
+
+class CodenameConstraint(ValidationConstraint):
+    """Check that the column uses a limited alphabet."""
+    @staticmethod
+    def get_condition(column):
+        return column.op("~")("^[A-Za-z0-9_-]+$")
+
+
+class FilenameConstraint(ValidationConstraint):
+    """Check that the column is a valid POSIX path component."""
+    @staticmethod
+    def get_condition(column):
+        # These are the path components allowed by ext2/3/4 and by any
+        # other POSIX-compliant filesystem. Observe that NULL (0) chars
+        # are forbidden but they are already forbidden by PostgreSQL.
+        return not_(or_(
+            column == "", column == ".", column == "..", column.contains("/")))
+
+
+class DigestConstraint(ValidationConstraint):
+    """Check that the column is a valid SHA1 hex digest."""
+    @staticmethod
+    def get_condition(column):
+        return column.op("~")("^[0-9a-f]{40}$")
+
+
+class IPv4Constraint(ValidationConstraint):
+    """Check that the column is a valid IPv4 address (plus subnet?)."""
+    RE255 = "[0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5]"
+    RE32 = "[0-9]|[1-2][0-9]|3[0-2]"
+
+    @staticmethod
+    def get_condition(column):
+        return column.op("~")(
+            "^(%(re255)s)\.(%(re255)s)\.(%(re255)s)\.(%(re255)s)(/(%(re32)s))?$"
+            % {"re255": IPv4Constraint.RE255, "re32": IPv4Constraint.RE32})
--- a/cms/db/validation.py
+++ b/cms/db/validation.py
@@ -17,82 +17,54 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+"""Constraints adding server-side validation to a SQLAlchemy Column.
+
+This module provides classes that can be used to add constraints to
+textual Columns in order to ensure that the values in these columns
+always match a given regular expression (unless they are NULL). These
+conditions are enforced server-side, by the database itself, and are
+added to the table upon creation. Only PostgreSQL and psycopg2 are
+supported.
+
+"""
+
 from __future__ import absolute_import
 from __future__ import print_function
 from __future__ import unicode_literals
 
-from sqlalchemy import or_, not_
-from sqlalchemy.event import listens_for
-from sqlalchemy.orm import mapper
+from sqlalchemy import and_, literal_column
 from sqlalchemy.schema import CheckConstraint
+from sqlalchemy.sql.expression import ColumnClause
 
 from cms.db.filecacher import FileCacher
 
 
-class ValidationConstraint(object):
-    """Constraint adding server-side validation to a SQLAlchemy Column.
+# TODO In SQLAlchemy 1.2 we can get rid of literal_column. See:
+# https://bitbucket.org/zzzeek/sqlalchemy/issues/3957/incorrect-rendering-of-sqltext-for-column
+# https://groups.google.com/d/topic/sqlalchemy/J6VPG_1Fj2U/discussion
 
-    An instance of (a subclass of) this object has to be provided as an
-    argument to the constructor of the Column object to which it has to
-    apply. It results in a CheckConstraint being added to the table of
-    the class said column is defined on.
-
-    """
-    # In order to create a CheckConstraint we need to know the name of
-    # the column, but the Column object doesn't know this itself until
-    # it gets eventually assigned to a class-level attribute and the
-    # metaclass magic of SQLAlchemy kicks in. We could ask the user to
-    # provide the name as an argument, but that would lead to ugly
-    # redundancy. The solution we adopt stores the Column objects we
-    # want to add a constraint to, listens to an SQLAlchemy that gets
-    # fired when all the mappers are ready and only then attaches the
-    # CheckConstraint to the table. This is all mostly legal, with the
-    # exception of this class exposing a method that mimics an
-    # undocumented internal SQLAlchemy interface.
-    columns = list()
-
-    @staticmethod
-    def get_condition(column):
-        """Build an expression checking that the given Column is valid."""
-        raise NotImplementedError()
-
-    @staticmethod
-    @listens_for(mapper, "mapper_configured")
-    def _mapper_configured_handler(class_mapper, _):
-        """Add the constraints to the table of the given mapper."""
-        table = class_mapper.mapped_table
-        for obj, column in ValidationConstraint.columns:
-            if table.columns.contains_column(column):
-                table.append_constraint(
-                    CheckConstraint(obj.get_condition(column)))
-
-    # This is the method called by the Column object's constructor to
-    # let us bind to it. We sort of abuse this.
-    def _set_parent_with_dispatch(self, column):
-        self.columns.append((self, column))
-
-
-class CodenameConstraint(ValidationConstraint):
+class CodenameConstraint(CheckConstraint):
     """Check that the column uses a limited alphabet."""
-    @staticmethod
-    def get_condition(column):
-        return column.op("~")("^[A-Za-z0-9_-]+$")
+    def __init__(self, column_name):
+        column = ColumnClause(column_name)
+        super(CodenameConstraint, self).__init__(
+            column.op("~")(literal_column("'^[A-Za-z0-9_-]+$'")))
 
 
-class FilenameConstraint(ValidationConstraint):
-    """Check that the column is a valid POSIX path component."""
-    @staticmethod
-    def get_condition(column):
-        # These are the path components allowed by ext2/3/4 and by any
-        # other POSIX-compliant filesystem. Observe that NULL (0) chars
-        # are forbidden but they are already forbidden by PostgreSQL.
-        return not_(or_(
-            column == "", column == ".", column == "..", column.contains("/")))
+class FilenameConstraint(CheckConstraint):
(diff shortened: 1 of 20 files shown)


cms/db/validation.py
```diff
@@ -0,0 +1,107 @@
...
+
+class FilenameConstraint(ValidationConstraint):
+    """Check that the column is a valid POSIX path component."""
+    @staticmethod
+    def get_condition(column):
+        # These are the path components allowed by ext2/3/4 and by any
+        # other POSIX-compliant filesystem. Observe that NULL (0) chars
+        # are forbidden but they are already forbidden by PostgreSQL.
+        return not_(or_(
+            column == "", column == ".", column == "..", column.contains("/")))
```
akmohtashami: Currently we are using the following regex in admin for matching file names.
[a-zA-Z0-9_.-]+
This doesn't allow other symbols such as ! or ?. I think it would be better if we would use a regex in this constraint making sure all files will be downloadable.

cms/db/validation.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+class DigestConstraint(ValidationConstraint):
+    """Check that the column is a valid SHA1 hex digest."""
+    @staticmethod
+    def get_condition(column):
+        return column.op("~")("^[0-9a-f]{40}$")
```
wil93: With the "tombstone-ization" work by @veluca93 this should also support digests being just a "x" string, shouldn't it?

lerks:
I see your point and I don't have a strong opinion on this. My reasoning
was that codenames could end up in all sorts of places so we'd rather err
on the safe side whereas filenames only end in a few very similar-behaving
places (names of files on the filesystem and URL components) so we can be
more tolerant. In URLs any character can be escaped so that shouldn't be a
problem. Do you know of any endpoint where we're not escaping properly?

lerks:
I don't know what this tombstone-ization is; I'll look into it. But all
tests are passing...

wil93:
For tombstone-ization see: #658 

For the "x" digest, see https://github.com/cms-dev/cms/pull/658/files#diff-60b25826f3307de7400d19ff11fd9d00R807

lerks:
@wil93 Does the fixup address your concerns?

wil93:
Looks good 👍

stefano-maggiolo:
What is the plan for updating dumps that do not satisfy these constraints?

---

Review status: 0 of 11 files reviewed at latest revision, 5 unresolved discussions, some commit checks broke.

---

*[cms/db/validation.py, line 49 at r1](https://reviewable.io:443/reviews/cms-dev/cms/738#-KgmsjU3RL0zZ54fbG9j:-KgmsjU3RL0zZ54fbG9k:byozf5x) ([raw file](https://github.com/cms-dev/cms/blob/4f2f4b587a0dbb926052beedbbca73f6a65ac342/cms/db/validation.py#L49)):*
> ```Python
>     # CheckConstraint to the table. This is all mostly legal, with the
>     # exception of this class exposing a method that mimics an
>     # undocumented internal SQLAlchemy interface.
> ```

Is this another SMC? Why not suffer a minimal redundancy with the gain of using only public features, and a much simpler mechanism?

---

*[cms/db/validation.py, line 107 at r1](https://reviewable.io:443/reviews/cms-dev/cms/738#-Kgmtltv7FDx6YnDNihs:-Kgmtltv7FDx6YnDNiht:b-i6y7m3) ([raw file](https://github.com/cms-dev/cms/blob/4f2f4b587a0dbb926052beedbbca73f6a65ac342/cms/db/validation.py#L107)):*
> ```Python
>         return column.op("~")(
>             "^(%(re255)s)\.(%(re255)s)\.(%(re255)s)\.(%(re255)s)(/(%(re32)s))?$"
>             % {"re255": IPv4Constraint.RE255, "re32": IPv4Constraint.RE32})
> ```

Can we use https://docs.python.org/3/library/ipaddress.html and support ipv6 too?

---

*[cms/db/validation.py, line 32 at r3](https://reviewable.io:443/reviews/cms-dev/cms/738#-KgmudwgfpJAChn3okQF:-KgmudwgfpJAChn3okQG:b-1rzpgf) ([raw file](https://github.com/cms-dev/cms/blob/fe913c9d6f32401b6c109d92a386d1abfc2daec4/cms/db/validation.py#L32)):*
> ```Python
> 
> 
> class ValidationConstraint(object):
> ```

This file is a good candidate for a unittest (if better detached from alchemy's internals)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/738#-:-Kgmup4joICkPNXXGCpY:b-gq31nd)*

wil93:
I was thinking: how about changing `default=` to `server_default=` whenever possible in `cms/db/*.py`?

lerks:
@stefano-maggiolo: I thought that wasn't needed, but I was wrong: done.

@wil93: out of scope for this review.

---

Review status: 0 of 12 files reviewed at latest revision, 5 unresolved discussions.

---

*[cms/db/validation.py, line 49 at r1](https://reviewable.io:443/reviews/cms-dev/cms/738#-KgmsjU3RL0zZ54fbG9j:-KgnxgNNHx5WWMy8vjhb:b-7assv7) ([raw file](https://github.com/cms-dev/cms/blob/4f2f4b587a0dbb926052beedbbca73f6a65ac342/cms/db/validation.py#L49)):*

Is this another SMC? Why not suffer a minimal redundancy with the gain of using only public features, and a much simpler mechanism?
</blockquote></details>

I was afraid you would say that. In some sense, yes, this could break at any time if SQLAlchemy changed its internals (I think this is unlikely as it's a very simple interface). On the other hand, unlike SMC, we can easily revert to a "safe" implementation at any time by adding that redundancy.

One more reason I prefer this that I forgot to mention is that, in this way, we can express the condition using Column objects, operators and methods in Python. If we were to go the other way all the conditions would have to be encoded as opaque SQL strings.

If you still feel we should go the other way just tell me, I'll do it, I don't feel too strongly about this.

---

*[cms/db/validation.py, line 107 at r1](https://reviewable.io:443/reviews/cms-dev/cms/738#-Kgmtltv7FDx6YnDNihs:-KgnyyE5vtRMLFYE-wuD:bq5tsnr) ([raw file](https://github.com/cms-dev/cms/blob/4f2f4b587a0dbb926052beedbbca73f6a65ac342/cms/db/validation.py#L107)):*

Can we use https://docs.python.org/3/library/ipaddress.html and support ipv6 too?
</blockquote></details>

I'm not sure I understand what you mean. The validation checks we're adding here will occur on PostgreSQL's side meaning that, even though we define them using Python, they need to be expressed in pure SQL (possibly with Postgres' extensions). I couldn't find in `ipaddress` a regular expression for IPv6 addresses that we could use here. Did I miss it?

Moreover, `ipaddress` isn't available in py2.

---

*[cms/db/validation.py, line 88 at r2](https://reviewable.io:443/reviews/cms-dev/cms/738#-KgggpC3PK6YTS03GPIq-r2-88:-KgnyoygV-rhz2gBRVIT:b-eq7uvl) ([raw file](https://github.com/cms-dev/cms/blob/e1f0ca6dca6397216aa70accbd976310652a8505/cms/db/validation.py#L88)):*
<details><summary><i>Previously, akmohtashami (Amir Keivan Mohtashami) wrote…</i></summary><blockquote>

Currently we are using the following regex in admin for matching file names.
[a-zA-Z0-9_.-]+
This doesn't allow other symbols such as ! or ?. I think it would be better if we would use a regex in this constraint making sure all files will be downloadable.
</blockquote></details>

@stefano-maggiolo: your opinion here?

---

*[cms/db/validation.py, line 95 at r2](https://reviewable.io:443/reviews/cms-dev/cms/738#-KgggpC3PK6YTS03GPIq-r2-95:-KgnywYYBDlq3O-0YPzF:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/e1f0ca6dca6397216aa70accbd976310652a8505/cms/db/validation.py#L95)):*

With the "tombstone-ization" work by @veluca93 this should also support digests being just a "x" string, shouldn't it?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/738#-:-KgoAbMiKE5pcJ0uvFKj:b-mnbnq1)*

wil93:
@lerks why out of scope? If there is one PR that should make those changes (if we want those changes) it's this one 😕

lerks:
As the title suggests, this PR is about DB validation; it's not "everything DB-related on people's wishlists". If you want those changes do them and send a PR in.

---

Review status: 0 of 12 files reviewed at latest revision, 5 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/738#-:-KgoJlu2JtyZisTQtjnw:bhjv0h7)*

lerks:
Review status: 0 of 12 files reviewed at latest revision, 5 unresolved discussions.

---

*[cms/db/validation.py, line 32 at r3](https://reviewable.io:443/reviews/cms-dev/cms/738#-KgmudwgfpJAChn3okQF:-KgoKMoDrpD6fIeSa66N:b-i6mc9k) ([raw file](https://github.com/cms-dev/cms/blob/fe913c9d6f32401b6c109d92a386d1abfc2daec4/cms/db/validation.py#L32)):*

This file is a good candidate for a unittest (if better detached from alchemy's internals)
</blockquote></details>

Oh, just realized: if we go the other way (pass column name as argument instead of doing alchemy black magic) unit tests become extremely hard to perform as we would have to parse an opaque SQL string...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/738)*

wil93:
As the title suggests, this is about **adding in-database check constraints**. The `default` clause does not generate a in-database check contstraint. The `server_default` clause does. I fail to see how this does not qualify as "exactly what this PR is about"? 🤔

stefano-maggiolo:
PR numbers are cheap, we can create a new one for server_default.

---

Review status: 0 of 12 files reviewed at latest revision, 4 unresolved discussions.

---

*[cms/db/validation.py, line 107 at r1](https://reviewable.io:443/reviews/cms-dev/cms/738#-Kgmtltv7FDx6YnDNihs:-KgocUDbeOUDo4yg7--o:b-bs7xkg) ([raw file](https://github.com/cms-dev/cms/blob/4f2f4b587a0dbb926052beedbbca73f6a65ac342/cms/db/validation.py#L107)):*

I'm not sure I understand what you mean. The validation checks we're adding here will occur on PostgreSQL's side meaning that, even though we define them using Python, they need to be expressed in pure SQL (possibly with Postgres' extensions). I couldn't find in `ipaddress` a regular expression for IPv6 addresses that we could use here. Did I miss it?

Moreover, `ipaddress` isn't available in py2.
</blockquote></details>

oh right (for py2 we could require it, but that's not the main point).

In that case I'm a bit torned - do we want to prevent people in an ipv6 network to use these functionalities?

---

*[cms/db/validation.py, line 88 at r2](https://reviewable.io:443/reviews/cms-dev/cms/738#-KgggpC3PK6YTS03GPIq-r2-88:-KgocJIhTjltijYxHlh3:b49a4ai) ([raw file](https://github.com/cms-dev/cms/blob/e1f0ca6dca6397216aa70accbd976310652a8505/cms/db/validation.py#L88)):*

@stefano-maggiolo: your opinion here?
</blockquote></details>

I don't mind being more restrictive.

---

*[cms/db/validation.py, line 32 at r3](https://reviewable.io:443/reviews/cms-dev/cms/738#-KgmudwgfpJAChn3okQF:-Kgod8g-mkBZesNot0UB:b7ibrog) ([raw file](https://github.com/cms-dev/cms/blob/fe913c9d6f32401b6c109d92a386d1abfc2daec4/cms/db/validation.py#L32)):*

Oh, just realized: if we go the other way (pass column name as argument instead of doing alchemy black magic) unit tests become extremely hard to perform as we would have to parse an opaque SQL string...
</blockquote></details>

Given all are regexps (once you change the filename), you could have constants instead of methods and build the constraint in cms/db/x.py

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/738#-:-KgoaGcZ2vyEE7v7sJcD:b4ityi7)*

wil93:
> PR numbers are cheap, we can create a new one

Of course 😃

lerks:
For IP addresses I did the "right thing": I used the CIDR type inside Postgres, and had it mapped to the types in the ipaddress module you mentioned (which, I found out, has been backported to py2).

For the other fields I got rid of the black magic (adding the redundancy of specifying each column's name again). I didn't do exactly as we agreed: I kept using separate classes, and used alchemy conditions instead of textual SQL expressions. I didn't add tests as the regexps we have now are extremely simple.

The ball is back in your court.

---

Review status: 0 of 20 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/738#-:-Kgtia2UCDMOpads-W8k:b-cvpf60)*

lerks:
@wil93 Could you check whether the IPv6 change breaks anything in the loaders? I'm unfamiliar with how they work...

lerks:
The IPv6 commit relates to #157.

cms/server/contest/handlers/base.py
```diff
@@ -66,26 +67,23 @@
...
-        or client belongs to one of the subnets in the whitelist.
+    return (bool): whether the address belongs to one of the networks.
 
     """
-    for wanted in whitelist.split(","):
-        wanted, sep, subnet = wanted.partition('/')
+    try:
+        address = ipaddress.ip_address(address)
+    except ValueError:
+        return False
```
wil93: This should never happen because of DB validation, right? If that's the case, wouldn't it be better to throw an error?

lerks: No, the address is what Tornado reports the client's address to be (the one of the machine that is trying to log in). In theory that is supposed to be a valid IP, but I've learnt to rather be safe than sorry when it comes to Tornado.

stefano-maggiolo:
Thanks, I like it much more now.

---

Review status: 0 of 20 files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[cms/db/base.py, line 36 at r6](https://reviewable.io:443/reviews/cms-dev/cms/738#-KgtiT4ZjGEMfiOw8il0-r6-36:-Kgwi1vI3JYrIryD9a-E:bqlsr7o) ([raw file](https://github.com/cms-dev/cms/blob/7a08a079d10d504802cf4e6bb16d2ec712341e96/cms/db/base.py#L36)):*

Codacy Issue found: [Unused ARRAY imported from sqlalchemy.dialects.postgresql](https://www.codacy.com/app/s-maggiolo/cms/pullRequest?prid=588948)
</blockquote></details>

Please fix

---

*[cms/db/types.py, line 39 at r6](https://reviewable.io:443/reviews/cms-dev/cms/738#-Kgwi5vnXdFIQ7BXxcMP:-Kgwi5vnXdFIQ7BXxcMQ:b7jxg54) ([raw file](https://github.com/cms-dev/cms/blob/7a08a079d10d504802cf4e6bb16d2ec712341e96/cms/db/types.py#L39)):*
> ```Python
> 
> 
> # Taken from:
> ```

Can you add also two lines on what this does?

---

*[cms/server/admin/handlers/base.py, line 124 at r6](https://reviewable.io:443/reviews/cms-dev/cms/738#-KgwirchXSLu5YqfD9i4:-KgwirciUSdmA8TZVmOQ:b4urf3q) ([raw file](https://github.com/cms-dev/cms/blob/7a08a079d10d504802cf4e6bb16d2ec712341e96/cms/server/admin/handlers/base.py#L124)):*
> ```Python
> 
> 
> def parse_ip_networks(networks):
> ```

Can you add pydoc for the argument?

---

*[cms/server/contest/handlers/base.py, line 83 at r6](https://reviewable.io:443/reviews/cms-dev/cms/738#-KgtiT4ZjGEMfiOw8il5-r6-83:-KgwjBAE6O6GV_Zh17rV:b-f2dc9k) ([raw file](https://github.com/cms-dev/cms/blob/7a08a079d10d504802cf4e6bb16d2ec712341e96/cms/server/contest/handlers/base.py#L83)):*

No, the address is what Tornado reports the client's address to be (the one of the machine that is trying to log in). In theory that is supposed to be a valid IP, but I've learnt to rather be safe than sorry when it comes to Tornado.
</blockquote></details>

Agree with Luca

---

*[cmscontrib/updaters/update_26.py, line 138 at r6](https://reviewable.io:443/reviews/cms-dev/cms/738#-KgwjP5BODT0UVp6d4Tq:-KgwjP5BODT0UVp6d4Tr:b-ddo4cg) ([raw file](https://github.com/cms-dev/cms/blob/7a08a079d10d504802cf4e6bb16d2ec712341e96/cmscontrib/updaters/update_26.py#L138)):*
> ```Python
>                             logger.critical(
>                                 "The dump contains an instance of %s whose %s "
>                                 "field contains an invalid IPv4 address: %s",
> ```

`%s' for the last one (similarly above)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/738#-:-Kgwji8sQ1UpFQvpg28x:bjww3px)*

stefano-maggiolo:
For me it's ok after the fixes

---

Review status: 0 of 20 files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/738#-:-KgwjoNzGjB3tVU6jjam:b9scxft)*

lerks:
Can you have one last look?

---

Review status: 0 of 20 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[cms/db/base.py, line 36 at r6](https://reviewable.io:443/reviews/cms-dev/cms/738#-KgtiT4ZjGEMfiOw8il0-r6-36:-KgyK5JjKdE3T524adn5:b8bbleb) ([raw file](https://github.com/cms-dev/cms/blob/7a08a079d10d504802cf4e6bb16d2ec712341e96/cms/db/base.py#L36)):*

Please fix
</blockquote></details>

Apparently CIDR was unused as well but undetected. Removed both imports.

---

*[cms/db/types.py, line 39 at r6](https://reviewable.io:443/reviews/cms-dev/cms/738#-Kgwi5vnXdFIQ7BXxcMP:-KgyQZiMZbrh2LcRviRZ:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/7a08a079d10d504802cf4e6bb16d2ec712341e96/cms/db/types.py#L39)):*

Can you add also two lines on what this does?
</blockquote></details>

Done.

---

*[cms/db/validation.py, line 107 at r1](https://reviewable.io:443/reviews/cms-dev/cms/738#-Kgmtltv7FDx6YnDNihs:-KgyN-QbXzSy_szzNoZM:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/4f2f4b587a0dbb926052beedbbca73f6a65ac342/cms/db/validation.py#L107)):*

oh right (for py2 we could require it, but that's not the main point).

In that case I'm a bit torned - do we want to prevent people in an ipv6 network to use these functionalities?
</blockquote></details>

Done.

---

*[cms/db/validation.py, line 88 at r2](https://reviewable.io:443/reviews/cms-dev/cms/738#-KgggpC3PK6YTS03GPIq-r2-88:-KgyKptYIy1SZGYqKC7o:b-2gtqu0) ([raw file](https://github.com/cms-dev/cms/blob/e1f0ca6dca6397216aa70accbd976310652a8505/cms/db/validation.py#L88)):*

I don't mind being more restrictive.
</blockquote></details>

Done. Had to add % as well for SubmissionFormatElements.

---

*[cms/db/validation.py, line 32 at r3](https://reviewable.io:443/reviews/cms-dev/cms/738#-KgmudwgfpJAChn3okQF:-KgyKZ2GF-hMjGTpXXpj:bes8n0t) ([raw file](https://github.com/cms-dev/cms/blob/fe913c9d6f32401b6c109d92a386d1abfc2daec4/cms/db/validation.py#L32)):*

Given all are regexps (once you change the filename), you could have constants instead of methods and build the constraint in cms/db/x.py
</blockquote></details>

Is this a requirement? I kind of prefer adding this small abstraction layer that hides away how these constraints are enforced (regexp, inequalities, etc.). What's the reason you seem to dislike it?

---

*[cms/server/admin/handlers/base.py, line 124 at r6](https://reviewable.io:443/reviews/cms-dev/cms/738#-KgwirchXSLu5YqfD9i4:-KgyMghQZ_IWSQEuT3M_:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/7a08a079d10d504802cf4e6bb16d2ec712341e96/cms/server/admin/handlers/base.py#L124)):*

Can you add pydoc for the argument?
</blockquote></details>

Done.

---

*[cmscontrib/updaters/update_26.py, line 138 at r6](https://reviewable.io:443/reviews/cms-dev/cms/738#-KgwjP5BODT0UVp6d4Tq:-KgyNQXegFMSrtHFExKP:bkc45w8) ([raw file](https://github.com/cms-dev/cms/blob/7a08a079d10d504802cf4e6bb16d2ec712341e96/cmscontrib/updaters/update_26.py#L138)):*

`%s' for the last one (similarly above)
</blockquote></details>

Not sure I understood. Is what I did OK?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/738#-:-KgyQ_Pm3OEN2hM-O4Pf:b-u0n1f2)*

stefano-maggiolo:
:lgtm:

---

Review status: 0 of 20 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[cms/db/validation.py, line 32 at r3](https://reviewable.io:443/reviews/cms-dev/cms/738#-KgmudwgfpJAChn3okQF:-KgySb2fOoMPd0ctGOpK:blvw0eu) ([raw file](https://github.com/cms-dev/cms/blob/fe913c9d6f32401b6c109d92a386d1abfc2daec4/cms/db/validation.py#L32)):*

Is this a requirement? I kind of prefer adding this small abstraction layer that hides away how these constraints are enforced (regexp, inequalities, etc.). What's the reason you seem to dislike it?
</blockquote></details>

This was from before you changed the black magic. Now it's ok.

---

*[cmscontrib/updaters/update_26.py, line 138 at r6](https://reviewable.io:443/reviews/cms-dev/cms/738#-KgwjP5BODT0UVp6d4Tq:-KgySN6cxGWsetH4mUyC:b-j6u9bs) ([raw file](https://github.com/cms-dev/cms/blob/7a08a079d10d504802cf4e6bb16d2ec712341e96/cmscontrib/updaters/update_26.py#L138)):*

Not sure I understood. Is what I did OK?
</blockquote></details>

Perfect

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/738#-:-KgySSMuVzyTpMbwXFk6:bnfp4nl)*

lerks:
Pushed.

PR #739, opened 2017-04-02
Multi CWS taking over from #594

stefano-maggiolo:
I rebased onto master and added two commits to the original; the first to make it work and improve a bit (see commit message for details), the second to minimize the diff between master (once merged all commits together).

For this reason, comparing with `git diff -C` might be useful - contest.{py,html} are actually very close to base.{py,html}, and git is able to recognize this.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/739)


stefano-maggiolo:
@wil93 please review

lerks:
:lgtm:

I have essentially one big blocking comment that needs to be addressed.

Moreover, even though you asked me not to, I made some nit-picking comments. (Don't hate me because I love good code.) I marked them as non-blocking, meaning you can ignore them entirely (not even an ACK is needed).

But, if you give me access to your branch [1] I am willing to fix them myself.

[1] https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/

---

Reviewed 27 of 27 files at r1, 2 of 2 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, 2 of 2 files at r5, 6 of 7 files at r6, 2 of 2 files at r9.
Review status: all files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[cms/util.py, line 223 at r1](https://reviewable.io:443/reviews/cms-dev/cms/739#-Kgl2fswgwQYU2cpXFgc:-Kgl2fswgwQYU2cpXFgd:bmy82qd) ([raw file](https://github.com/cms-dev/cms/blob/3605fc992d9f7db69be275def96eaf6cf06c9995/cms/util.py#L223)):*
> ```Python
>     contest_id_help = "id of the contest to automatically load"
>     if ask_contest is None:
>         contest_id_help += " (ignored)"
> ```

This causes the `-c` argument to be listed as ignored in CWS's help, while it's just optional.

---

*[cms/util.py, line 235 at r1](https://reviewable.io:443/reviews/cms-dev/cms/739#-KgkhgwBDGjLHjXK0cv6:-KgkhgwBDGjLHjXK0cv7:b-gimddu) ([raw file](https://github.com/cms-dev/cms/blob/3605fc992d9f7db69be275def96eaf6cf06c9995/cms/util.py#L235)):*
> ```Python
>                           "quitting." % (cls.__name__))
> 
>     if args.contest_id is not None:
> ```

This breaks RS launching AWS, SS, Checker and PrintingService because RS will provide the `-c <id>` parameter which, now, won't be ignored but will actually forwarded to the constructor of these services, which don't have a `contest_id` parameter, thus raising an exception.

Observe moreover that using positional arguments is dangerous: Worker accepts a second parameter, but it isn't `contest_id`, it is `fake_worker_time`. I think using keyword arguments is safer here.

---

*[cms/server/util.py, line 186 at r1](https://reviewable.io:443/reviews/cms-dev/cms/739#-KgkjeWa1dPGwzprP9Gn:-KgkjeWa1dPGwzprP9Go:b-ibcz1b) ([raw file](https://github.com/cms-dev/cms/blob/3605fc992d9f7db69be275def96eaf6cf06c9995/cms/server/util.py#L186)):*
> ```Python
> def multi_contest(f):
>     """Return decorator allowing not to specify the "contest_name"
>     parameter, whenever a method requires it.
> ```

Docstring should start with a one-liner.

---

*[cms/server/util.py, line 196 at r1](https://reviewable.io:443/reviews/cms-dev/cms/739#-Kgkjaxp7Dsq9yWOEaL1:-KgkjaxqPA_Ap53enysg:b-qrqh29) ([raw file](https://github.com/cms-dev/cms/blob/3605fc992d9f7db69be275def96eaf6cf06c9995/cms/server/util.py#L196)):*
> ```Python
>         else:
>             # Send a fake "contest_name" URL parameter
>             f(args[0], None, *(args[1:]))
> ```

Parenthesis around `args[1:]` appear to be useless.

---

*[cms/server/util.py, line 197 at r1](https://reviewable.io:443/reviews/cms-dev/cms/739#-KgkpqFxjDwwQgjynDHr:-KgkpqFyUU86QS05JJ9v:b-jnj2ct) ([raw file](https://github.com/cms-dev/cms/blob/3605fc992d9f7db69be275def96eaf6cf06c9995/cms/server/util.py#L197)):*
> ```Python
>             # Send a fake "contest_name" URL parameter
>             f(args[0], None, *(args[1:]))
>     return wrapped_f
> ```

From looking at how this is used it appears that the `contest_name` parameter is always useless. I propose to modify this decorator to something like
```
def wrapped_f(self, *args):
    if self.application.service.contest is None:
        f(self, *args[1:])
    else:
        f(self, *args)
```

---

*[cms/server/contest/server.py, line 97 at r1](https://reviewable.io:443/reviews/cms-dev/cms/739#-KgkkPmj8YgdZvUzp2-D:-KgkkPmj8YgdZvUzp2-E:b-54uil4) ([raw file](https://github.com/cms-dev/cms/blob/3605fc992d9f7db69be275def96eaf6cf06c9995/cms/server/contest/server.py#L97)):*
> ```Python
>                 handlers.append((r'/([^/]+)' + h[0],) + h[1:])
>         else:
>             HANDLERS.append((r"/", MainHandler))
> ```

Please don't modify a global supposedly-constant variable. 

---

*[cms/server/contest/handlers/contest.py, line 126 at r1](https://reviewable.io:443/reviews/cms-dev/cms/739#-Kgko4IoahKCWGl4qMZz:-Kgko4IoahKCWGl4qM_-:b-cvg18a) ([raw file](https://github.com/cms-dev/cms/blob/3605fc992d9f7db69be275def96eaf6cf06c9995/cms/server/contest/handlers/contest.py#L126)):*
> ```Python
>         # Run render_params() now, not at the beginning of the request,
>         # because we need contest_name
>         self.r_params = self.render_params()
> ```

I'd move this into `prepare`, as this is the way it was before.

---

*[cms/server/contest/templates/contest.html, line 17 at r1](https://reviewable.io:443/reviews/cms-dev/cms/739#-Kgl1LL2O9UJOBhKEQi6:-Kgl1LL3ZUqS8EiQJ6dF:b81dd9d) ([raw file](https://github.com/cms-dev/cms/blob/3605fc992d9f7db69be275def96eaf6cf06c9995/cms/server/contest/templates/contest.html#L17)):*
> ```HTML
> {% from cms.grading.languagemanager import get_language %}
> {% from cmscommon.datetime import make_timestamp, utc %}
> {% from cmscommon.isocodes import translate_language_country_code %}
> ```

Can the imports go at the top of the file, outside of the `js` block?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/739#-:-KglM_kEXoA1W-X5JmBf:bgh3tv5)*

stefano-maggiolo:
Review status: 25 of 30 files reviewed at latest revision, 7 unresolved discussions.

---

*[cms/util.py, line 223 at r1](https://reviewable.io:443/reviews/cms-dev/cms/739#-Kgl2fswgwQYU2cpXFgc:-Kgm_ZHP6TAgfcu-S8ZE:b-886gsx) ([raw file](https://github.com/cms-dev/cms/blob/3605fc992d9f7db69be275def96eaf6cf06c9995/cms/util.py#L223)):*

This causes the `-c` argument to be listed as ignored in CWS's help, while it's just optional.
</blockquote></details>

Fixed

---

*[cms/util.py, line 235 at r1](https://reviewable.io:443/reviews/cms-dev/cms/739#-KgkhgwBDGjLHjXK0cv6:-Kgm__aikOiRMoj2iW5-:b-886gsx) ([raw file](https://github.com/cms-dev/cms/blob/3605fc992d9f7db69be275def96eaf6cf06c9995/cms/util.py#L235)):*

This breaks RS launching AWS, SS, Checker and PrintingService because RS will provide the `-c <id>` parameter which, now, won't be ignored but will actually forwarded to the constructor of these services, which don't have a `contest_id` parameter, thus raising an exception.

Observe moreover that using positional arguments is dangerous: Worker accepts a second parameter, but it isn't `contest_id`, it is `fake_worker_time`. I think using keyword arguments is safer here.
</blockquote></details>

Fixed

---

*[cms/server/util.py, line 196 at r1](https://reviewable.io:443/reviews/cms-dev/cms/739#-Kgkjaxp7Dsq9yWOEaL1:-Kgm_cnWQuUAjdfpt6Kt:b-nsxg6v) ([raw file](https://github.com/cms-dev/cms/blob/3605fc992d9f7db69be275def96eaf6cf06c9995/cms/server/util.py#L196)):*

Parenthesis around `args[1:]` appear to be useless.
</blockquote></details>

I like them

---

*[cms/server/util.py, line 197 at r1](https://reviewable.io:443/reviews/cms-dev/cms/739#-KgkpqFxjDwwQgjynDHr:-Kgm_tOO5k9v1tJadoYv:b-t8fveu) ([raw file](https://github.com/cms-dev/cms/blob/3605fc992d9f7db69be275def96eaf6cf06c9995/cms/server/util.py#L197)):*

From looking at how this is used it appears that the `contest_name` parameter is always useless. I propose to modify this decorator to something like
```
def wrapped_f(self, *args):
    if self.application.service.contest is None:
        f(self, *args[1:])
    else:
        f(self, *args)
```
</blockquote></details>

Do you mean remove it from all get()?

---

*[cms/server/contest/server.py, line 97 at r1](https://reviewable.io:443/reviews/cms-dev/cms/739#-KgkkPmj8YgdZvUzp2-D:-Kgma0elsdbI1Z0n8ZAJ:b2pcb1v) ([raw file](https://github.com/cms-dev/cms/blob/3605fc992d9f7db69be275def96eaf6cf06c9995/cms/server/contest/server.py#L97)):*

Please don't modify a global supposedly-constant variable. 
</blockquote></details>

I don't see a big problem, but if you want to improve after feel free ;)

---

*[cms/server/contest/handlers/contest.py, line 126 at r1](https://reviewable.io:443/reviews/cms-dev/cms/739#-Kgko4IoahKCWGl4qMZz:-KgmacpiAHmDV8uZOUYY:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/3605fc992d9f7db69be275def96eaf6cf06c9995/cms/server/contest/handlers/contest.py#L126)):*

I'd move this into `prepare`, as this is the way it was before.
</blockquote></details>

Done.

---

*[cms/server/contest/templates/contest.html, line 17 at r1](https://reviewable.io:443/reviews/cms-dev/cms/739#-Kgl1LL2O9UJOBhKEQi6:-KgmaI24Kxqnvn0BAbP-:blyvmxc) ([raw file](https://github.com/cms-dev/cms/blob/3605fc992d9f7db69be275def96eaf6cf06c9995/cms/server/contest/templates/contest.html#L17)):*

Can the imports go at the top of the file, outside of the `js` block?
</blockquote></details>

No, I tried but it doesn't work.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/739)*

wil93:
I didn't look at the code yet, but I think this looks good. The only thing I didn't like when testing it was the new `-a, --all-contests` flag. It's unnecessarily different from what we do with other services (ES, SS) that work with multiple contests. I think just having an optional `-c` was better.

stefano-maggiolo:
Yeah, I decided to change it mostly because I was scared by recent communications of people messing up because of easily mistakable or unexpected flag combinations. I agree it's not pretty we don't do the same for ES and SS (SS doesn't even have an option to restrict to a given contest).

Maybe my preference would be
- `cmsX` fails
- `cmsX -c` ask for which contest
- `cmsX -c 1` loads the contest 1
- `cmsX -a` (or `cmsX -c ALL` to avoid having the same short option as autorestart) loads all contests

@lerks wdyt?

lerks:
I don't like your specific proposal because both `-c` and `-a` start with a dash and should hence be optional, but if they're not given the service won't start.

I actually quite like the current way:
* `cmsX` loads all contests
* `cmsX -c` asks which contest to load
* `cmsX -c <id>` loads contest `<id>`

This doesn't work for ProxyService, which is currently the only service that requires a contest id. Yet, I think PS can be easily made contest-agnostic as well. (I'm not asking you to do it here; you can ignore PS for now and I will take care of making it contest-agnostic soon.)

---

Reviewed 4 of 4 files at r10, 1 of 1 files at r11.

---

*[cms/server/util.py, line 197 at r1](https://reviewable.io:443/reviews/cms-dev/cms/739#-KgkpqFxjDwwQgjynDHr:-Kgnvd30I3pRc8ZN8U7i:b-jik7dj) ([raw file](https://github.com/cms-dev/cms/blob/3605fc992d9f7db69be275def96eaf6cf06c9995/cms/server/util.py#L197)):*

Do you mean remove it from all get()?
</blockquote></details>

Yes. Basically, now, every handler that is decorated with `@multi_contest` has `contest_name` as its first argument (which may or may not be None), but none of these handler ever uses that argument. When they need the contest name they use `self.contest.name`, as that is set correctly in all cases. Therefore why not remove `contest_name` from the argument list altogether, and modify `multi_contest` accordingly (i.e., instead of adding a fake `contest_name` when it is missing, remove it when it is present)?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/739#-:-KgoOPWJ5FbrDXH34CM_:b-ac9sov)*

stefano-maggiolo:
As I was saying I don't like much the implicit loading of all contests (with potentially bad outcomes) by omitting an option, especially counting that it wasn't the default before. I definitely see this as something that bites people and could cause from mild annoyance to data leaks.

---

---

*[cms/server/util.py, line 197 at r1](https://reviewable.io:443/reviews/cms-dev/cms/739#-KgkpqFxjDwwQgjynDHr:-Kgo_k9i8Gqg61X-h6UR:b-tuv2mf) ([raw file](https://github.com/cms-dev/cms/blob/3605fc992d9f7db69be275def96eaf6cf06c9995/cms/server/util.py#L197)):*

Yes. Basically, now, every handler that is decorated with `@multi_contest` has `contest_name` as its first argument (which may or may not be None), but none of these handler ever uses that argument. When they need the contest name they use `self.contest.name`, as that is set correctly in all cases. Therefore why not remove `contest_name` from the argument list altogether, and modify `multi_contest` accordingly (i.e., instead of adding a fake `contest_name` when it is missing, remove it when it is present)?
</blockquote></details>

K, I'll do that (later)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/739#-:-Kgo_mzGn5YxsLsYJL38:b-by4nvj)*

stefano-maggiolo:
From private discussion, we agree on
- `cmsX` prompts for contest
- `cmsX -c ` fails
- `cmsX -c ALL` loads all contests
- `cmsX -c N` loads contest N.

I'll try to make this PR conformant to this (even if maybe not complete).

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/739#-:-KgodgZBP80pWVnTSuvq:b-c9ncyh)*

stefano-maggiolo:
I meant, "-c ALL instead of no args for all services accepting all contests".

---

Review status: 24 of 30 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/739#-:-KgpZfUc7kivITfoRxtI:b-klqio5)*

lerks:
Fix the two comments I added and you're good to push.

---

Reviewed 5 of 5 files at r12, 1 of 1 files at r13.

---

*[cms/server/util.py, line 197 at r1](https://reviewable.io:443/reviews/cms-dev/cms/739#-KgkpqFxjDwwQgjynDHr:-KgqpKQTn7i9t-7jMp2Y:b-9eqt1h) ([raw file](https://github.com/cms-dev/cms/blob/3605fc992d9f7db69be275def96eaf6cf06c9995/cms/server/util.py#L197)):*

Done.
</blockquote></details>

Fix the comments.

---

*[cms/server/contest/handlers/task.py, line 59 at r13](https://reviewable.io:443/reviews/cms-dev/cms/739#-Kgqq2hMGaMDSTPS7Zoy:-Kgqq2hMGaMDSTPS7Zoz:b-3roxvz) ([raw file](https://github.com/cms-dev/cms/blob/4a0a9d2d027f05c408cb43fdf1d81f0375ab0f30/cms/server/contest/handlers/task.py#L59)):*
> ```Python
>     @actual_phase_required(0)
>     @multi_contest
>     def get(self, contest_name, task_name):
> ```

You forgot to remove `contest_name` from the handlers in this file.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/739#-:-KgqqmdonqzEZtcFup2Q:b-g78ta2)*

stefano-maggiolo:

---

*[cms/server/util.py, line 197 at r1](https://reviewable.io:443/reviews/cms-dev/cms/739#-KgkpqFxjDwwQgjynDHr:-KgrexMXzDyeWvK1xEyW:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/3605fc992d9f7db69be275def96eaf6cf06c9995/cms/server/util.py#L197)):*

Fix the comments.
</blockquote></details>

Done.

---

*[cms/server/contest/handlers/task.py, line 59 at r13](https://reviewable.io:443/reviews/cms-dev/cms/739#-Kgqq2hMGaMDSTPS7Zoy:-KgreymsIPuEhhFzXi71:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/4a0a9d2d027f05c408cb43fdf1d81f0375ab0f30/cms/server/contest/handlers/task.py#L59)):*

You forgot to remove `contest_name` from the handlers in this file.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/739)*

stefano-maggiolo:
Merged

PR #742, opened 2017-04-06
Fix a few issues with the new CIDR and ipaddress usage

lerks:
We were using a too recent API of psycopg2, CWS was broken and some
code in cmscontrib had not been updated.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/742)

### DIFF
--- a/cms/db/types.py
+++ b/cms/db/types.py
@@ -26,6 +26,8 @@ from __future__ import absolute_import
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import ipaddress
+
 import psycopg2.extras
 import sqlalchemy
 from sqlalchemy.dialects.postgresql import ARRAY
@@ -33,7 +35,42 @@ from sqlalchemy.types import TypeDecorator, Unicode
 
 
 # Have psycopg2 use the types in ipaddress to represent the CIDR type.
-psycopg2.extras.register_ipaddress()
+# A handy function to do that has been added in v2.7, for versions
+# before that we do it ourselves. See:
+# https://github.com/psycopg/psycopg2/commit/643ba70bad0f19a68c06ec95de2691c28e060e48
+
+if tuple(int(x) for x in psycopg2.__version__.split('.')[:2]) >= (2, 7):
+    psycopg2.extras.register_ipaddress()
+
+else:
+    from psycopg2.extensions import new_type, new_array_type, register_type, \
+        register_adapter, QuotedString
+
+    def cast_interface(s, cur=None):
+        if s is None:
+            return None
+        return ipaddress.ip_interface(unicode(s))
+
+    def cast_ipnetwork(s, cur=None):
+        if s is None:
+            return None
+        return ipaddress.ip_network(unicode(s))
+
+    inet = new_type((869,), b'INET', cast_interface)
+    ainet = new_array_type((1041,), b'INET[]', inet)
+
+    cidr = new_type((650,), b'CIDR', cast_ipnetwork)
+    acidr = new_array_type((651,), b'CIDR[]', cidr)
+
+    for c in [inet, ainet, cidr, acidr]:
+        register_type(c, None)
+
+    def adapt_ipaddress(obj):
+        return QuotedString(str(obj))
+
+    for t in [ipaddress.IPv4Interface, ipaddress.IPv6Interface,
+              ipaddress.IPv4Network, ipaddress.IPv6Network]:
+        register_adapter(t, adapt_ipaddress)
 
 
 # Taken from:
--- a/cms/server/contest/handlers/contest.py
+++ b/cms/server/contest/handlers/contest.py
@@ -34,6 +34,7 @@ from __future__ import absolute_import
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import ipaddress
 import logging
 import pickle
 import socket
@@ -69,7 +70,7 @@ NOTIFICATION_SUCCESS = "success"
 def check_ip(address, networks):
     """Return if client IP belongs to one of the accepted networks.
 
-    address (string): IP address to verify.
+    address (bytes): IP address to verify.
     networks ([ipaddress.IPv4Network|ipaddress.IPv6Network]): IP
         networks (addresses w/ subnets) to check against.
 
@@ -77,7 +78,7 @@ def check_ip(address, networks):
 
     """
     try:
-        address = ipaddress.ip_address(address)
+        address = ipaddress.ip_address(unicode(address))
     except ValueError:
         return False
 
@@ -196,10 +197,16 @@ class ContestHandler(BaseHandler):
             matching the remote IP address.
 
         """
-        remote_ip = self.request.remote_ip
+        try:
+            # We encode it as a network (i.e., we assign it a /32 or
+            # /128 mask) since we're comparing it for equality with
+            # other networks.
+            remote_ip = ipaddress.ip_network(unicode(self.request.remote_ip))
+        except ValueError:
+            return None
         participations = self.sql_session.query(Participation)\
             .filter(Participation.contest == self.contest)\
-            .filter(Participation.ip == remote_ip)
+            .filter(Participation.ip.any(remote_ip))
 
         # If hidden users are blocked we ignore them completely.
         if self.contest.block_hidden_participations:
--- a/cmscontrib/AddParticipation.py
+++ b/cmscontrib/AddParticipation.py
@@ -33,6 +33,7 @@ gevent.monkey.patch_all()
 
 import argparse
 import datetime
+import ipaddress
 import logging
 import sys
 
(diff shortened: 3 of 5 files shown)


stefano-maggiolo:
Did you test this manually?

---

Review status: 0 of 5 files reviewed at latest revision, 5 unresolved discussions.

---

*[cmscontrib/ImportContest.py, line 206 at r1](https://reviewable.io:443/reviews/cms-dev/cms/742#-Kh2qUPJ4Ugbbar192Cx:-Kh2qUPJ4Ugbbar192Cy:be0zsvr) ([raw file](https://github.com/cms-dev/cms/blob/6b6832d986cebab3dc3b8d321d6f43a900954c5f/cmscontrib/ImportContest.py#L206)):*
> ```Python
>                         args["hidden"] = p["hidden"]
>                     if "ip" in p:
>                         args["ip"] = [ipaddress.ip_address(p["ip"])]
> ```

Why ip_address here and ip_network earlier?

---

*[docs/Configuring a contest.rst, line 124 at r1](https://reviewable.io:443/reviews/cms-dev/cms/742#-Kh2o9FDMMbp2Qzzp7Pt:-Kh2o9FDMMbp2Qzzp7Pu:b-3y4wv4) ([raw file](https://github.com/cms-dev/cms/blob/6b6832d986cebab3dc3b8d321d6f43a900954c5f/docs/Configuring a contest.rst#L124)):*
> ```reStructuredText
> ==========
> 
> Users can log into CWS manually, using their credentials (username and a password), or they can get logged in automatically by CMS based on the IP addresses their requests are coming from.
> ```

I think it sounds better with singular (`the IP address their requests are coming from`)

---

*[docs/Configuring a contest.rst, line 129 at r1](https://reviewable.io:443/reviews/cms-dev/cms/742#-Kh2ooHnLJbWwKkZ01Hn:-Kh2ooHnLJbWwKkZ01Ho:b4jv7r2) ([raw file](https://github.com/cms-dev/cms/blob/6b6832d986cebab3dc3b8d321d6f43a900954c5f/docs/Configuring a contest.rst#L129)):*
> ```reStructuredText
> ----------------------------------
> 
> If the "IP-based autologin" option in the contest configuration is set, CWS tries to find a user that matches the IP address the request is coming from. If it finds exactly one user, the requester is automatically logged in as that user. If zero or more than one user match, CWS does not let the user in (and the incident is logged to allow troubleshooting).
> ```

There are other occurrences of "IP based" without dash later, maybe try to make them uniform.

---

*[docs/Configuring a contest.rst, line 131 at r1](https://reviewable.io:443/reviews/cms-dev/cms/742#-Kh2obfJGRxoawpmHT8g:-Kh2obfJGRxoawpmHT8h:b-oyylie) ([raw file](https://github.com/cms-dev/cms/blob/6b6832d986cebab3dc3b8d321d6f43a900954c5f/docs/Configuring a contest.rst#L131)):*
> ```reStructuredText
> If the "IP-based autologin" option in the contest configuration is set, CWS tries to find a user that matches the IP address the request is coming from. If it finds exactly one user, the requester is automatically logged in as that user. If zero or more than one user match, CWS does not let the user in (and the incident is logged to allow troubleshooting).
> 
> In general, each user can have multiple ranges of IP addresses associated to it. These are defined as a list of subnets in CIDR format (e.g., `192.168.1.0/24`). Only the subnets whose mask is maximal (i.e., `/32` for IPv4 or `/128` for IPv6) are considered for autologin purposes (subnets with non-maximal mask are still useful for IP-based restrictions, see below). The autologin will kick in if *any* of the subnets matches the IP of the request.
> ```

Do we require that the /32 or /128 to be spelled out?

---

*[docs/Configuring a contest.rst, line 151 at r1](https://reviewable.io:443/reviews/cms-dev/cms/742#-Kh2qCcVUtqyYP7ETskU:-Kh2qCcVUtqyYP7ETskV:b-mwathf) ([raw file](https://github.com/cms-dev/cms/blob/6b6832d986cebab3dc3b8d321d6f43a900954c5f/docs/Configuring a contest.rst#L151)):*
> ```reStructuredText
> The following are some common reasons for login failures, all of them coming with some useful log message from CWS.
> 
> - IP address mismatch (with IP based autologin): if IP address doesn't match any subnet of any participation or if it matches some subnets of more than one participation, then the login fails. Note that if the user is using the IP address of a different user, CWS will happily log them in without noticing anything.
> ```

*the* IP address

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/742#-:-Kh2qbe5_qp4sWWo-gUv:bobjwkx)*

lerks:
I tested CWS and AWS manually. I didn't test the importers as I don't know how they work nowadays...

---

Review status: 0 of 5 files reviewed at latest revision, 5 unresolved discussions.

---

*[cmscontrib/ImportContest.py, line 206 at r1](https://reviewable.io:443/reviews/cms-dev/cms/742#-Kh2qUPJ4Ugbbar192Cx:-Kh33C_WuNqi047J53fN:b-e9bl14) ([raw file](https://github.com/cms-dev/cms/blob/6b6832d986cebab3dc3b8d321d6f43a900954c5f/cmscontrib/ImportContest.py#L206)):*

Why ip_address here and ip_network earlier?
</blockquote></details>

It's supposed to be `network` here as well, nice catch! Changed.

(It didn't actually have any effect, since psycopg2 converts both to the same string and then postgres converts them back to the type it expects)

---

*[docs/Configuring a contest.rst, line 124 at r1](https://reviewable.io:443/reviews/cms-dev/cms/742#-Kh2o9FDMMbp2Qzzp7Pt:-Kh348H99PwOX7Sn0wyA:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/6b6832d986cebab3dc3b8d321d6f43a900954c5f/docs/Configuring a contest.rst#L124)):*

I think it sounds better with singular (`the IP address their requests are coming from`)
</blockquote></details>

Done.

---

*[docs/Configuring a contest.rst, line 129 at r1](https://reviewable.io:443/reviews/cms-dev/cms/742#-Kh2ooHnLJbWwKkZ01Hn:-Kh348YBFYzgzDRIshW8:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/6b6832d986cebab3dc3b8d321d6f43a900954c5f/docs/Configuring a contest.rst#L129)):*

There are other occurrences of "IP based" without dash later, maybe try to make them uniform.
</blockquote></details>

Done.

---

*[docs/Configuring a contest.rst, line 131 at r1](https://reviewable.io:443/reviews/cms-dev/cms/742#-Kh2obfJGRxoawpmHT8g:-Kh33sVDB83nGeq2Mb_S:brhsusy) ([raw file](https://github.com/cms-dev/cms/blob/6b6832d986cebab3dc3b8d321d6f43a900954c5f/docs/Configuring a contest.rst#L131)):*

Do we require that the /32 or /128 to be spelled out?
</blockquote></details>

If the value is input through AWS one can omit the `/32` or `/128` but then it will be added automatically (i.e., when you hit the "Update" button the value that shows up in the `ip` field has the `/32` or `/128`).

---

*[docs/Configuring a contest.rst, line 151 at r1](https://reviewable.io:443/reviews/cms-dev/cms/742#-Kh2qCcVUtqyYP7ETskU:-Kh34fWbKNhItkMgnfTV:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/6b6832d986cebab3dc3b8d321d6f43a900954c5f/docs/Configuring a contest.rst#L151)):*

*the* IP address
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/742#-:-Kh32X2bSm8DkQTUOrCP:bg0qbwx)*

stefano-maggiolo:
You can run `cmsImportContest /path/to/con_test -i` to test (well, not now, since that is broken too...)
Answer the comment than :lgtm:

---

Reviewed 3 of 5 files at r1, 2 of 2 files at r2.

---

*[cmscontrib/ImportContest.py, line 206 at r2](https://reviewable.io:443/reviews/cms-dev/cms/742#-Kh5mXgA4kkhWzxJ2vTE:-Kh5mXgByP353AMfhcba:b-xphn6w) ([raw file](https://github.com/cms-dev/cms/blob/0e6cee6595a428eea141ba8d20164e9e05e91dc8/cmscontrib/ImportContest.py#L206)):*
> ```Python
>                         args["hidden"] = p["hidden"]
>                     if "ip" in p:
>                         args["ip"] = [ipaddress.ip_network(p["ip"])]
> ```

You must allow None as an input value (at least in con_test it is set to null)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/742#-:-Kh5cPqa1iUXPrU4w4on:btavpfr)*

lerks:
Pushed as 825afb83d56660b7c3bffa740aad386f006766ab.

PR #752, opened 2017-04-12
Fix cmsMake errors when running 'cmsMake -c' from a path with spaces in it

wil93:
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/752)

### DIFF
--- a/cmstaskenv/cmsMake.py
+++ b/cmstaskenv/cmsMake.py
@@ -579,9 +579,9 @@ def clean(base_dir, generated_list):
         pass
 
     # Delete compiled and/or backup files
-    os.system("find %s -name '*.o' -delete" % (base_dir))
-    os.system("find %s -name '*.pyc' -delete" % (base_dir))
-    os.system("find %s -name '*~' -delete" % (base_dir))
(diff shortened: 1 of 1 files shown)


stefano-maggiolo:
How about using `quote()`? (For example, your version still doesn't handle filenames with doublequotes in them)
```python
try:
    from shlex import quote as quote
except ImportError:
    from pipes import quote as quote
```
[...]

lerks:
Are those `as quote`s necessary?

karliss:
There is also `subprocess.call` which accepts list of arguments and doesn't require manual quoting.

stefano-maggiolo:
@lerks no I just copy pasted from stack overflow 😀

Agree with @karliss subprocess is better.

lerks:
Another (maybe even nicer way) to do this is to handle it entirely in Python:
```python
for dirname, _, filenames in os.walk(base_dir):
    for filename in filenames:
        if any(filename.endswith(ext) for ext in {".o", ".pyc", "~"}):
            os.remove(os.path.join(dirname, filename)
```

PR #753, opened 2017-04-13
Fixed communication notification

AmarJayR:
The communication unread notification would disappear on reload even when the communication tab was not clicked. This was because an incorrect cookie was being fetched ("unread_count" instead of "{contest.name)_unread_count").

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/753)


lerks:
The first commit seems to be valid: I can confirm the issue and I can confirm that is the fix; I pushed it as 6345064e5404389f54431e5a1b9eb955a0909f63. (Nice catch, by the way! ;) )

I don't understand what happened to the rest of the PR though. The later commits seem to undo some recent changes (e.g. analysis mode), maybe they're the result of an unsuccessful rebase? If there was anything else other than the first commit that you wanted pushed please clean up and reopen.

PR #759, opened 2017-05-10
Add the shell programming language

veluca93:
See title. This fixes running cmsMake with bash scripts as generators.

### DIFF
--- a/cms/grading/languages/sh.py
+++ b/cms/grading/languages/sh.py
@@ -1,0 +1,58 @@
+#!/usr/bin/env python2
+# -*- coding: utf-8 -*-
+
+# Contest Management System - http://cms-dev.github.io/
+# Copyright © 2017 Luca Versari <veluca93@gmail.com>
+#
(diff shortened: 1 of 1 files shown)


lerks:
Could you update the documentation and possibly add some tests? See #731 for an example.

stefano-maggiolo:
Why SH and not Sh or Shell?

CristianCantoro:
I would advise against using `/bin/sh` as the shell to execute as it may be a different shell on different systems. I think the decision on what to do with `/bin/sh` is up to the single OSes or Linux distros.

On Debian and Ubuntu `/bin/sh` is a symbolic link to `dash`, but [previously](https://wiki.ubuntu.com/DashAsBinSh) it was `bash`. On Mac OS X it's `bash` (precisely `bash v. 3.X`, which is the default that ships with OS X).

It would be better using a more specific shell and also indicate the version,  as for example Bash 4.0 introduced support for [associative arrays](http://www.linuxjournal.com/content/bash-associative-arrays).

PR #760, opened 2017-05-14
Fix cleanup errors when 'cmsMake -c' is run from a path containing spaces

lerks:
Replaces #752.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/760)


stefano-maggiolo:
Ok, only possible difference with previous behaviour is that IIUC before a directory named something~ was deleted, now it doesn't. Don't think it's a problem.

PR #761, opened 2017-05-15
Use robust mechanism to produce URLs in CWS and AWS

lerks:
This would have mitigated (and possibly prevented) #582.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/761)

### DIFF
--- a/cms/locale/bg/LC_MESSAGES/cms.po
+++ b/cms/locale/bg/LC_MESSAGES/cms.po
@@ -942,7 +942,7 @@ msgstr "Команди за компилиране"
 
 #, python-format
 msgid ""
-"You can find the rules for the %(type_pl)s on the <a href=\"%(url_root)s/"
+"You can find the rules for the %(type_pl)s on the <a href=\"%(contest_root)s/"
 "\">contest overview page</a>."
 msgstr ""
 
--- a/cms/locale/bs/LC_MESSAGES/cms.po
+++ b/cms/locale/bs/LC_MESSAGES/cms.po
@@ -734,8 +734,8 @@ msgid "Compilation commands"
 msgstr "Naredbe kompajlera"
 
 #, python-format
-msgid "You can find the rules for the %(type_pl)s on the <a href=\"%(url_root)s/\">contest overview page</a>."
-msgstr "Možete pronaći pravila za %(type_pl)s na <a href=\"%(url_root)s/\">stranici pregleda takmičenja</a>."
+msgid "You can find the rules for the %(type_pl)s on the <a href=\"%(contest_root)s/\">contest overview page</a>."
+msgstr "Možete pronaći pravila za %(type_pl)s na <a href=\"%(contest_root)s/\">stranici pregleda takmičenja</a>."
 
 msgid "Remeber that to see the detailed result of a submission you need to use both a contest-token and a task-token."
 msgstr "Zapamtite da ne možete vidjeti detaljan rezultat prijave osim ako iskoristite i contest-token i task-token."
--- a/cms/locale/es/LC_MESSAGES/cms.po
+++ b/cms/locale/es/LC_MESSAGES/cms.po
@@ -1016,10 +1016,10 @@ msgstr "Comandos de compilación"
 
 #, python-format
 msgid ""
-"You can find the rules for the %(type_pl)s on the <a href=\"%(url_root)s/"
+"You can find the rules for the %(type_pl)s on the <a href=\"%(contest_root)s/"
 "\">contest overview page</a>."
 msgstr ""
-"Puede consultar las reglas de los %(type_pl)s <a href=\"%(url_root)s/"
+"Puede consultar las reglas de los %(type_pl)s <a href=\"%(contest_root)s/"
 "\">aquí</a>."
 
 msgid ""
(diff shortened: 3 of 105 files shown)


lerks:
PS: I am aware that many lines are now too long, that's the result of (almost) automatic find-and-replace; I'm waiting to get approval of the function name before fixing them all.

stefano-maggiolo:
Btw, there are also some unused imports. Ok for the patch, as long as you actually tested a lot of links for each of the 4 cases (the cross product of "with or without multicontest" and "url root empty or non-empty".

In general, for these methods so commonly used I'd prefer a shorter name. How about url() and absurl()?

---

Review status: 0 of 101 files reviewed at latest revision, 26 unresolved discussions, some commit checks failed.

---

*[cms/server/util.py, line 736 at r3](https://reviewable.io:443/reviews/cms-dev/cms/761#-KkB65ziYFmv4DlpN_QY:-KkB65ziYFmv4DlpN_QZ:bqzijku) ([raw file](https://github.com/cms-dev/cms/blob/aee83e76ce8cc4b85074f55bfeac8ba16e6f5621/cms/server/util.py#L736)):*
> ```Python
>             url += quote("%s" % component, safe="")
>         if kwargs:
>             url += "?" + urlencode(kwargs)
> ```

This should go in the previous commit

---

*[cms/server/contest/handlers/taskusertest.py, line 156 at r2](https://reviewable.io:443/reviews/cms-dev/cms/761#-KkBA57LWvNU_W2qhUdH:-KkBA57MA8PmXGKVn2oW:brya9tm) ([raw file](https://github.com/cms-dev/cms/blob/4c2ec806755522aa70983bbd4436988e7f114d06/cms/server/contest/handlers/taskusertest.py#L156)):*
> ```Python
> 
>         self.fallback_page = \
>             self.make_unprefixed_contest_href("testing", task_name=task.name)
> ```

Is this the same behaviour? Wouldn't you create testing?task_name=<task.name> instead of testing?<task.name>

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/761#-:-KkBAQSs-rMcmCGblJ-5:b6hxzan)*

lerks:
Renamed by mapping `make` -> nothing, `href` -> `url`,  `absolute` -> contest, `unprefixed` -> abs (this is different from what discussed in private).

Some more details on how I got rid of our custom overridden `redirect`. The issue with just removing it, relying on the superclass method and having every caller provide a relative path was that tornado handles the redirect of unauthenticated users to the login page automatically, by calling the `redirect` method of the handler with the `login_url` parameter provided to the constructor of the server, which is an absolute path. By reverting to tornado's `redirect` there was no point in which that absolute path was made relative by prepending the `url_root`.

There were two solutions to that: use the `get_login_url` method on the handlers instead of the `login_url` parameter (CWS already does that), or have nginx reverse-translate these paths. It can only do so with the URLs that are in the HTTP `Location` header (it won't parse the payload of the HTTP response to look for paths there). This is what I ended up doing, and that is why I added the two `proxy_redirect` directives to the sample nginx config. In the end, all URLs that are used for redirects are in absolute form (no `url_root` is prepended) whereas all the ones that are in the body are relative.

I tested further and ended up fixing a few more things.

I haven't wrapped the overlong lines yet; I'll have codacy tell me which ones they are.

---

Review status: 0 of 104 files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[cms/server/util.py, line 736 at r3](https://reviewable.io:443/reviews/cms-dev/cms/761#-KkB65ziYFmv4DlpN_QY:-KkIGvx1UnWkMSJ_UoAY:b-8ued8s) ([raw file](https://github.com/cms-dev/cms/blob/aee83e76ce8cc4b85074f55bfeac8ba16e6f5621/cms/server/util.py#L736)):*

This should go in the previous commit
</blockquote></details>

I was thinking of squashing the CWS and AWS commit into one, since with the redirect changes AWS would be broken after the CWS one.

---

*[cms/server/contest/handlers/taskusertest.py, line 156 at r2](https://reviewable.io:443/reviews/cms-dev/cms/761#-KkBA57LWvNU_W2qhUdH:-KkIHB210UdSBd6TPpa5:b-q2c53g) ([raw file](https://github.com/cms-dev/cms/blob/4c2ec806755522aa70983bbd4436988e7f114d06/cms/server/contest/handlers/taskusertest.py#L156)):*

Is this the same behaviour? Wouldn't you create testing?task_name=<task.name> instead of testing?<task.name>
</blockquote></details>

Nice catch, I forgot to update the code that reads that value back. Done now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/761#-:-KkIDF7ttp29CFnq2tfs:b-i3gvn4)*

lerks:
Back to you, @stefano-maggiolo.

stefano-maggiolo:
I found a couple of problems, none due to your changes though.

The first you should fix here because it's related: for some reason when the admins answer (or ignore) a question, the redirect URL is encoded by the frontend in the form instead of being computed in the backend. Not sure why, but it seems the only place, and it breaks when aws is served from non-root.

The second is that when I try to change locale in CWS it automatically goes back to automatic, but let's investigate elsewhere. 

After the fix of the first problem, and flake8 fixes, feel free to submit, thanks!

---

Review status: 0 of 105 files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/761#-:-KkgR6kjaDnjJSujE6t3:b-bicwb)*

lerks:
Done and pushed.

BTW, I can't reproduce your issue with locale selection in CWS.

PR #768, opened 2017-05-20
ability to store private files in admin

PJeBeK:
I add this ability to store all files that admin wants to save for a task but it wouldn't shown to contestants and wouldn't run with other files.


stefano-maggiolo:
Hi @PJeBeK! What did you use this for?

wil93:
@stefano-maggiolo OK, I asked, and apparently the rationale behind it is: some files (like checkers) are only stored in binary format, thus it's not easy to review them during a contest because you'd have to find the source code somewhere on your hard drive. It's easier to have it as a "private attachment" so you can review it directly from AWS.

PR #769, opened 2017-05-20
add a message in submission page of output only tasks

PJeBeK:
add message "You may submit any subset of outputs in a single submission." in submission page of output only tasks

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/769)

### DIFF
--- a/cms/grading/TaskType.py
+++ b/cms/grading/TaskType.py
@@ -6,6 +6,7 @@
 # Copyright © 2010-2017 Stefano Maggiolo <s.maggiolo@gmail.com>
 # Copyright © 2010-2012 Matteo Boscariol <boscarim@hotmail.com>
 # Copyright © 2012-2014 Luca Wehrstedt <luca.wehrstedt@gmail.com>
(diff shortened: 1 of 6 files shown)


lerks:
I agree that it's nice to show a message to remind user that they have the option of submitting a partial solution. Yet that behavior is already controlled by another flag: ALLOW_PARTIAL_SUBMISSION. Instead of adding a new attribute to TaskTypes you could just add a conditional in the HTML page that shows that message if that flag is True for the TaskType being used.

Also, you need to allow that message to be localized. And it's generally bad practice to insert style annotations in backend values, and then insert them in the page unescaped.

stefano-maggiolo:
My take: https://github.com/stefano-maggiolo/cms/commit/e7f5940a80be9c6743a26ae91ce6e67f2e6b78a5
2017-08-19-225040_893x387_scrot

@lerks  you good with this?

stefano-maggiolo:
Merged the alternative after @lerks  agreed in private.

PR #770, opened 2017-05-23
Fix syntax error introduced in 207825c

lucach:
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/770)


lerks:
Thanks! (and sorry!)

PR #772, opened 2017-05-28
Exporting submissions grouped by username

myungwoo:
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/772)

### DIFF
--- a/cmscontrib/ExportSubmissions.py
+++ b/cmscontrib/ExportSubmissions.py
@@ -4,6 +4,7 @@
 # Contest Management System - http://cms-dev.github.io/
 # Copyright © 2015-2016 William Di Luigi <williamdiluigi@gmail.com>
 # Copyright © 2016-2017 Stefano Maggiolo <s.maggiolo@gmail.com>
+# Copyright © 2017 Myungwoo Chun <mc.tamaki@gmail.com>
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as
@@ -182,7 +183,7 @@ def main():
 
         print("%s file(s) will be created." % len(results))
         if raw_input("Continue? [Y/n] ").lower() not in ["y", ""]:
-            sys.exit(0)
+            return 0
 
         done = 0
         for row in results:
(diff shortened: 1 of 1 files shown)


myungwoo:
If there's any opinion, such as '--grouped-user' instead of '--by-user' or better help message, etc. Please just tell me.

wil93:
Isn't this use-case already supported? It should be enough to pass `--format "{user}/{id}.{name}{ext}"`, but maybe that would not work when the directory doesn't exist (we could try to create it, if it fails).

myungwoo:
I didn't realize that.

How about my new commit? It makes subdirectories with [os.makedirs](https://docs.python.org/2/library/os.html#os.makedirs) python2 function.

stefano-maggiolo:
Idea is good, but I'd add two checks: one that the filename pattern doesn't start with `os.sep` and second that there are no `..` in the format (you can iteratively call `os.path.split()`, or maybe it is enough to use `split(os.sep)`). Thanks!

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/772#-:-KlN2QKiy9u9U3BwuPSQ:b-bf3azs)*

myungwoo:
Sorry for late response.

I've tested with two situations you mentioned above.
Firstly, filename starts with `os.sep` works well because result of `os.path.join('test123', '/home/user/1.cpp')` is `'/home/user/1/.cpp'`. It ignores first argument.
Secondly, even if filename contains `..` [os.makedirs](https://docs.python.org/2/library/os.html#os.makedirs) works well.

I'm not sure what you want is warning message or something, however, the tests were successful.

---

Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/772#-:-Kpqx1m0xcp7YK292LWG:b-jax3yp)*

stefano-maggiolo:
:lgtm:

---

Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/772#-:-Kpr1YMYJ6VRBae_4HpG:bnfp4nl)*

PR #782, opened 2017-06-02
Force checker results to be ASCII

wil93:
Sometimes an italian message can screw with a remote ssh terminal
because of accents and stuff.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/782)

### DIFF
--- a/cmstaskenv/Test.py
+++ b/cmstaskenv/Test.py
@@ -40,6 +40,7 @@ from cms.grading.tasktypes import get_task_type
 from cms.service.esoperations import ESOperation
 from cmscommon.terminal import move_cursor, add_color_to_string, \
     colors, directions
+from unidecode import unidecode
 
 
 # TODO - Use a context object instead of global variables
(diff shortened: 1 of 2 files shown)


stefano-maggiolo:
Sorry, but not really a fan of this:
1. if your terminal gets screwed up by non-ascii characters is not CMS's fault;
2. more importantly you're changing text that maybe you wanted to match exactly in other places (this is the checker's output, right?); i.e., I see people wasting time trying to understand why the output they see here is different from what they expect;
3. if the checker output is in a language using an alphabet very different from latin, unidecode might give a completely garbage output.

wil93:
Mmm this should only affect `cmsMake`'s output though, shouldn't it?

The checker's output will be untouched, it's only processed right before showing it in `cmsMake`

lerks:
I totally agree with Stefano and I don't understand the argument that this only applies to cmsMake.

PR #789, opened 2017-08-02
[ExportSubmission] Support output-only task & task name in format

myungwoo:
1) `s_language` for an output-only task is None. ---> Fixed extension for an output-only task as `.txt`.
2) Make template of `.txt` same as `.cpp`.
3) Add task name to file format variable named `task`.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/789)


stefano-maggiolo:
Merged with a comment to warn that the additional data is a semantic change for output only submissions.

PR #790, opened 2017-08-05
Do not ignore unrecognized non-trivial lines in gen/GEN

### LINKED ISSUE #779: Improve parsing of non-trivial lines in `gen/GEN`
Consider the following `gen/GEN`:

```
# Subtask 1: Examples
#ST: 0
#COPY testo/input0.txt
#COPY testo/input1.txt
```

The first line is a comment and it's correctly ignored by `cmsMake`. The second one is correctly interpreted as "new subtask with 0 points".
The third and the fourth, **while being non-trivial** (as they do not start with `#<space>`), are silently ignored because of missing `:` after `#COPY`.

At the bare minimum we should log a warning.

lucach:
Fix #779 
This makes cmsMake a bit more strict while parsing `gen/GEN` directives, but I strongly think that advantages far exceed disadvantages.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/790)

### DIFF
--- a/cmstaskenv/cmsMake.py
+++ b/cmstaskenv/cmsMake.py
@@ -77,6 +77,7 @@ def detect_data_dir():
         if os.path.exists(_dir):
             return os.path.abspath(_dir)
 
+
 DATA_DIR = detect_data_dir()
 
 
@@ -340,6 +341,7 @@ def iter_GEN(name):
 
         else:
             testcase, comment = splitted
+            is_trivial = comment.startswith(" ")
             testcase = testcase.strip()
             comment = comment.strip()
             testcase_detected = testcase != ''
@@ -349,9 +351,15 @@ def iter_GEN(name):
             flags = [testcase_detected,
                      copy_testcase_detected,
                      subtask_detected]
-            if len([x for x in flags if x]) > 1:
+
+            flags_count = len([x for x in flags if x])
+
+            if flags_count > 1:
                 raise Exception("No testcase and command in"
                                 " the same line allowed")
+
+            if flags_count == 0 and not is_trivial:
+                raise Exception("Unrecognized non-trivial line")
 
             if testcase_detected:
                 yield (False, testcase, st)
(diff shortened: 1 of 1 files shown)


lerks:
I'm sorry, I don't understand what you're trying to achieve here. Maybe it's just that I am unfamiliar with the format. Could you explain better, maybe by adding comments?

lerks:
Sorry, I only just looked at the issue you referenced and understood. I must say that the file format is very brittle. I would see myself typing the following line:
```
# ST: 0
```
which would be ignored even with your fix.

While my dream is for the format to change I understand that this won't happen anytime soon. In the meantime we could make the parsing more robust, though. For example we could consider the part after the first `#`, find the first sequence of letters in it (ignoring all whitespaces, punctuation, etc.) and if it is `ST` or `COPY` interpret it as a command. If we fail to parse the rest of that line as a command we raise an error. How does that sound?

lucach:
> which would be ignored even with your fix.

Yes, but that's one more level of depth.
As a general rule, we used to distinguish between:
- trivial lines (`#` followed by a whitespace), intended as comments
- non-trivial lines (`#` followed by a keyword), intended as commands

This PR wanted to ensure that non-trivial lines are not silently skipped (I already screwed up a contest due to this).
IIUC your proposal introduces a relatively heavy change. I'm not completely against it, but in that case we should at least ping @wil93 .

lerks:
I really don't know your format, but I think it shouldn't be so dependent on whitespace.

stefano-maggiolo:
Discussions about the format are irrelevant here, imho. Merging the change since it makes a bad situation slightly less bad.

stefano-maggiolo:
Merged.

lerks:
Right, I derailed it, sorry.

For the record, my concrete proposal was something like:
```py
text, _, comment = line.partition('#')
commands = re.findall('[A-Za-z]+', comment)
if commands[0] == "COPY":
    ...
elif commands[0] == "ST":
    ...
else:
    ...
```

PR #793, opened 2017-08-14
Add internationalization of decimal separator

andreyv:
This pull request adds internationalization of the decimal separator as seen in the decimal fractions in ContestWebServer. Since `tornado.locale` doesn't seem to offer this functionality, it is implemented directly in CMS instead. See https://gitter.im/cms-dev/cms?at=58b5d7e000c00c3d4f9f083d for a related discussion.

A new class `cms.locale.Formatter`, derived from `string.Formatter`, is added. Each instance of the class receives the current user locale and formats strings accordingly. Relevant places in the code are converted to use this class instead of raw formats like `"%g"`.

A previous version of this pull request used a `format_decimal()` function instead of a formatter class (see commit history).

I added a few languages that I know use comma as the separator, and for all other languages the default, period, is selected.

TODO (pending soon):
- [x] Re-generate the `.pot` file
- [x] Try to update existing translations in-line (`%g seconds` → `{0:g} seconds` etc.)

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/793)


stefano-maggiolo:
Thanks for working hard to work around a missing feature! I have one possible idea to avoid two problems: if we had a translatable message from which we could extract the comma, then we wouldn't need to have the hardcoded list of languages, and also we wouldn't need to change the API for score type's format_score (since the formatter could take only _ instead of the locale). In that case you could make the formatter for a given language a singleton to avoid having to recompute all the time the separator.

For example the string could be "Translate 0.123 to use the correct decimal separator" or something like this.

---

Reviewed 4 of 15 files at r1, 1 of 12 files at r3, 11 of 11 files at r5, 1 of 1 files at r6, 16 of 16 files at r7.

---

*[cms/locale/locale.py, line 136 at r2](https://reviewable.io:443/reviews/cms-dev/cms/793#-KrwaVS4IDLICedujxrC:-KrwaVS4IDLICedujxrD:bf8dia0) ([raw file](https://github.com/cms-dev/cms/blob/e77313afd0ba6b349884d37e1fdd94176bfa9c77/cms/locale/locale.py#L136)):*
> ```Python
>     trans.pgettext = pgettext
> 
>     # Add the "code" field
> ```

This was already clear from the code, but I still don't know what the code field does :)

---

*[cms/server/util.py, line 255 at r2](https://reviewable.io:443/reviews/cms-dev/cms/793#-Krwast7tT4WKrtAGxAv:-Krwast7tT4WKrtAGxAw:bace4a9) ([raw file](https://github.com/cms-dev/cms/blob/e77313afd0ba6b349884d37e1fdd94176bfa9c77/cms/server/util.py#L255)):*
> ```Python
> 
> 
> COMMA_LANGS = [ "et", "lv", "lt", "ru" ]
> ```

"it" is also one of these.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/793#-:-Krwdukx8sQlwYGNpIA1:b-xc1nfh)*

andreyv:
I think it is a good idea. It moves localization data out from the code to `.po` files, and translators will be more likely to add a correct value for their language.

To clarify, this would mean changing all places where `locale` is now passed (back) to `_`, including the `Formatter` class and the `locale_format` function. Is this what you had in mind?

I didn't understand the part about singletons: the `Formatter` class still needs to receive the translation function in its initializer. Did you mean maintaining a cache dict of `Formatter` classes per translation function (or maybe per each decimal separator)?

stefano-maggiolo:
Yes, that's exactly what I meant. As for Formatter, I was proposing to cache the Formatter based on the translation of that particular string, so that you don't have to extract the separator from every time you want to translate a number. For example using __new__ like in http://python-3-patterns-idioms-test.readthedocs.io/en/latest/Singleton.html (just you'll have a map instead of a single instance).

andreyv:

---

*[cms/locale/locale.py, line 136 at r2](https://reviewable.io:443/reviews/cms-dev/cms/793#-KrwaVS4IDLICedujxrC:-Ks4A2jUlcgSH-VtqICo:bhgq6xl) ([raw file](https://github.com/cms-dev/cms/blob/e77313afd0ba6b349884d37e1fdd94176bfa9c77/cms/locale/locale.py#L136)):*

This was already clear from the code, but I still don't know what the code field does :)
</blockquote></details>

Removed in favor of `_(".")`.

---

*[cms/server/util.py, line 255 at r2](https://reviewable.io:443/reviews/cms-dev/cms/793#-Krwast7tT4WKrtAGxAv:-Ks4AZD02T6FjYpLxSk_:b9eoe38) ([raw file](https://github.com/cms-dev/cms/blob/e77313afd0ba6b349884d37e1fdd94176bfa9c77/cms/server/util.py#L255)):*

"it" is also one of these.
</blockquote></details>

Added a translation of `"."` in the Italian `cms.po`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/793)*

andreyv:
I replaced the explicit list of comma separator languages with a translatable `"."` string. The `locale` parameter is replaced with a translation function.

I didn't add a cache yet, because this would mean either "leaking" some `Formatter` logic from `__init__` to `__new__` and passing the decimal separator as a parameter to `__init__` (`Formatter` could also be expanded later…), or using `_` as a key, for which I see no guarantee there is only one function object per language. Probably using `locale.code` (locale's language code) would have been cleaner, but `locale` is now removed from parameters.

stefano-maggiolo:
:lgtm:

---

Review status: 12 of 31 files reviewed at latest revision, 2 unresolved discussions.

---

*[cms/locale/locale.py, line 192 at r10](https://reviewable.io:443/reviews/cms-dev/cms/793#-Ks63sNFh8AIk16Xwbwn:-Ks63sNFh8AIk16Xwbwo:bq8cvds) ([raw file](https://github.com/cms-dev/cms/blob/c527c4a5adac30a3ab95474f4f32070a43ec4061/cms/locale/locale.py#L192)):*
> ```Python
> 
>         """
>         # translators: decimal mark in fractions (e.g., in "3.14")
> ```

Thanks! I was looking for how to do this (and did not find it, whence my verbose translatable message to explain it to translators. Since we have this, there is no obvious need to cache formatters given that the retrieval of the decimal mark is trivial.

Only nit is that "fractions" might be confusing here, I will change it to "decimal numbers" 

---

*[cms/server/util.py, line 255 at r2](https://reviewable.io:443/reviews/cms-dev/cms/793#-Krwast7tT4WKrtAGxAv:-Ks65SBHiZaIWFfBjBG5:b-ucfknh) ([raw file](https://github.com/cms-dev/cms/blob/e77313afd0ba6b349884d37e1fdd94176bfa9c77/cms/server/util.py#L255)):*
<details><summary><i>Previously, andreyv (Andrey Vihrov) wrote…</i></summary><blockquote>

Added a translation of `"."` in the Italian `cms.po`.
</blockquote></details>

Thanks!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/793#-:-Ks69DvfSUIXu4yVYrH1:bnfp4nl)*

stefano-maggiolo:
Merged

andreyv:
Thanks!

PR #794, opened 2017-08-17
Add multicontest autorestart support to ResourceService

karliss:
ProxyService still requires contest_id, nothing changed there. Code could be slightly cleaner if it wasn't necessary to maintain backwards compatibility with previous CLI, but I think it is no worse than current master and makes testing cms in multi-contest mode easier.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/794)

### DIFF
--- a/cms/service/ResourceService.py
+++ b/cms/service/ResourceService.py
@@ -176,7 +176,7 @@ class ResourceService(Service):
     upon request.
 
     """
-    def __init__(self, shard, contest_id=None):
+    def __init__(self, shard, contest_id=None, autorestart=False):
         """If contest_id is not None, we assume the user wants the
         autorestart feature.
 
@@ -184,6 +184,7 @@ class ResourceService(Service):
         Service.__init__(self, shard)
 
         self.contest_id = contest_id
+        self.autorestart = autorestart or (contest_id is not None)
 
         # _local_store is a dictionary indexed by time in int(epoch)
         self._local_store = []
@@ -195,7 +196,7 @@ class ResourceService(Service):
         self._local_services = self._find_local_services()
         # Dict service with bool to mark if we will restart them.
         self._will_restart = dict((service,
-                                   None if self.contest_id is None else True)
+                                   None if not self.autorestart else True)
                                   for service in self._local_services)
         # Found process associate to the ServiceCoord.
         self._procs = dict((service, None)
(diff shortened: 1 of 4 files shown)


stefano-maggiolo:
Apart from the change to the command line, there is another problem: now `cmsResourceService -a 1` does not work (it says there is no such contest, but with `cmsResourceService -a` I can see and use it).

---

Reviewed 4 of 4 files at r1.

---

*[scripts/cmsResourceService, line 52 at r1](https://reviewable.io:443/reviews/cms-dev/cms/794#-Ks0jP3WBEJSWCQf09dl:-Ks0jP3Xp_NzYZewAI0K:bh0hcai) ([raw file](https://github.com/cms-dev/cms/blob/4835c72e3f0b81b926659039fb7729cf37f4fc28/scripts/cmsResourceService#L52)):*
> ```
>                         dest="contest_id",
>                         help="restart automatically services on its machine")
>     parser.add_argument("-r", dest="autorestart", action="store_true",
> ```

Can this be a `-a ALL` instead of `-r`? I think as it is now you can do weird stuff like `-a 1 -r` which does not make sense.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/794#-:-Ks0llWrjAHV6fRei7Vc:b8sf8cn)*

karliss:
Both problems solved by using contest_id_from_args.

---

Review status: 3 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[scripts/cmsResourceService, line 52 at r1](https://reviewable.io:443/reviews/cms-dev/cms/794#-Ks0jP3WBEJSWCQf09dl:-Ks5LVf1d5u_25mQQQz1:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/4835c72e3f0b81b926659039fb7729cf37f4fc28/scripts/cmsResourceService#L52)):*

Can this be a `-a ALL` instead of `-r`? I think as it is now you can do weird stuff like `-a 1 -r` which does not make sense.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/794#-:-Ks5Lb507fzb8GyQ4k5-:b-m8r26f)*

stefano-maggiolo:
:lgtm: Will only change the help for `autorestart` to make clear ALL is a possibility.

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/794#-:-Ks6-nWppXkd9uK--7Lu:b-x6acrf)*

stefano-maggiolo:
Do you like this? https://github.com/stefano-maggiolo/cms/commit/44f4eb950c999f4de34fb79428f94e29f7169fe3

karliss:
No objections against 44f4eb9

stefano-maggiolo:
Merged.

PR #795, opened 2017-08-19
Sort AWS rankings by score

andreyv:
It is often useful for admins to see the current ranking by score, especially if RWS is not running. This pull request changes the sorting order for rankings generated by AdminWebServer.

Previously, task scores and total score for each participation were calculated during output generation. Now this is moved up and before output begins. This allows to use the calculated total score as the sorting key, at the cost of one additional list object per participation. A side effect is that score calculation and rounding are joined in one place from three separate places.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/795)


stefano-maggiolo:
Reviewed 3 of 3 files at r1.

---

*[cms/server/admin/handlers/contestranking.py, line 70 at r1](https://reviewable.io:443/reviews/cms-dev/cms/795#-Ks0iN2kSuiXIhxMfj2U:-Ks0iN2kSuiXIhxMfj2V:b-9tem35) ([raw file](https://github.com/cms-dev/cms/blob/58b747953e691464aefc250e80131e3c94ae8369/cms/server/admin/handlers/contestranking.py#L70)):*
> ```Python
>             show_teams = show_teams or p.team_id
> 
>             p.scores = []
> ```

This is a bit hacky, as `scores` and `total_score` are not declared fields of a participation... I'm inclined to allow this given it's pretty local usage, but could you add a comment to the template (at least the html) saying that you expect participations to have these nonstandard fields? And maybe a todo that sorting should be done client side (if ever we move to some widget library).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/795)*

andreyv:
Review status: 1 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[cms/server/admin/handlers/contestranking.py, line 70 at r1](https://reviewable.io:443/reviews/cms-dev/cms/795#-Ks0iN2kSuiXIhxMfj2U:-Ks5Vc739QAzLccQpF-5:bfpjmx1) ([raw file](https://github.com/cms-dev/cms/blob/95cfce94b947e9c97c39ae6b30e2eb2180fd7f23/cms/server/admin/handlers/contestranking.py#L70)):*

This is a bit hacky, as `scores` and `total_score` are not declared fields of a participation... I'm inclined to allow this given it's pretty local usage, but could you add a comment to the template (at least the html) saying that you expect participations to have these nonstandard fields? And maybe a todo that sorting should be done client side (if ever we move to some widget library).
</blockquote></details>

Added some comments. Do they look good?

Some alternatives to custom object fields would be either creating a new array of `(participation, score, total_score)` and sorting/printing that, or adding some methods (with caching) to `Participation`, but both of these would require more code or resources.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/795)*

wil93:
What about just plugging this into the existing `<table>`?

https://datatables.net/

stefano-maggiolo:
@wil93 Agree, if we ever move AWS to some more modern FE framework I'd prefer this to be client side, but for the moment if we have to pick an order, by score is probably more interesting than by username.

stefano-maggiolo:
Reviewed 2 of 2 files at r2.

---

*[cms/server/admin/handlers/contestranking.py, line 70 at r1](https://reviewable.io:443/reviews/cms-dev/cms/795#-Ks0iN2kSuiXIhxMfj2U:-Ks6LjKxHLd7mBOJeQsV:bnbynl) ([raw file](https://github.com/cms-dev/cms/blob/95cfce94b947e9c97c39ae6b30e2eb2180fd7f23/cms/server/admin/handlers/contestranking.py#L70)):*
<details><summary><i>Previously, andreyv (Andrey Vihrov) wrote…</i></summary><blockquote>

Added some comments. Do they look good?

Some alternatives to custom object fields would be either creating a new array of `(participation, score, total_score)` and sorting/printing that, or adding some methods (with caching) to `Participation`, but both of these would require more code or resources.
</blockquote></details>

TODO is good, for the other I was thinking more on the line of "this template requires this and this additional field in the participation". I'll change it, hope you don't mind.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/795)*

stefano-maggiolo:
Merged

PR #796, opened 2017-08-20
Show max used time and memory in AWS submissions

### LINKED ISSUE #720: Show max used time and memory in AWS submissions
This would be useful, for example, to see how good does a model solution perform on the server's hardware, and set appropriate limits for the task. Right now this requires going through all test rows to find the maximum values manually.

andreyv:
This change adds maximum used time and memory to the submission details page in AWS.

Both `SubmissionResult.evaluations` and `SubmissionResult.score_details` could be used to calculate these values. Here `evaluations` is used for simpler code, however, it incurs a database query. This is not a problem now since the `templates/submission.html` queries `sr.evaluations` later anyway.

Closes #720.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/796)

### DIFF
--- a/cms/db/submission.py
+++ b/cms/db/submission.py
@@ -424,6 +424,20 @@ class SubmissionResult(Base):
             .filter(Evaluation.testcase == testcase)\
             .first()
 
+    def get_used_resources(self):
+        """Return the maximum time and memory used by this SR.
+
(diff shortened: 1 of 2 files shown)


stefano-maggiolo:
Merged with minor wording edits, thanks (it was also in my post-IOI todo list)!

PR #800, opened 2017-09-24
Update Latvian translation

andreyv:
CC @karliss

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/800)

### DIFF
--- a/cms/locale/lv/LC_MESSAGES/cms.po
+++ b/cms/locale/lv/LC_MESSAGES/cms.po
@@ -9,7 +9,7 @@ msgstr ""
 "Project-Id-Version: 0.1\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2017-08-21 15:35+0300\n"
-"PO-Revision-Date: 2017-01-30 17:19+0200\n"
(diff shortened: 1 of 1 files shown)


lerks:
Could you rebase and squash the four commits into one? Thanks!

karliss:
Translation for "Official" submission looks suspicious. I can't think of better translation yet. All the names I can think of are long and mean the opposite.

lerks:
@karliss I didn't understannd if you are OK with the translation proposed by @andreyv or if you would like some more time to propose a better one.

andreyv:
This occurrence is quite hard to translate (would be much easier if it was "Unofficial" instead).

@karliss Do you like either of "Ieskaitīts" or "Pieņemts vērtēšanai"? There was also a suggestion "Sacensībās".

karliss:
All three seem better than "Oficiāls".

andreyv:
Updated the translation with "Ieskaitīts", it can be improved later if needed.

andreyv:
This pull request should be good to merge now.

stefano-maggiolo:
Thanks!

PR #801, opened 2017-09-24
Enable optimization in Python language

andreyv:
See https://docs.python.org/2/using/cmdline.html#cmdoption-o for the description.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/801)

### DIFF
--- a/cms/grading/languages/python2_cpython.py
+++ b/cms/grading/languages/python2_cpython.py
@@ -52,9 +52,9 @@ class Python2CPython(CompiledLanguage):
                                  source_filenames, executable_filename,
                                  for_evaluation=True):
         """See Language.get_compilation_commands."""
-        py_command = ["/usr/bin/python2", "-m", "py_compile",
+        py_command = ["/usr/bin/python2", "-O", "-m", "py_compile",
                       source_filenames[0]]
-        mv_command = ["/bin/mv", "%s.pyc" % os.path.splitext(os.path.basename(
+        mv_command = ["/bin/mv", "%s.pyo" % os.path.splitext(os.path.basename(
                       source_filenames[0]))[0], executable_filename]
         return [py_command, mv_command]
 


lerks:
I don't see any reason we wouldn't want this, but then I am wondering why we didn't activate it in the beginning. Does @stefano-maggiolo know? What is the performance gain?

stefano-maggiolo:
No idea, I didn't know this existed. From a random stackoverflow search, it seems it won't change much especially for competition-style programs:

`
when invoking the python interpreter, you may pass the -O or -OO option to generate a .pyo file. Using -O will throw out the line numbers, assertions, and some debugging info.  -OO will result in a .pyo file also stripped of docstrings.
`

No objections to specify -O or -OO

karliss:
if it only removes asserts and docstrings is it worth changing default behaviour? It is not like participants will spend time writing large amounts of docstrings or asserts during contest. Shouldn't it be handled the same way as NDEBUG for c++ which cms currently  doesn't define. Having asserts seems like easy way of checking that basic  assumptions about input or task are correct and avoid wasting time debugging rest of the solution when assumptions are wrong. I suggest not enabling it at least until it enables some meaningful optimizations. This is not  c++ where STL is 50 times slower without optimizer inclining templates.

stefano-maggiolo:
That's a good point for assertions - I think it's fine to remove them as long as it's clear in the contest rules, but it is something that contest admins might overlook very easily. For keeping surprise low might be better to avoid merging this (and to have a more optimized python eval you can always create a new language for pypy :)

stefano-maggiolo:
@andreyv wdyt?

lerks:
Does a Python program that dies because of a failed assertion show up differently in CWS than one that fails because of an uncaught exception or with an explicit error code? I expect not but just checking.

andreyv:
That's a good point about assertions. I agree that if `NDEBUG` is not defined, then this flag probably should not be used too. One difference is that in C++ you can use

```
#ifdef EVAL
#  define NDEBUG
#endif
```

for local-only debugging, whereas here you can't.

I compared `.pyc` and `.opt-1.pyc` for some Python 3.6 packages, and they had mostly the same sizes, in many cases even being identical to each other. So it looks like there is not much improvement from `-O`.

andreyv:
If there is no agreement, feel free to close this pull request. On one hand, `-O` is closer to "production" code, on the other hand, the improvements are minimal and one could argue that PyPy should be introduced instead if greater performance is needed.

stefano-maggiolo:
Thanks.

andreyv:
I recently realized that there could still be a use case for `-O`. While in terms of optimization it doesn't seem to do much, it would effectively allow to distinguish whether the code is running on the server or on the contestant's PC, the same way `-DEVAL` works for C/C++.

So one could write
```python
if __debug__:
  do_expensive_checks()
  print_a_lot_of_debug_info()
```

and send such code to the server unchanged.

This can be considered in the future if there is such a need.

stefano-maggiolo:
Honestly even -DEVAL is a relic of the past. We used to have it 20 years ago in the system we used in Italy and it percolated in CMS. I didn't think it was something people used until seeing how much @andreyv wants it :)

If there is an equivalent for Python I'm happy to add it, but optimizing just to get this seems a bit counter-intuitive.

Issue #803, opened 2017-10-16
How add python3 in CMS1.3


jonhte:
Hi all.

I am an user of CMS1.3 (https://cms-dev.github.io).

I have a questions:
How add Python3 in CMS1.3.

Please guide step by step.

Thanks a lot.

stefano-maggiolo:
Hi @jonhte - I think @andreyv was working on this. Andrey, do you have something we can submit or at least share?

andreyv:
Sorry, I didn't look further than initial evaluation.

Some notes cat be found at https://gitter.im/cms-dev/cms?at=59c76356cfeed2eb65441332.

stefano-maggiolo:
Thanks! I can probably cook up something even if maybe not perfect.

jonhte:
I read: https://www.freelists.org/post/contestms/Wall-clock-Python3-and-Nginx-in-CMS13

I copy file python2_cpython.py ==> python3_cpython.py, and change cmd line python2 to python3 inside python3_cpython.py.  (in folder cms/grading/languages)
**But everything not change**

PR #805, opened 2017-10-23
Python 3 & better support for isolate.

stefano-maggiolo:
Sorry for two commits not fully related...

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/805)

### DIFF
--- a/cms/grading/Sandbox.py
+++ b/cms/grading/Sandbox.py
@@ -734,6 +734,15 @@ class StupidSandbox(SandboxBase):
         """
         return True
 
+    def cleanup(self):
+        """Cleanup the sandbox.
+
+        To be called at the end of the execution, regardless of
+        whether the sandbox should be deleted or not.
+
+        """
+        pass
+
     def delete(self):
         """Delete the directory where the sandbox operated.
 
@@ -842,14 +851,17 @@ class IsolateSandbox(SandboxBase):
         if os.path.isdir("/etc/alternatives"):
             self.add_mapped_directories(["/etc/alternatives"])
 
-        # Tell isolate to get the sandbox ready.
-        box_cmd = [self.box_exec] + (["--cg"] if self.cgroup else []) \
+        # Tell isolate to get the sandbox ready. We do our best to
+        # cleanup after ourselves, but we might have missed something
+        # if the worker was interrupted in the middle of an execution.
+        self.cleanup()
+        init_cmd = [self.box_exec] + (["--cg"] if self.cgroup else []) \
             + ["--box-id=%d" % self.box_id] + ["--init"]
-        ret = subprocess.call(box_cmd)
+        ret = subprocess.call(init_cmd)
         if ret != 0:
             raise SandboxInterfaceException(
                 "Failed to initialize sandbox with command: %s "
-                "(error %d)" % (pretty_print_cmdline(box_cmd), ret))
+                "(error %d)" % (pretty_print_cmdline(init_cmd), ret))
 
     def add_mapped_directories(self, dirs):
         """Add dirs to the external dirs visible to the sandboxed command.
@@ -1314,16 +1326,25 @@ class IsolateSandbox(SandboxBase):
             raise SandboxInterfaceException("Sandbox exit status (%d) unknown"
                                             % exitcode)
 
+    def cleanup(self):
+        """Cleanup the sandbox.
+
+        To be called at the end of the execution, regardless of
+        whether the sandbox should be deleted or not.
+
+        """
+        # Tell isolate to cleanup the sandbox.
+        subprocess.call(
+            [self.box_exec]
+            + (["--cg"] if self.cgroup else [])
+            + ["--box-id=%d" % self.box_id]
+            + ["--cleanup"])
+
     def delete(self):
         """Delete the directory where the sandbox operated.
 
         """
         logger.debug("Deleting sandbox in %s.", self.path)
-
-        # Tell isolate to cleanup the sandbox.
-        box_cmd = [self.box_exec] + (["--cg"] if self.cgroup else []) \
-            + ["--box-id=%d" % self.box_id]
-        subprocess.call(box_cmd + ["--cleanup"])
 
         # Delete the working directory.
         rmtree(self.outer_temp_dir)
--- a/cms/grading/languages/python3_cpython.py
+++ b/cms/grading/languages/python3_cpython.py
@@ -1,0 +1,67 @@
+#!/usr/bin/env python2
+# -*- coding: utf-8 -*-
+
+# Contest Management System - http://cms-dev.github.io/
+# Copyright © 2016-2017 Stefano Maggiolo <s.maggiolo@gmail.com>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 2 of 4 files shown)


stefano-maggiolo:
@andreyv @jonhte - closes #803

lerks:
I'll look into this shortly, hopefully by tomorrow.

FYI, the Travis error isn't your fault. I just triggered a new test on master and now it fails there too, in the same way. The error appears to be due to the fact that the version number of Postgres 10.0 (released a couple of weeks ago, but which as far as I can tell we are not using) cannot be correctly parsed by Psycopg2. Do you have any ideas?

stefano-maggiolo:
Mmm, [this](http://oshanebailey.jamaicandevelopers.com/2017/10/11/error-could-not-determine-postgresql-version-10-0/) seems relevant... might try to play with it tonight and see if that's the reason.

lerks:
Also the fact that we're telling Travis that it should setup an environment for Ruby may mess a few things up.

stefano-maggiolo:
https://github.com/psycopg/psycopg2/issues/489 is the actual source of truth - regex to detect version was incorrect, was fixed in 2.7.* but not backported to 2.6. Bumping from 2.6 to 2.7 makes the build green again.

I think our decision was to keep versions of python packages in sync with the latest LTS versions, but I think we can make an exception here. Opinions? In case I can add it to the PR.

lerks:
Nice find, that's the bug we're hitting! Apparently they didn't backport the fix to the 2.6 branch. We could either ask them to do that (and then depend on the next micro they release) or bump to 2.7. I'm OK with bumping (but please add a comment in requirements.txt explaining the reason). This could create issues to people running on the latest Ubuntu LTS using native packages (I know we recommend PIP and that it's sort-of their fault, but Ubuntu used to be our recommended environment and we pledged to support it out-of-the-box).

cms/grading/languages/python3_cpython.py
```diff
@@ -0,0 +1,67 @@
...
+        return [".py"]
+
+    def get_compilation_commands(self,
+                                 source_filenames, executable_filename,
+                                 for_evaluation=True):
+        """See Language.get_compilation_commands."""
+        basename = os.path.splitext(os.path.basename(source_filenames[0]))[0]
+        compilation_program = ";".join([
+            "import py_compile as m",
+            "m.compile(\"%s.py\", \"%s\", doraise=True)" % (
```
lerks: Here we're injecting unescaped strings into a shell command. It's true that they are coming from the admin, and it's true that they are run in a sandbox (are they?), though maybe we could escape them anyway? What do you think? Is the risk worth the effort?

andreyv: I had thought about this too. One option is manual escaping (is there a built-in Python function for this?), another is to pass arguments on command line:
```
> python -c 'import sys;print(sys.argv[1])' test
test
```

cms/grading/Sandbox.py
```diff
@@ -734,7 +734,17 @@ def translate_box_exitcode(self, _):
...
+        """Cleanup the sandbox.
+
+        To be called at the end of the execution, regardless of
+        whether the sandbox should be deleted or not.
+
+        """
+        pass
+
     def delete(self):
+
```
lerks: Why this empty line before the docstring?

cms/grading/Sandbox.py
```diff
@@ -842,14 +852,19 @@ def __init__(self, multithreaded, file_cacher, temp_dir=None):
...
             self.add_mapped_directories(["/etc/alternatives"])
 
-        # Tell isolate to get the sandbox ready.
-        box_cmd = [self.box_exec] + (["--cg"] if self.cgroup else []) \
+        # Tell isolate to get the sandbox ready. We do our best to
+        # cleanup after ourselves, but we might have missed something
+        # if the worker was interrupted in the middle of an execution.
+        cleanup_cmd = [self.box_exec] + (["--cg"] if self.cgroup else []) \
+            + ["--box-id=%d" % self.box_id] + ["--cleanup"]
+        subprocess.call(cleanup_cmd)
```
lerks: Is the return code interesting?

cms/grading/Sandbox.py
```diff
@@ -1314,17 +1329,24 @@ def translate_box_exitcode(self, exitcode):
...
-        """
-        logger.debug("Deleting sandbox in %s.", self.path)
+        To be called at the end of the execution, regardless of
+        whether the sandbox should be deleted or not.
 
+        """
         # Tell isolate to cleanup the sandbox.
         box_cmd = [self.box_exec] + (["--cg"] if self.cgroup else []) \
             + ["--box-id=%d" % self.box_id]
         subprocess.call(box_cmd + ["--cleanup"])
```
lerks: No need to have this argument be on its own.

cms/grading/Sandbox.py
lerks: And can't these three lines be replaced by a call to self.cleanup()?

cms/grading/languages/python3_cpython.py
lerks: There sure is: `repr(unsafe_string)`.

stefano-maggiolo:
Bumped to 2.7 and added the comment.

---

Review status: 0 of 3 files reviewed at latest revision, 4 unresolved discussions, some commit checks broke.

---

*[cms/grading/Sandbox.py, line 747 at r1](https://reviewable.io:443/reviews/cms-dev/cms/805#-Kx5_UrPEwLdCJ2MlsEQ-r1-747:-Kxd_4Eh522pV7Av6Nid:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/3380066163b1ceee576542f01211ff46be81d952/cms/grading/Sandbox.py#L747)):*

Why this empty line before the docstring?
</blockquote></details>

Done.

---

*[cms/grading/Sandbox.py, line 860 at r1](https://reviewable.io:443/reviews/cms-dev/cms/805#-Kx5_UrPEwLdCJ2MlsEQ-r1-860:-Kxd_fUUACo4HxjpxT6t:b-paq7f8) ([raw file](https://github.com/cms-dev/cms/blob/3380066163b1ceee576542f01211ff46be81d952/cms/grading/Sandbox.py#L860)):*

And can't these three lines be replaced by a call to self.cleanup()?
</blockquote></details>

Yes, replaced. Return code probably not - if there is an error we'll get an error in the subsequent init too, when really we cannot continue.

---

*[cms/grading/Sandbox.py, line 1342 at r1](https://reviewable.io:443/reviews/cms-dev/cms/805#-Kx5_UrPEwLdCJ2MlsEQ-r1-1342:-Kxda4-ERrClwoi4Z7BK:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/3380066163b1ceee576542f01211ff46be81d952/cms/grading/Sandbox.py#L1342)):*

No need to have this argument be on its own.
</blockquote></details>

Done.

---

*[cms/grading/languages/python3_cpython.py, line 58 at r2](https://reviewable.io:443/reviews/cms-dev/cms/805#-Kx5_UrQdI0KeUhyltmo-r2-58:-KxdcloZmy7FUGTOhG43:b-s2rc7i) ([raw file](https://github.com/cms-dev/cms/blob/d5e68da68213299db1416ed2377e2f99bcb3c909/cms/grading/languages/python3_cpython.py#L58)):*

There sure is: `repr(unsafe_string)`.
</blockquote></details>

Good point, thanks!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/805#-:-KxdcrxFhLb0Pl_W1Oyu:b-4dsl05)*

lerks:
:lgtm:

---

Reviewed 1 of 2 files at r1, 1 of 1 files at r3, 1 of 1 files at r4, 1 of 1 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/805#-:-Kxddj-FkwiG0pBfPZ8l:bnfp4nl)*

PR #809, opened 2017-11-26
Sanitize log in admin overview.

karliss:
Log messages might contain all kind of text. Use jquery instead of string concatenation to avoid interpreting it as html.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/809)


lerks:
Merged as part of #813.

PR #813, opened 2017-12-18
Logs in AWS are between <pre> tags

edomora97:
Since forever the logs in `cmsAdminWebServer` are shown in simple text format. When something goes wrong a stacktrace is generated, which always has many endlines and tabulation characters. In HTML this chars are collapsed in a single space making the stacktraces very long lines.

The fix I propose is just to put the message inside a `<pre>` tag which
preserves all the spaces.

screen

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/813)

### DIFF
--- a/cms/server/admin/static/aws_style.css
+++ b/cms/server/admin/static/aws_style.css
@@ -335,7 +335,7 @@ table th {
 table td {
     font-size: 0.875em;
     padding: 0 0.5em;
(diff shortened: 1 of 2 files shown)


karliss:
These changes partially overlap with #809 .

lerks:
Looks good to me, thanks!

PR #814, opened 2017-12-18
ES & AWS start up (very) slowly when SYN packets are dropped

edomora97:
## Scenario
During the OIS (IOI-like competition in teams in Italy) the contest is hosted on Google Cloud Engine which has a peculiar policy with unreachable hosts. If a packet is sent to an unreachable host it will be dropped. This causes all the TCP connection to fail by timeout if the host is down.

## Setup
`cms.conf` is configured to "allow" connection from a _big_ number of workers (where _big_ is like 20). In fact we may want to scale up the number of running workers during the contest without restarting all the services.

## Problem
During the boot of some services (e.g. ES and AWS) there is a loop that tries to connect to all the services listed in the config file. ES tries to connect to the workers and AWS to the RS. The problem occurs in this loop:

https://github.com/cms-dev/cms/blob/1635eddb115c0105b3d0ed323c35c3c6b11c6183/cms/service/EvaluationService.py#L95-L97

`self.pool.add_worker` is blocking, and in this scenario the service boots up (as said by `cmsLogService`) but it cannot process any request until the loop finishes. This loop could be very very long since the TCP handshake timeout on linux is pretty high, times the number of host means an infeasible startup time.

In AWS the incriminated loop is this:

https://github.com/cms-dev/cms/blob/1635eddb115c0105b3d0ed323c35c3c6b11c6183/cms/server/admin/server.py#L96-L98

## This solution
Instead of making that blocking call in the main "thread" we can spawn _N_ greenlets that take care of the connections. We (at the OIS) have used this approach for the last 2 rounds without any problems. There is only one side effect that I've seen: the eventlets are spawned, they get blocked at the `connect` call which is done before adding the worker to the list in the pool. That means the worker is not listed in AWS until the `connect` call returns, successfully or not. In this scenario this means that when a worker connects it is almost instantly registered, when it's down it will be listed (as down) after the timeout.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/814)

### DIFF
--- a/cms/server/admin/server.py
+++ b/cms/server/admin/server.py
@@ -33,6 +33,7 @@ from __future__ import unicode_literals
 import base64
 import logging
 import pkg_resources
+import gevent
 
 from sqlalchemy import func, not_
 
@@ -94,8 +95,8 @@ class AdminWebServer(WebService):
 
         self.resource_services = []
         for i in range(get_service_shards("ResourceService")):
-            self.resource_services.append(self.connect_to(
-                ServiceCoord("ResourceService", i)))
+            gevent.spawn(lambda self, i: self.resource_services.append(
+                self.connect_to(ServiceCoord("ResourceService", i))), self, i)
         self.logservice = self.connect_to(ServiceCoord("LogService", 0))
 
     def is_rpc_authorized(self, service, shard, method):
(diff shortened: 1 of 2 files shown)


lerks:
I understand the problem but I'm not sure I like this solution. Let me think a bit...

stefano-maggiolo:
To be on the safe side, we could do the connections in parallel, but join to all of them before proceeding? So you are waiting a long-ish, but constant time.

lerks:
Fixed, differently, in #1014 (sorry it took 200 PR numbers to do).

PR #817, opened 2018-01-05
Fixed corrupted users imported after eaee3a8

edomora97:
The new format of stored passwords (after eaee3a8) requires that in the DB the password is a 2 part string separated by ':'. Before this patch when an old contest (like [cms-dev/con_test](https://github.com/cms-dev/con_test/blob/caa33415d4858a3df11d3b228964c0053ef3b3da/contest.yaml#L15)) is imported the passwords in the db were corrupted (the "method" is not specified) and cmsContestWebServer will fire 500 errors on login.

I fixed the loaders prepending "plaintext:" to the passwords of the users and participations.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/817)

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -25,7 +25,7 @@ Artem Iglikov (aka artikz) <artem.iglikov@gmail.com>
 Damien Leroy <damien.leroy@be-oi.be>
 Tubérculo Martínez <nlehmann@dcc.uchile.cl>
 Evgeny Martynov <evgeny@epochfail.com>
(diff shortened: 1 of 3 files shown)


stefano-maggiolo:
This should probably go in the 1.3 branch as well.

lerks:
Thanks, this is much better! The password will now be stored hashed, but I believe this is acceptable (maybe even desirable?).

lerks:
I also cherry-picked this onto v1.3.

PR #818, opened 2018-01-05
C# language

nir-lavee:
Hi, this PR adds support for C# using the ``mcs`` compiler (see Ubuntu's [``mono-mcs``](https://packages.ubuntu.com/xenial/mono-mcs) package). All tests passed on my local VM.
Feedback appreciated! I hope this will help others.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/818)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ install:
   - sudo apt-get install -y fp-compiler fp-units-base fp-units-fcl fp-units-misc fp-units-math fp-units-rtl
   - sudo apt-get install -y postgresql-client gettext python2.7 iso-codes shared-mime-info stl-manual cgroup-lite libcap-dev
   - sudo apt-get install -y python-dev libpq-dev libcups2-dev libyaml-dev wget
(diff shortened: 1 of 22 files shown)


lerks:
At a first pass this looks good, thanks!

I see that Codacy is unhappy with the C# tests you provided. Could you see whether you can address these issues? Also, I don't really know C# and its style guide, but it seems that spacing around operators is inconsistent in your code.

Moreover, as was discussed in #731, we are in favor of adding languages but we are worried that they are making tests too slow. We were considering having tests run by default only on a subset of languages. I don't know if we should wait for that to happen before we merge this...

lerks:
Also, is there any reason to prefer "C# / MCS" to "C# / Mono"? As an outsider to the C# community I didn't know what MCS was but I knew Mono. Also, it appears that MCS is the only (or at least the main) compiler provided by Mono.

lerks:
Looks great, thanks!

PR #819, opened 2018-01-16
HTML fixes

andreyv:
These changes fix more HTML validation errors with https://validator.w3.org/.

`<html lang="en">` is added for AWS because AWS does not send the `Content-Language` HTTP header.

Note that `cms/server/contest/templates/contest.html` was already fixed a year ago in 08ea65064e34c6fc88ca583ec0253b3d00625782, but now changes need to be reapplied. This is because of a merge error from #594. @wil93 should probably check other files that were moved in that issue for similar problems.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/819)


lerks:
Thanks, I pushed this as 8e1811e86c58799689dfa769ffe42ae70ca80cc6.

PR #820, opened 2018-01-16
Do not close SQL session if it does not exist

### LINKED ISSUE #811: CWS warning "Couldn't close SQL connection" on invalid login form data
Sending an invalid login request to CWS may sometimes result in the above warning. I can reproduce this by removing the `_xsrf` parameter from the to-be-sent POST data and sending such a request.

Actual output:
```
2017-12-03 19:13:21,227 - WARNING [Contest,0 10 web::log_exception] 403 POST /test/login (10.0.2.2): '_xsrf' argument missing from POST
2017-12-03 19:13:21,229 - WARNING [Contest,0 10 base::finish] Couldn't close SQL connection: AttributeError("'NoneType' object has no attribute 'close'",)
2017-12-03 19:13:21,231 - WARNING [Contest,0 10 web::log_request] 403 POST /test/login (10.0.2.2) 3.87ms
```

The first and the third warning appear to be correct, however, the second warning shouldn't be there.

andreyv:
In the `CommonRequestHandler` class (`cms/server/util.py`) an SQL session set to `None` during initialization and later created during `prepare()`. On the other hand, Tornado checks CSRF cookies just before calling `prepare()`, see `check_xsrf_cookie()` in https://github.com/tornadoweb/tornado/blob/master/tornado/web.py.

If an error occurs during the CSRF check, then the `BaseHandler.finish()` function in `
cms/server/contest/handlers/base.py` will try to close an SQL session that does not exist. This can be fixed with an additional check.

Fixes #811.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/820)

### DIFF
--- a/cms/server/contest/handlers/base.py
+++ b/cms/server/contest/handlers/base.py
@@ -103,7 +103,7 @@ class BaseHandler(CommonRequestHandler):
         that. So far I'm leaving it to minimize changes.
 
         """
-        if hasattr(self, "sql_session"):
+        if hasattr(self, "sql_session") and self.sql_session:
             try:
                 self.sql_session.close()
             except Exception as error:


lerks:
Thanks, I cleaned this up and pushed this as ce5c12b4f001f26dc6e25c3987024d830cb7fab5. I did:
- add `is not None`, as it is recommended by the style guide,
- remove useless checks (`hasattr` is always true since the attribute is set in the constructor and never deleted),
- move that method to the superclass to avoid duplication.

PR #821, opened 2018-01-28
Move codebase to Py2+3 [1]

lerks:
This is the first in a series of pull requests that aim at making our codebase compatible with Python 2 and Python 3 at the same time.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/821)

### DIFF
--- a/cms/grading/tasktypes/Communication.py
+++ b/cms/grading/tasktypes/Communication.py
@@ -31,6 +31,7 @@ from future.builtins import *
 import logging
 import os
 import tempfile
+from functools import reduce
 
 from cms import config
 from cms.grading.Sandbox import wait_without_std, Sandbox
--- a/cms/grading/tasktypes/Communication.py
+++ b/cms/grading/tasktypes/Communication.py
@@ -22,8 +22,11 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 from __future__ import absolute_import
+from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
+from future.builtins.disabled import *
+from future.builtins import *
 
 import logging
 import os
--- a/cms/grading/tasktypes/Communication.py
+++ b/cms/grading/tasktypes/Communication.py
@@ -289,7 +289,7 @@ class Communication(TaskType):
         # the outcome is 0.0 and the text describes that problem.
         elif not is_evaluation_passed(plus_user):
             success = True
-            outcome, text = 0.0, human_evaluation_message(plus_user)
+            outcome, text = 0, human_evaluation_message(plus_user)
         # Otherwise, we use the manager to obtain the outcome.
         else:
             success = True
--- a/cms/grading/tasktypes/Communication.py
+++ b/cms/grading/tasktypes/Communication.py
@@ -179,7 +179,7 @@ class Communication(TaskType):
             num_processes = 1
         else:
             num_processes = self.parameters[0]
-        indices = range(num_processes)
+        indices = list(range(num_processes))
         # Create sandboxes and FIFOs
         sandbox_mgr = create_sandbox(file_cacher, job.multithreaded_sandbox)
         sandbox_user = [create_sandbox(file_cacher, job.multithreaded_sandbox)
--- a/cms/grading/tasktypes/Communication.py
+++ b/cms/grading/tasktypes/Communication.py
@@ -289,7 +289,7 @@ class Communication(TaskType):
         # the outcome is 0.0 and the text describes that problem.
         elif not is_evaluation_passed(plus_user):
             success = True
-            outcome, text = 0, human_evaluation_message(plus_user)
+            outcome, text = 0.0, human_evaluation_message(plus_user)
         # Otherwise, we use the manager to obtain the outcome.
         else:
             success = True
--- a/cms/grading/tasktypes/Communication.py
+++ b/cms/grading/tasktypes/Communication.py
@@ -180,7 +180,7 @@ class Communication(TaskType):
             num_processes = 1
         else:
             num_processes = self.parameters[0]
-        indices = list(range(num_processes))
+        indices = range(num_processes)
         # Create sandboxes and FIFOs
         sandbox_mgr = create_sandbox(file_cacher, job.multithreaded_sandbox)
         sandbox_user = [create_sandbox(file_cacher, job.multithreaded_sandbox)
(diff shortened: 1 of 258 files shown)


cmscontrib/DumpImporter.py
```diff
@@ -399,7 +402,7 @@ def add_relationships(self, data, obj):
             val = data[prp.key]
             if val is None:
                 setattr(obj, prp.key, None)
-            elif type(val) == unicode:
+            elif type(val) == str:
```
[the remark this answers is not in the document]
stefano-maggiolo: Please fix

cms/grading/tasktypes/Communication.py
```diff
@@ -31,6 +31,7 @@
 import logging
 import os
 import tempfile
+from functools import reduce
```
stefano-maggiolo: I think there is a newline between imports and froms?

cms/server/admin/handlers/dataset.py
```diff
@@ -42,7 +42,7 @@
 import tempfile
 import zipfile
 
-from StringIO import StringIO
+from io import StringIO
```
stefano-maggiolo: newline after this.

cmstestsuite/unit_tests/db/filecacher_test.py
```diff
@@ -34,7 +34,7 @@
 import io
 import os
 import random
-from StringIO import StringIO
+from io import StringIO
```
stefano-maggiolo: New lines

cms/grading/tasktypes/Communication.py
```diff
@@ -179,7 +179,7 @@ def evaluate(self, job, file_cacher):
             num_processes = 1
         else:
             num_processes = self.parameters[0]
-        indices = range(num_processes)
+        indices = list(range(num_processes))
```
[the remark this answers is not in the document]
lerks: I don't think so, it is iterated upon several times in the following, if it were an iterator it would be "exhausted" after the first.

cms/grading/tasktypes/Communication.py
lerks: I usually don't put any.

cmstaskenv/Test.py
```diff
@@ -21,8 +21,11 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 from __future__ import absolute_import
+from __future__ import division
```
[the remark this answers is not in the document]
lerks: Changed the `%4d` to a `%4.3g` for consistency.

cms/server/admin/handlers/dataset.py
lerks: Done (and removed the one before, as explained in the other comment)

cmscontrib/DumpImporter.py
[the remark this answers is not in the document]
lerks: Done.

cmstaskenv/cmsMake.py
```diff
@@ -453,7 +456,7 @@ def make_input(assume=None):
                 command.append("%s" % st)
             call(base_dir, command)
             n += 1
-            for i in xrange(3):
+            for i in range(3):
```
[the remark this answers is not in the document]
lerks: Done.

setup.py
```diff
@@ -33,6 +33,8 @@
 from __future__ import print_function
```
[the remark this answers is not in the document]
lerks: Done, and yes: adding unicode_literals gives `error in cms setup command: package_data must be a dictionary mapping package names to lists of wildcard patterns`.

cmstestsuite/unit_tests/db/filecacher_test.py
lerks: Done.

cms/server/util.py
```diff
@@ -493,7 +493,7 @@ def format_token_rules(tokens, t_type=None, locale=None):
         tokens["type_pl"] = _("tokens")
 
     tokens["min_interval"] = tokens["min_interval"].total_seconds()
-    tokens["gen_interval"] = tokens["gen_interval"].total_seconds() / 60
+    tokens["gen_interval"] = tokens["gen_interval"].total_seconds() // 60
```
stefano-maggiolo: You sure about this? total_seconds() should be a float, so before it was float division - I think.

cmscommon/datetime.py
```diff
@@ -227,7 +227,7 @@ def monotonic_time():
         from win32api import GetTickCount
 
         def monotonic_time():
-            return GetTickCount / 1000.0
+            return GetTickCount / 1000
```
stefano-maggiolo: Apart from the lack of usefulness of this branch, I'd say don't touch code borrowed from somewhere else.

cms/grading/tasktypes/Communication.py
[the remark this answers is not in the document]
stefano-maggiolo: Uhm...

```In [15]: a = xrange(3)

In [16]: print [x for x in a]
[0, 1, 2]

In [17]: print [x for x in a]
[0, 1, 2]

In [18]: print [x for x in a]
[0, 1, 2]

In [19]: print [x for x in a]
[0, 1, 2]
```

Works also in py3k with range - but I guess it's ok if you want to keep it, since is supersmall

stefano-maggiolo:
Also, I don't like most of the commit removing .0 - in almost all occasion is a nice way to hint to the developer the type of the constant, and I find the previous version way clearer.

cms/grading/tasktypes/Communication.py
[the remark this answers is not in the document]
lerks: Wow, didn't think that would work, I will fix this, thanks!

cms/server/util.py
lerks: It was float division, but the only uses of the result are comparing it to 1.0 and then formatting it with `%d`: I think it's better if the division returns an int.

cmscommon/datetime.py
lerks: Undone.

lerks:
Reverted. PTAL.

cms/server/util.py
stefano-maggiolo: I think I see a print with %g, and no comparison. You looking at the right one?

stefano-maggiolo:
After you settle the last comment, ok to push (to the temporary branch)

cms/server/util.py
lerks: Oh, yes, you're right, I must have misread, I changed this back to a float division.

lerks:
I pushed all of this to the `cms-dev/py3` branch. I will soon open a PR with the second batch of commits.

PR #822, opened 2018-01-28
Couple minor things in cmstestsuite (in particular for RunTimeTest

stefano-maggiolo:
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/822)

### DIFF
--- a/cmstestsuite/testrunner.py
+++ b/cmstestsuite/testrunner.py
@@ -47,6 +47,7 @@ logger = logging.getLogger(__name__)
 class TestRunner(object):
     def __init__(self, test_list, contest_id=None, workers=1):
         self.start_time = datetime.datetime.now()
+        self.last_end_time = self.start_time
 
         self.ps = ProgramStarter()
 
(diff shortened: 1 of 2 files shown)


cmstestsuite/testrunner.py
```diff
@@ -100,13 +101,24 @@ def load_cms_conf():
         return get_cms_config()
 
     def log_elapsed_time(self):
+        def duration_to_human(start, end):
+            millis = int((end - start).total_seconds() * 1000)
```
lerks: Tip: you could use `str(end - start)` which returns a similar format (though not exactly identical) to save yourself some effort.

cmstestsuite/testrunner.py
```diff
@@ -278,20 +278,29 @@ def _all_user_tests(self):
...
-    def submit_tests(self):
+    def submit_tests(self, concurrent_submit_and_eval=True):
         """Create the tasks, and submit for all languages in all tests.
 
+        concurrent_submit_and_eval (boolean): if False, stop ES while
+            we are submitting, so that we have a clearer view of the time
+            each step takes.
+
         """
+        def start_eval_services():
```
lerks: There's a discrepancy between what this function does and what the docstring says it does.

cmstestsuite/testrunner.py
```diff
@@ -278,20 +278,29 @@ def _all_user_tests(self):
...
+            self.ps.start("EvaluationService", contest=self.contest_id)
```
lerks: Delaying just ES is enough to achieve what you wanted. You could start PS (and possibly the workers) together with CWS at the beginning. Or not, doesn't matter much. As you prefer.

cmstestsuite/testrunner.py
stefano-maggiolo: Ah, I was looking for something like that -,-

cmstestsuite/testrunner.py
stefano-maggiolo: Done

cmstestsuite/testrunner.py
stefano-maggiolo: Done

lerks:
LGTM

stefano-maggiolo:
Merged

PR #823, opened 2018-01-29
Use Python's Abstract Base Classes for our abstract classes

lerks:
Rather than raising NotImplemented in abstract methods decorate them
with @abstractmethod. Abstract classes, which are those that have at
least one @abstractmethod, cannot be instantiated. This is more robust.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/823)

### DIFF
--- a/cms/grading/Sandbox.py
+++ b/cms/grading/Sandbox.py
@@ -32,6 +32,7 @@ import resource
 import select
 import stat
 import tempfile
+from abc import ABCMeta, abstractmethod
 from functools import wraps, partial
 
 import gevent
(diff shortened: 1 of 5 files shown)


cms/grading/Sandbox.py
```diff
@@ -186,6 +187,8 @@ class SandboxBase(object):
 
     """
 
+    __metaclass__ = ABCMeta
```
stefano-maggiolo: This looks like neither of the two recommended ways of doing it (reading https://docs.python.org/3/library/abc.html) - is there a reason not to follow documentation?

lerks: There are good reasons not to follow the Py3 doc and follow the Py2 instead: https://docs.python.org/2.7/library/abc.html

:P

stefano-maggiolo:
I see! In that case, push on!

PR #824, opened 2018-02-01
Move codebase to Py2+3 [2]

lerks:
This is the second in a series of pull requests that aim at making our codebase compatible with Python 2 and Python 3 at the same time.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/824)

### DIFF
--- a/cms/db/submission.py
+++ b/cms/db/submission.py
@@ -441,8 +441,12 @@ class SubmissionResult(Base):
         t, m = None, None
         if self.evaluated() and self.evaluations:
             for ev in self.evaluations:
-                t = max(t, ev.execution_time)
-                m = max(m, ev.execution_memory)
+                if ev.execution_time is not None \
+                        and (t is None or t < ev.execution_time):
+                    t = ev.execution_time
+                if ev.execution_memory is not None \
+                        and (m is None or m < ev.execution_memory):
+                    m = ev.execution_memory
         return (t, m)
 
     def compiled(self):
--- a/cms/grading/tasktypes/Batch.py
+++ b/cms/grading/tasktypes/Batch.py
@@ -27,6 +27,7 @@ from __future__ import print_function
 from __future__ import unicode_literals
 from future.builtins.disabled import *
 from future.builtins import *
+from six import iterkeys, iteritems
 
 import logging
 
@@ -163,7 +164,7 @@ class Batch(TaskType):
 
         # Prepare the source files in the sandbox
         files_to_get = {}
-        format_filename = job.files.keys()[0]
+        format_filename, = tuple(iterkeys(job.files))
         source_filenames = []
         source_filenames.append(format_filename.replace(".%l", source_ext))
         files_to_get[source_filenames[0]] = \
@@ -177,7 +178,7 @@ class Batch(TaskType):
                 job.managers["grader%s" % source_ext].digest
 
         # Also copy all managers that might be useful during compilation.
-        for filename in job.managers.iterkeys():
+        for filename in iterkeys(job.managers):
             if any(filename.endswith(header) for header in HEADER_EXTS):
                 files_to_get[filename] = \
                     job.managers[filename].digest
@@ -188,7 +189,7 @@ class Batch(TaskType):
                 files_to_get[filename] = \
                     job.managers[filename].digest
 
-        for filename, digest in files_to_get.iteritems():
+        for filename, digest in iteritems(files_to_get):
             sandbox.create_file_from_storage(filename, digest)
 
         # Prepare the compilation command
@@ -222,7 +223,7 @@ class Batch(TaskType):
         sandbox = create_sandbox(file_cacher, job.multithreaded_sandbox)
 
         # Prepare the execution
-        executable_filename = job.executables.keys()[0]
+        executable_filename, = tuple(iterkeys(job.executables))
         language = get_language(job.language)
         commands = language.get_evaluation_commands(
             executable_filename,
@@ -248,9 +249,9 @@ class Batch(TaskType):
             }
 
         # Put the required files into the sandbox
-        for filename, digest in executables_to_get.iteritems():
+        for filename, digest in iteritems(executables_to_get):
             sandbox.create_file_from_storage(filename, digest, executable=True)
-        for filename, digest in files_to_get.iteritems():
+        for filename, digest in iteritems(files_to_get):
             sandbox.create_file_from_storage(filename, digest)
 
         # Actually performs the execution
--- a/cms/grading/tasktypes/Batch.py
+++ b/cms/grading/tasktypes/Batch.py
@@ -164,7 +164,7 @@ class Batch(TaskType):
 
         # Prepare the source files in the sandbox
         files_to_get = {}
-        format_filename, = tuple(iterkeys(job.files))
+        format_filename = next(iterkeys(job.files))
         source_filenames = []
         source_filenames.append(format_filename.replace(".%l", source_ext))
         files_to_get[source_filenames[0]] = \
@@ -223,7 +223,8 @@ class Batch(TaskType):
         sandbox = create_sandbox(file_cacher, job.multithreaded_sandbox)
 
         # Prepare the execution
-        executable_filename, = tuple(iterkeys(job.executables))
+        assert len(job.executables) == 1
+        executable_filename = next(iterkeys(job.executables))
         language = get_language(job.language)
         commands = language.get_evaluation_commands(
             executable_filename,
--- a/cms/grading/tasktypes/Communication.py
+++ b/cms/grading/tasktypes/Communication.py
@@ -27,6 +27,7 @@ from __future__ import print_function
 from __future__ import unicode_literals
 from future.builtins.disabled import *
 from future.builtins import *
+from six import iterkeys, iteritems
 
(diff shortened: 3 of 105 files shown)


cms/grading/tasktypes/Batch.py
```diff
@@ -163,7 +164,7 @@ def compile(self, job, file_cacher):
 
         # Prepare the source files in the sandbox
         files_to_get = {}
-        format_filename = job.files.keys()[0]
+        format_filename, = tuple(iterkeys(job.files))
```
stefano-maggiolo: `next(iterkeys(job.files))` looks more like what you want here.

cms/grading/tasktypes/Batch.py
```diff
@@ -222,7 +223,7 @@ def evaluate(self, job, file_cacher):
         sandbox = create_sandbox(file_cacher, job.multithreaded_sandbox)
 
         # Prepare the execution
-        executable_filename = job.executables.keys()[0]
+        executable_filename, = tuple(iterkeys(job.executables))
```
stefano-maggiolo: Same

cms/grading/tasktypes/Communication.py
```diff
@@ -209,22 +210,22 @@ def evaluate(self, job, file_cacher):
...
                 filename, digest, executable=True)
-        for filename, digest in manager_files_to_get.iteritems():
+        for filename, digest in iteritems(manager_files_to_get):
             sandbox_mgr.create_file_from_storage(filename, digest)
 
         # Second step: load the executables for the user processes
         # (done before launching the manager so that it does not
         # impact its wall clock time).
-        executable_filename = job.executables.keys()[0]
+        executable_filename, = tuple(iterkeys(job.executables))
```
stefano-maggiolo: Same.

cms/server/contest/handlers/tasksubmission.py
```diff
@@ -184,7 +185,7 @@ def post(self, task_name):
         # not for submissions requiring a programming language
         # identification).
         if len(self.request.files) == 1 and \
-                self.request.files.keys()[0] == "submission":
+                tuple(iterkeys(self.request.files)) == ("submission",):
```
stefano-maggiolo: Same.

cms/server/contest/handlers/taskusertest.py
```diff
@@ -253,7 +254,7 @@ def post(self, task_name):
         # not for submissions requiring a programming language
         # identification).
         if len(self.request.files) == 1 and \
-                self.request.files.keys()[0] == "submission":
+                tuple(iterkeys(self.request.files)) == ("submission",):
```
stefano-maggiolo: Same

cms/server/contest/templates/submission_row.html
```diff
@@ -97,7 +97,7 @@
             {{ _("None") }}
         </a>
         {% elif len(s.files) == 1 %}
-            {% set filename = s.files.keys()[0] %}
+            {% set filename, = tuple(iterkeys(s.files)) %}
```
stefano-maggiolo: Same

cms/server/contest/templates/user_test_row.html
```diff
@@ -76,13 +77,13 @@
...
         {% comment the common case. Yet, it still allows the marginal case of both 'foo.%l' and 'foo.c' in the submission format. %}
-        {% set files = t.files.keys() + t.managers.keys() %}
+        {% set files = list(itertools.chain(iterkeys(t.files), iterkeys(t.managers))) %}
         {% if len(files) == 0 %}
         <a class="btn disabled">
             {{ _("None") }}
         </a>
         {% elif len(files) == 1 %}
-            {% set filename = t.files.keys()[0] %}
+            {% set filename, = tuple(iterkeys(t.files)) %}
```
stefano-maggiolo: Same

cms/server/contest/templates/user_test_row.html
```diff
@@ -76,13 +77,13 @@
         {% comment We replace '.%l' with the actual language only when it occurs as an extension at the end of the string and only when %}
         {% comment there isn't another file with that name. This allows to securily reverse the replacement and should work great in %}
         {% comment the common case. Yet, it still allows the marginal case of both 'foo.%l' and 'foo.c' in the submission format. %}
-        {% set files = t.files.keys() + t.managers.keys() %}
+        {% set files = list(itertools.chain(iterkeys(t.files), iterkeys(t.managers))) %}
```
stefano-maggiolo: Why not list(a) + list(b) and one import less?

cmscontrib/__init__.py
```diff
@@ -163,7 +165,8 @@ def _update_object(self, old_object, new_object, ignore=None):
 
             # General case #1: a dict
             elif isinstance(old_value, dict):
-                for key in set(old_value.keys()) | set(new_value.keys()):
+                for key in set(itertools.chain(iterkeys(old_value),
+                                               iterkeys(new_value))):
```
stefano-maggiolo: Same here, but here it might be a bit faster to do it like you did (if there are repeated elements).

cmstestsuite/unit_tests/service/esoperationstest.py
```diff
@@ -191,7 +192,8 @@ def test_get_submissions_operations_partially_evaluate(self):
             self.tasks[0], self.participation, True)
         evaluated_codenames = set()
         for result in results:
-            evaluated_codename = result.dataset.testcases.keys()[0]
+            # Pick one arbitrary testcase.
+            evaluated_codename = set(iterkeys(result.dataset.testcases)).pop()
```
stefano-maggiolo: If you like it more, next(iter) applies here too.

cmstestsuite/unit_tests/service/scoringoperationstest.py
```diff
@@ -102,7 +103,8 @@ def test_get_operations_no_operations(self):
             self.tasks[0], self.participation, True)
         evaluated_codenames = set()
         for result in results:
-            evaluated_codename = result.dataset.testcases.keys()[0]
+            # Pick one arbitrary testcase.
+            evaluated_codename = set(iterkeys(result.dataset.testcases)).pop()
```
stefano-maggiolo: And here.

cmstestsuite/web/__init__.py
```diff
@@ -212,16 +213,15 @@ def specific_info(self):
         res = "URL: %s\n" % self.url
         if self.response is not None:
             res += "\nREQUEST HEADERS\n"
-            for (key, value) in self.response.request.headers.iteritems():
+            for key, value in iteritems(self.response.request.headers()):
```
stefano-maggiolo: Are the new `()` intended?

cms/db/submission.py
```diff
@@ -441,8 +441,12 @@ def get_max_evaluation_resources(self):
...
         if self.evaluated() and self.evaluations:
             for ev in self.evaluations:
-                t = max(t, ev.execution_time)
-                m = max(m, ev.execution_memory)
+                if ev.execution_time is not None \
+                        and (t is None or t < ev.execution_time):
+                    t = ev.execution_time
+                if ev.execution_memory is not None \
+                        and (m is None or m < ev.execution_memory):
+                    m = ev.execution_memory
```
stefano-maggiolo: Optional: I like the condition more if phrased like this:
```python
if t is None or (ev.execution_time is not None and t < ev.execution_time):
```

cms/server/contest/templates/user_test_row.html
lerks: Because I expected this to be more efficient in terms of memory and hence of running time. Yet I just did some tests and it turns out that they are essentially equal (actually, doing `list(chain)` is slower by a few percentage points). I'm reverting to `list + list`.

cmscontrib/__init__.py
lerks: I ran some tests and apparently the running time isn't influenced by how many keys the two dicts have in common. Yet, doing it with `set | set` is about 50% slower than doing it `set(chain)`. We're talking about ridiculously small values (2 and 3 microseconds for two dicts of 26 elements each) so I don't think performance is really relevant. For consistency with the above comment I'm going to revert to `set | set`.

cms/db/submission.py
lerks: One reason I prefer the condition I proposed to the one you propose is that yours leads to overwrite `None` with `None`.

cmstestsuite/web/__init__.py
lerks: Nope, this was an oversight, good catch!

lerks:
One one hand I like the `a, = ...` syntax more, mainly because it implicitly asserts that there's exactly one element in the iterator. (Moreover, I just figured out that the `tuple` part can be left out, making it just `a, = iterkeys(d)`, which is even shorter). But I agree that this is an uncommon syntax which may confuse some readers, and that in most cases the singleton-ness of the iterator is already checked elsewhere. I'm therefore doing as you suggested, using `a = next(iterkeys(d))` and adding an explicit assert where there wasn't any other check.

PTAL.

stefano-maggiolo:
LGTM, next!

lerks:
Pushed to the cms-dev/py3 branch.

PR #825, opened 2018-02-03
Move codebase to Py2+3 [3]

lerks:
This is the third in a series of pull requests that aim at making our codebase compatible with Python 2 and Python 3 at the same time.

### DIFF
--- a/cms/locale/locale.py
+++ b/cms/locale/locale.py
@@ -36,6 +36,7 @@ from __future__ import unicode_literals
 from future.builtins.disabled import *
 from future.builtins import *
 from six import iteritems
+import six
 
 import pkg_resources
 import gettext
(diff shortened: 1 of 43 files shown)


cms/locale/locale.py
```diff
@@ -127,9 +128,15 @@ def wrap_translations_for_tornado(trans):
...
+            if six.PY3:
+                return trans.ngettext(message, plural_message, count)
+            else:
+                return trans.ungettext(message, plural_message, count)
         else:
-            return trans.ugettext(message)
+            if six.PY3:
+                return trans.gettext(message)
+            else:
+                return trans.ugettext(message)
```
stefano-maggiolo: Optional: mild preference for assigning only once at the top `_gettext` and `_ngettext` to the correct variant

PR #827, opened 2018-02-04
Move codebase to Py2+3 [5]

lerks:
This is the fifth in a series of pull requests that aim at making our codebase compatible with Python 2 and Python 3 at the same time.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/827)

### DIFF
--- a/cmstestsuite/testrunner.py
+++ b/cmstestsuite/testrunner.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python
 # -*- coding: utf-8 -*-
 
 # Contest Management System - http://cms-dev.github.io/
--- a/cmstestsuite/testrunner.py
+++ b/cmstestsuite/testrunner.py
@@ -304,6 +304,15 @@ class TestRunner(object):
             self.ps.start("Worker", shard)
         if concurrent_submit_and_eval:
             self.ps.start("EvaluationService", contest=self.contest_id)
+            # We need to wait on ES as otherwise CWS would receive
+            # submissions before it has established a connection with
+            # ES, and this would cause ES not to receive notifications
+            # about new submissions, which would only get picked up at
+            # the following sweep, delaying judging. Moreover a race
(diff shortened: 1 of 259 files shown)


cmstestsuite/testrunner.py
```diff
@@ -304,6 +304,15 @@ def submit_tests(self, concurrent_submit_and_eval=True):
...
         if concurrent_submit_and_eval:
             self.ps.start("EvaluationService", contest=self.contest_id)
+            # We need to wait on ES as otherwise CWS would receive
+            # submissions before it has established a connection with
+            # ES, and this would cause ES not to receive notifications
+            # about new submissions, which would only get picked up at
+            # the following sweep, delaying judging. Moreover a race
+            # condition causes the first sweep of ES not to pick up
+            # some of the submissions that are being committed by CWS
+            # to the DB but that aren't being notified via RPC.
```
stefano-maggiolo: Can I rephrase?

```
A submission can arrive after ES' first sweep, but before CWS connects
to ES; if so, it will be ignored until the second sweep, making the test
flaky. By waiting for ES to start before submitting, we ensure CWS can
send the notification for all submissions.
```

cmstestsuite/testrunner.py
```diff
@@ -304,6 +304,15 @@ def submit_tests(self, concurrent_submit_and_eval=True):
             self.ps.start("Worker", shard)
         if concurrent_submit_and_eval:
             self.ps.start("EvaluationService", contest=self.contest_id)
```
stefano-maggiolo: I think there is a better arrangement:

<start CWS>
<if concurrent, start ES>
<wait>
<start PS and Workers>
<submit>
<if not concurrent start ES>
<wait>

As you wrote it, if concurrent we wait for workers and PS without the need to, and if not we never wait for PS and workers, which means instead of outputting "cannot start PS" we will output some other unclear error long after the real problem.

lerks:
Done.

PR #828, opened 2018-02-04
Fixed performance issue reevaluating a dataset

### LINKED ISSUE #662: Per-task submission invalidation very slow with a lot of submissions
Hi,
Mass invalidation of submissions on a per-task basis is very slow (takes minutes) on a database with ~100k submissions. 

Moreover, it also causes some queries on the database that really don't make any sense to me:

```
2016-10-09 00:09:06 CEST allenamenti LOG:  duration: 936.352 ms  statement: SELECT submissions.id AS submissions_id, datasets.id AS datasets_id, CASE WHEN (datasets.id != tasks.active_dataset_id) THEN 4 ELSE 1 END AS anon_1, submissions.timestamp AS submissions_timestamp 
    FROM submissions JOIN tasks ON tasks.id = submissions.task_id JOIN datasets ON tasks.id = datasets.task_id LEFT OUTER JOIN submission_results ON datasets.id = submission_results.dataset_id AND submissions.id = submission_results.submission_id 
    WHERE true AND (datasets.id = tasks.active_dataset_id OR datasets.autojudge IS true) AND submission_results.dataset_id IS NULL
```

or queries that just read the whole submission table into python:

```
2016-10-09 00:10:09 CEST allenamenti LOG:  duration: 370.948 ms  statement: SELECT submissions.id AS submissions_id, submissions.participation_id AS submissions_participation_id, submissions.task_id AS submissions_task_id, submissions.timestamp AS submissions_timestamp, submissions.language AS submissions_language, submissions.comment AS submissions_comment 
    FROM submissions
```

(actually, I'm not 100% sure these queries are all caused by a submission invalidation request, but the times match - moreover, these queries (imho) should be considered a problem anyway).

edomora97:
When reevaluating a dataset all the evaluations of the contest are pulled by `get_submissions` and then most of them are discarded. This is caused because `get_submissions` knows anything about datasets and the `task_id` is not passed.

When a request of reevaluation of a dataset is made, the task_id is pulled from the db and passed to `get_evaluation` to filter only the _interesting_ submissions.

Since #662 the submissions in the db have grown to over 230k and ES freezes when a reevaluation is requested even for a dataset with few submissions.

Fixes #662

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/828)

### DIFF
--- a/cms/service/EvaluationService.py
+++ b/cms/service/EvaluationService.py
@@ -888,6 +888,10 @@ class EvaluationService(TriggeredService):
             contest_id = self.contest_id
 
         with SessionGen() as session:
+            # When invalidating a dataset we need to know the task_id, otherwise
+            # get_submissions will return all the submissions of the contest.
+            if dataset_id is not None and task_id is None:
+                task_id = Dataset.get_from_id(dataset_id, session).task_id
             # First we load all involved submissions.
             submissions = get_submissions(
                 # Give contest_id only if all others are None.
@@ -921,7 +925,11 @@ class EvaluationService(TriggeredService):
                     submission_id,
                     dataset_id} == {None}
                 else None,
-                participation_id, task_id, submission_id, dataset_id, session)
+                participation_id,
+                # Provide the task_id only if the entire task has to be
+                # reevaluated and not only a specific dataset.
+                task_id if dataset_id is None else None,
+                submission_id, dataset_id, session)
             logger.info("Submission results to invalidate %s for: %d.",
                         level, len(submission_results))
             for submission_result in submission_results:


cms/service/EvaluationService.py
```diff
@@ -921,7 +925,11 @@ def invalidate_submission(self,
                     submission_id,
                     dataset_id} == {None}
                 else None,
-                participation_id, task_id, submission_id, dataset_id, session)
+                participation_id,
+                # Provide the task_id only if the entire task has to be
+                # reevaluated and not only a specific dataset.
```
wil93: just out of curiosity, is it even possible to reevaluate a task and not just a dataset?

edomora97: AFAIK AWS has no such option, but the RPC API allows it.

stefano-maggiolo:
OMG this method is so bad.
 Thank you Edoardo!

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/828#-:-L4_I-PDIOdPx5KsrZlj:b1esysy)*

stefano-maggiolo:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/828#-:-L4_I7vFsAYAdvwxD2kV:bnfp4nl)*

stefano-maggiolo:
Merged

PR #830, opened 2018-02-04
Add tests for some grading/ functions, and better docs for white-diff

stefano-maggiolo:
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/830)

### DIFF
--- a/cmstestsuite/unit_tests/grading/__init__.py
+++ b/cmstestsuite/unit_tests/grading/__init__.py
@@ -1,0 +1,204 @@
+#!/usr/bin/env python2
+# -*- coding: utf-8 -*-
+
+# Contest Management System - http://cms-dev.github.io/
+# Copyright © 2018 Stefano Maggiolo <s.maggiolo@gmail.com>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Tests for common grading functions.
+
+"""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+from future.builtins.disabled import *
+from future.builtins import *
+
+import unittest
+from io import BytesIO
+
+from cms.grading import Sandbox, WHITES, \
+    format_status_text, merge_evaluation_results, white_diff
+
+
+class TestFormatStatusText(unittest.TestCase):
+
+    @staticmethod
+    def _tr(s):
+        return s.replace("A", "E")
+
+    def test_success_no_placeholders(self):
+        self.assertEqual(format_status_text([]), "N/A")
+        self.assertEqual(format_status_text([""]), "")
+        self.assertEqual(format_status_text(["ASD"]), "ASD")
+        self.assertEqual(format_status_text(["你好"]), "你好")
+
+    def test_success_with_placeholders(self):
+        self.assertEqual(format_status_text(["%s", "ASD"]), "ASD")
+        self.assertEqual(format_status_text(["ASD%s\n%s", "QWE", "123"]),
+                         "ASDQWE\n123")
+
+    def test_success_with_translator(self):
+        self.assertEqual(format_status_text([""], self._tr), "")
+        self.assertEqual(format_status_text(["ASD"], self._tr), "ESD")
+        # Translation is applied before formatting.
+        self.assertEqual(format_status_text(["A%s", "ASD"], self._tr), "EASD")
+        self.assertEqual(
+            format_status_text(["AAA %s\n%s", "AAA", "AE"], self._tr),
+            "EEE AAA\nAE")
+
+    def test_insuccess(self):
+        # Not enough elements for the placeholders.
+        self.assertEqual(format_status_text(["%s"]), "N/A")
+        self.assertEqual(format_status_text(["%s"], self._tr), "N/E")
+        # No elements at all.
+        self.assertEqual(format_status_text([]), "N/A")
+        self.assertEqual(format_status_text([], self._tr), "N/E")
+
+
+class TestMergeEvaluationResults(unittest.TestCase):
+
+    @staticmethod
+    def _res(execution_time, execution_wall_clock_time, execution_memory,
+             exit_status, signal=None, filename=None):
+        r = {
+            "execution_time": execution_time,
+            "execution_wall_clock_time": execution_wall_clock_time,
+            "execution_memory": execution_memory,
+            "exit_status": exit_status,
+        }
+        if signal is not None:
+            r["signal"] = signal
+        if filename is not None:
+            r["filename"] = filename
+        return r
+
+    def assertRes(self, r0, r1):
+        """Assert that r0 and r1 are the same result."""
+        self.assertAlmostEqual(r0["execution_time"], r1["execution_time"])
+        self.assertAlmostEqual(r0["execution_wall_clock_time"],
+                               r1["execution_wall_clock_time"])
+        self.assertAlmostEqual(r0["execution_memory"], r1["execution_memory"])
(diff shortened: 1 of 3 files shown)


cmstestsuite/unit_tests/grading/__init__.py
```diff
@@ -0,0 +1,174 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Tests for common grading functions.
+
+"""
+
+from __future__ import absolute_import
+from __future__ import print_function
+from __future__ import unicode_literals
+
```
lerks: Add:
```
from __future__ import division
```
and
```
from future.builtins.disabled import *
from future.builtins import *
```

cmstestsuite/unit_tests/grading/__init__.py
```diff
@@ -0,0 +1,174 @@
...
+import StringIO
```
lerks: Use the one from the `io` module.

cmstestsuite/unit_tests/grading/__init__.py
```diff
@@ -0,0 +1,174 @@
...
+        self.assertEqual(format_status_text(["你好"]), "你好")
+
+    def test_success_with_placeholders(self):
+        self.assertEqual(format_status_text(["%s", "ASD"]), "ASD")
+        self.assertEqual(format_status_text(["ASD%s\n%s", "QWE", "123"]),
+                         "ASDQWE\n123")
+
+    def test_success_with_translator(self):
+        # Translation is applied before formatting.
+        self.assertEqual(format_status_text(["A%s", "ASD"], self._tr), "EASD")
```
lerks: Maybe also translator w/o placeholder?

cmstestsuite/unit_tests/grading/__init__.py
```diff
@@ -0,0 +1,174 @@
...
+        self.assertEqual(
+            format_status_text(["AAA %s\n%s", "AAA", "AE"], self._tr),
+            "EEE AAA\nAE")
+
+    def test_insuccess(self):
+        self.assertEqual(format_status_text(["%s"]), "N/A")
+        self.assertEqual(format_status_text(["%s"], self._tr), "N/E")
```
lerks: Maybe empty list too?

cmstestsuite/unit_tests/grading/__init__.py
stefano-maggiolo: Done.

cmstestsuite/unit_tests/grading/__init__.py
```diff
@@ -0,0 +1,174 @@
...
+            "exit_status": exit_status,
+        }
+        if signal is not None:
+            r["signal"] = signal
+        if filename is not None:
+            r["filename"] = filename
+        return r
+
+    def test_success_status_ok(self):
+        self.assertEqual(
```
lerks: Isn't it bad practice to compare floats for exact equality? Looking at the doc I see that starting with Python 3.1 special type-specific comparators are used. So maybe here we use a dict comparator, that then in turn uses a float approximate comparator? Could you check?

cmstestsuite/unit_tests/grading/__init__.py
```diff
@@ -0,0 +1,174 @@
...
+            merge_evaluation_results(
+                self._res(0, 0, 0, Sandbox.EXIT_FILE_ACCESS, filename="asd"),
+                self._res(0, 0, 0, Sandbox.EXIT_SIGNAL, signal="11")),
+            self._res(0, 0, 0, Sandbox.EXIT_FILE_ACCESS, filename="asd"))
+        self.assertEqual(
+            merge_evaluation_results(
+                self._res(0, 0, 0, Sandbox.EXIT_SIGNAL, signal="9"),
+                self._res(0, 0, 0, Sandbox.EXIT_SIGNAL, signal="11")),
+            self._res(0, 0, 0, Sandbox.EXIT_SIGNAL, signal="9"))
```
lerks: Maybe add one where the second status is ok?

cmstestsuite/unit_tests/grading/__init__.py
```diff
@@ -0,0 +1,174 @@
...
+            r0, self._res(1.0, 2.0, 300, Sandbox.EXIT_OK))
+        self.assertEqual(
+            r1, self._res(0.1, 0.2, 0.3, Sandbox.EXIT_SIGNAL, signal="11"))
+
+
+class TestWhiteDiff(unittest.TestCase):
+
+    @staticmethod
+    def _diff(s1, s2):
+        return white_diff(StringIO.StringIO(s1), StringIO.StringIO(s2))
```
lerks: White diff is designed to operate on bytes, not strings.

lerks: Which also means that all string literals in this test should become `b"..."`. (I guess right now an implicit encoding using utf-8 is happening somewhere)

lerks: About that, maybe add some tests that use non-ascii-encoded strings...

docs/Task types.rst
```diff
@@ -92,3 +92,20 @@ The resulting executable is run twice (one acting as the computer, one acting as
...
 
 Normally, the standard output of the second invocation of the manager is compared to a provided reference output file using the white-diff comparator. However, the admins may provide a :file:`checker` executable, with the same properties as for Batch. If a file with such a name is found in the uploaded manager files, it will be run instead of the white-diff comparator.
+
+
+.. _tasktypes_white_diff:
+
+White-diff comparator
+=====================
+
+White-diff is the only built-in comparator, and is used in most task types. It can be used when each testcase has a single correct output files, up to whitespaces.
```
lerks: s/single/unique/?

and s/files/file/

cmstestsuite/unit_tests/grading/__init__.py
stefano-maggiolo: Done

cmstestsuite/unit_tests/grading/__init__.py
stefano-maggiolo: Done.

cmstestsuite/unit_tests/grading/__init__.py
stefano-maggiolo: Done. The dict-specific is the same as assertEqual but in addition tests they are dicts, so no special float comparison.

cmstestsuite/unit_tests/grading/__init__.py
stefano-maggiolo: Done.

cmstestsuite/unit_tests/grading/__init__.py
stefano-maggiolo: Done.

docs/Task types.rst
stefano-maggiolo: Done

stefano-maggiolo:
PTAL

cmstestsuite/unit_tests/grading/__init__.py
```diff
@@ -22,10 +22,16 @@
...
 from __future__ import print_function
 from __future__ import unicode_literals
+from future.builtins.disabled import *
+from future.builtins import *
 
-import StringIO
+try:
+    from BytesIO import BytesIO
+except ImportError:
+    from io import BytesIO
```
lerks: No need: `io.BytesIO` exists in both py2 and py3.

cmstestsuite/unit_tests/grading/__init__.py
```diff
@@ -77,74 +90,94 @@ def _res(execution_time, execution_wall_clock_time, execution_memory,
             r["filename"] = filename
         return r
 
+    def assertRes(self, r0, r1):
+        """Assert that r0 and r1 are the same result."""
+        self.assertAlmostEqual(r0["execution_time"], r1["execution_time"])
```
lerks: Would you mind adding something equivalent to `assert {"execution_time", "execution_wall_clock_time", "execution_memory", "syscall", "signal", "filename"}.issubset(iterkeys(r0))` and similarly for `r1`?

cmstestsuite/unit_tests/grading/__init__.py
stefano-maggiolo: Done.

cmstestsuite/unit_tests/grading/__init__.py
stefano-maggiolo: Apart that it'd be wrong (the last three are usually not present), I disagree. I might add set(iterkeys(r0)) == set(iterkeys(r1)), if that's your goal?

stefano-maggiolo: Agreed offline not to add anything.

PR #831, opened 2018-02-05
Notify ES if it is active while adding submissions from command line

stefano-maggiolo:
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/831)

### DIFF
--- a/cmscontrib/AddSubmission.py
+++ b/cmscontrib/AddSubmission.py
@@ -2,7 +2,7 @@
 # -*- coding: utf-8 -*-
 
 # Contest Management System - http://cms-dev.github.io/
-# Copyright © 2015-2017 Stefano Maggiolo <s.maggiolo@gmail.com>
+# Copyright © 2015-2018 Stefano Maggiolo <s.maggiolo@gmail.com>
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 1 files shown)


cmscontrib/AddSubmission.py
```diff
@@ -33,17 +33,26 @@
...
 
 logger = logging.getLogger(__name__)
 
 
+def maybe_send_notification(submission_id):
+    """Non-blocking attempt to notify a running ES of the submission"""
+    rs = RemoteServiceClient(ServiceCoord("EvaluationService", 0))
+    rs.connect()
+    rs.new_submission(submission_id=submission_id)
+    rs.finalize("Closing script")
```
lerks: You may want to call `disconnect()` instead, as `finalize` is more lower-level...

stefano-maggiolo:
PTAL

stefano-maggiolo:

---

*[cmscontrib/AddSubmission.py, line 53 at r1](https://reviewable.io:443/reviews/cms-dev/cms/831#-L4_BX8gY_mIgqcLtg09-r1-53:-L4adFfaafDLKkSIJGXh:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/67e971caef266fcd59135894589c1011d6414b80/cmscontrib/AddSubmission.py#L53)):*

You may want to call `disconnect()` instead, as `finalize` is more lower-level...
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/831)*

PR #841, opened 2018-02-18
Use Babel's built-in support for language and country ISO codes

lerks:
Rather than parsing the files of the iso-codes package ourselves.

A downside is that we lose the formatting of country codes by
themselves, but my feeling is that their usage was limited.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/841)


stefano-maggiolo:
Looks fine, but WDYM with

>  formatting of country codes by themselves

---

Review status: 0 of 7 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/841#-:-L5c6ZTj0waG7GaDCqt3:b-t3mrqy)*

lerks:
I mean that if `statement.language` is just `US`, before we showed `United States of America`, now we show just `US`. One needs to put `en_US` to get a meaningful output, i.e., `American English`.

---

Review status: 0 of 7 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/cms-dev/cms/841#-:-L5c7DJD5sb-XedUzFBR:b-9pik7e)*

stefano-maggiolo:
I see. This is only for when we display the language of statements, not for
the list of translations, right? If so, ok, otherwise we might want to move
all translations to specify the country too.

lerks:
I probably wasn't clear. The way available translations are displayed in the drop-down menu and the way a statement's language is displayed are the same, and are as follows:
* for lone language codes the behavior is unchanged: `en` was and will be rendered as `English`;
* for joint language and country codes the behavior is slightly changed: `en_US` was rendered as `English (United States of America)` and will be rendered as `American English`;
* for lone country codes the behavior changed: `US` was rendered as `United States of America` and now will be rendered unformatted just as `US` (`IT` will be rendered as `IT`, `FR` as `FR`, etc.);
* anything else was and will be rendered unformatted, i.e., `foobar` as `foobar`.

stefano-maggiolo:
Ah ok, sounds good, lgtm

PR #842, opened 2018-02-18
Use Babel instead of using pytz (directly)

lerks:
Babel wraps and enhances pytz: we can use its tools instead of some of
our own machinery and remove our (explicit) dependency on pytz.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/842)

### DIFF
--- a/cmscommon/datetime.py
+++ b/cmscommon/datetime.py
@@ -24,20 +24,20 @@ from __future__ import print_function
 from __future__ import unicode_literals
 from future.builtins.disabled import *
 from future.builtins import *
-from six import iteritems
 
 import os
-import sys
 import time
-from datetime import tzinfo, timedelta, datetime
-from pytz import timezone, all_timezones
+import sys
+from datetime import datetime
+
+import babel.dates
 
 
 __all__ = [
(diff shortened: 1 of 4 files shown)


stefano-maggiolo:
:lgtm:

---

Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[cmscommon/datetime.py, line 109 at r1](https://reviewable.io:443/reviews/cms-dev/cms/842#-L5cVRGlqvhkGOI6ITJe:-L5cVRGlqvhkGOI6ITJf:bkwwcqb) ([raw file](https://github.com/cms-dev/cms/blob/63923cbc08c37706ac7513d964476b3b98bda024/cmscommon/datetime.py#L109)):*
> ```Python
> 
>     See http://stackoverflow.com/questions/7669938/
>         get-the-olson-tz-name-for-the-local-timezone
> ```

This should be removed

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/842#-:-L5cVZzo5MveezuRTYp_:bnfp4nl)*

PR #843, opened 2018-02-18
Improve retrieval of information from shared-mime-info

lerks:
We use shared-mime-info to get two types of information:
* The translations of the descriptions of MIME types (e.g., "PDF
  document" becomes "Documento PDF" in Italian). We used to specify the
  exact location of the .mo files containing them, but the domain name
  appears to be enough (I believe this was the case even before we
  started using Babel, it's just that we didn't know it).
* The list of all MIME types, possible aliases, their natural language
  descriptions and their icon names. We were manually parsing all the
  relevant system files, of which we needed to know the exact locations.
  The xdg library can do most of these tasks for us. It doesn't provide
  information about the generic icon names (yet?), which means we have
  to keep retrieving it by hand, but even there it can help us figure
  out where the files are located.

This, in fact, removes our direct dependency on shared-mime-info, as we
have good fallbacks in place. And, anyways, it comes preinstalled by
default on all modern platforms.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/843)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,6 @@ addons:
       - fp-units-rtl
       - postgresql-client
       - python2.7
-      - shared-mime-info
       - stl-manual
       - cgroup-lite
       - libcap-dev
--- a/cms/conf.py
+++ b/cms/conf.py
@@ -137,6 +137,12 @@ class Config(object):
         self.max_submission_length = 100000
         self.max_input_length = 5000000
         self.stl_path = "/usr/share/doc/stl-manual/html/"
+        # Prefix of 'shared-mime-info'[1] installation. It can be found
+        # out using `pkg-config --variable=prefix shared-mime-info`, but
+        # it's almost universally the same (i.e. '/usr') so it's hardly
+        # necessary to change it.
+        # [1] http://freedesktop.org/wiki/Software/shared-mime-info
+        self.shared_mime_info_prefix = "/usr"
 
         # AdminWebServer.
         self.admin_listen_address = ""
--- a/cms/conf.py
+++ b/cms/conf.py
@@ -137,12 +137,6 @@ class Config(object):
         self.max_submission_length = 100000
         self.max_input_length = 5000000
         self.stl_path = "/usr/share/doc/stl-manual/html/"
-        # Prefix of 'shared-mime-info'[1] installation. It can be found
-        # out using `pkg-config --variable=prefix shared-mime-info`, but
-        # it's almost universally the same (i.e. '/usr') so it's hardly
-        # necessary to change it.
-        # [1] http://freedesktop.org/wiki/Software/shared-mime-info
-        self.shared_mime_info_prefix = "/usr"
 
         # AdminWebServer.
         self.admin_listen_address = ""
--- a/cms/locale/__init__.py
+++ b/cms/locale/__init__.py
@@ -24,11 +24,10 @@ from future.builtins.disabled import *
 from future.builtins import *
 
 from .locale import \
-    get_system_translations, Translation, DEFAULT_TRANSLATION, \
-    get_translations, filter_language_codes
+    Translation, DEFAULT_TRANSLATION, get_translations, filter_language_codes
 
 
 __all__ = [
-    "get_system_translations", "Translation", "DEFAULT_TRANSLATION",
-    "get_translations", "filter_language_codes"
+    "Translation", "DEFAULT_TRANSLATION", "get_translations",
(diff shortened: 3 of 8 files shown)


stefano-maggiolo:
:lgtm:

---

Review status: 0 of 8 files reviewed at latest revision, 2 unresolved discussions.

---

*[cmscommon/mimetypes.py, line 52 at r2](https://reviewable.io:443/reviews/cms-dev/cms/843#-L5dHwDFTflqgwbnhyC9:-L5dHwDFTflqgwbnhyCA:btcspc8) ([raw file](https://github.com/cms-dev/cms/blob/aaad9b056d6bdc2c6eb94c28c4289bd28e7c574c/cmscommon/mimetypes.py#L52)):*
> ```Python
>     except IOError as err:
>         if err.errno != errno.ENOENT:
>             raise
> ```

Optional: I think code as complicate as this should go in a function e.g.

```
def _cache_icons():
    [...]
    return icons

_icons = _cache_icons()
```

---

*[cmscommon/mimetypes.py, line 55 at r2](https://reviewable.io:443/reviews/cms-dev/cms/843#-L5dIP8WnyQzl7JxhuBn:-L5dIP8WnyQzl7JxhuBo:b-d4e5ec) ([raw file](https://github.com/cms-dev/cms/blob/aaad9b056d6bdc2c6eb94c28c4289bd28e7c574c/cmscommon/mimetypes.py#L55)):*
> ```Python
> 
> 
> def get_icon_for_type(typename):
> ```

Would be nice to have minimal docstring to make it obvious the format of typename and the return value. (As well later) 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/843#-:-L5dIijo1IWNlZSyAdgv:bnfp4nl)*

lerks:
I had to revert part of the PR because it turned out it didn't work inside a virtualenv: it looked for system translations inside Python's prefix, which for a virtualenv is the virtualenv's directory.

PTAL

---

Review status: 0 of 7 files reviewed at latest revision, 2 unresolved discussions.

---

*[cmscommon/mimetypes.py, line 52 at r2](https://reviewable.io:443/reviews/cms-dev/cms/843#-L5dHwDFTflqgwbnhyC9:-L5dXWOWioNjedK7rJ4x:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/aaad9b056d6bdc2c6eb94c28c4289bd28e7c574c/cmscommon/mimetypes.py#L52)):*

Optional: I think code as complicate as this should go in a function e.g.

```
def _cache_icons():
    [...]
    return icons

_icons = _cache_icons()
```
</blockquote></details>

Done.

---

*[cmscommon/mimetypes.py, line 55 at r2](https://reviewable.io:443/reviews/cms-dev/cms/843#-L5dIP8WnyQzl7JxhuBn:-L5dXXg7Isee36UzD4q4:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/aaad9b056d6bdc2c6eb94c28c4289bd28e7c574c/cmscommon/mimetypes.py#L55)):*

Would be nice to have minimal docstring to make it obvious the format of typename and the return value. (As well later) 
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/843#-:-L5dXZOSPdMG_k3jCKkr:bpbfy92)*

stefano-maggiolo:
Review status: 0 of 7 files reviewed at latest revision, 3 unresolved discussions.

---

*[cmscommon/mimetypes.py, line 90 at r3](https://reviewable.io:443/reviews/cms-dev/cms/843#-L5dZGbnXlxjZ2hzkCF-:-L5dZGboiUvE2l8Q-Q_6:b-8yzkk6) ([raw file](https://github.com/cms-dev/cms/blob/f7b88c9f2ebb52f59745978e42e4da7bb848803f/cmscommon/mimetypes.py#L90)):*
> ```Python
>     """Guess the MIME type of a file given its name.
> 
>     filename (str): the name of a file (just the base name, without the
> ```

|None, as well as in the return value.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/843)*

stefano-maggiolo:
:lgtm:

---

Review status: 0 of 7 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/843#-:-L5dZh8dlPZJkSX8gxom:bnfp4nl)*

PR #846, opened 2018-02-20
Continue porting the CWS templates from Tornado to Jinja2

lerks:
This is the second of a series of pull requests that ports CWS's templates from the Tornado engine to Jinja2. This pull request improves the way we do localization in the templates.
* Using context-aware filters allows us to reduce the boilerplate for most formatting operations.
* Using the `{% trans %}` blocks allows to better manage the escaping (e.g., no need to disable autoescaping when the message contains HTML) and the parameters injection.
* Finally, since Jinja2 provides a Babel extractor, we can reenable templates in Babel's mapping. The fact that the extractor does in fact perform a complete parsing of the templates, builds an AST and then extracts messages from there makes sure that they are all caught (unlike in Tornado, where xgettext was parsing the templates as if they were Python files and mysteriously missed some messages (it is in fact surprising that it *only* missed a couple)).

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/846)

### DIFF
--- a/cms/server/contest/templates/contest.html
+++ b/cms/server/contest/templates/contest.html
@@ -35,7 +35,7 @@ $(document).ready(function () {
     setInterval(function() {
         utils.update_time({{ "true" if contest.per_user_time is not none else "false" }});
     }, 1000);
-    utils.update_unread_count(0{% if request.path == '/communication' %}, 0{% endif %});
+    utils.update_unread_count(0{% if page == "communication" %}, 0{% endif %});
     utils.update_notifications(true);
     setInterval(function() { utils.update_notifications(); }, 30000);
     $('#main').css('top', $('#navigation_bar').outerHeight());
@@ -164,10 +164,10 @@ $(document).ready(function () {
                     <div class="well" style="padding: 8px 0;">
                         <ul class="nav nav-list">
 
-                            <li{% if request.path == '/' %} class="active"{% endif %}>
+                            <li{% if page == "overview" %} class="active"{% endif %}>
                                 <a href="{{ contest_url() }}">{% trans %}Overview{% endtrans %}</a>
                             </li>
-                            <li{% if request.path == '/communication' %} class="active"{% endif %}>
+                            <li{% if page == "communication" %} class="active"{% endif %}>
                                 <a href="{{ contest_url("communication") }}">{% trans %}Communication{% endtrans %}
                                     <span id="unread_count" class="label label-warning no_unread"></span>
                                 </a>
@@ -177,26 +177,26 @@ $(document).ready(function () {
                             <li class="nav-header">
                                 {{ t_iter.name }}
                             </li>
-                            <li{% if request.path == '/tasks/%s/description' % encode_for_url(t_iter.name) %} class="active"{% endif %}>
+                            <li{% if page == "task_description" and task == t_iter %} class="active"{% endif %}>
                                 <a href="{{ contest_url("tasks", t_iter.name, "description") }}">{% trans %}Statement{% endtrans %}</a>
                             </li>
-                            <li{% if request.path == '/tasks/%s/submissions' % encode_for_url(t_iter.name) %} class="active"{% endif %}>
+                            <li{% if page == "task_submissions" and task == t_iter %} class="active"{% endif %}>
                                 <a href="{{ contest_url("tasks", t_iter.name, "submissions") }}">{% trans %}Submissions{% endtrans %}</a>
                             </li>
         {% endfor %}
     {% endif %}
                             <li class="divider"></li>
-                            <li{% if request.path == '/documentation' %} class="active"{% endif %}>
+                            <li{% if page == "documentation" %} class="active"{% endif %}>
                                 <a href="{{ contest_url("documentation") }}">{% trans %}Documentation{% endtrans %}</a>
                             </li>
     {% if actual_phase == 0 or participation.unrestricted %}{# FIXME maybe >= 0? #}
         {% if testing_enabled %}
-                            <li{% if request.path == '/testing' %} class="active"{% endif %}>
+                            <li{% if page == "testing" %} class="active"{% endif %}>
                                 <a href="{{ contest_url("testing") }}">{% trans %}Testing{% endtrans %}</a>
                             </li>
         {% endif %}
         {% if printing_enabled %}
-                            <li{% if request.path == '/printing' %} class="active"{% endif %}>
+                            <li{% if page == "printing" %} class="active"{% endif %}>
                                 <a href="{{ contest_url("printing") }}">{% trans %}Printing{% endtrans %}</a>
                             </li>
         {% endif %}
--- a/cms/server/contest/templates/contest.html
+++ b/cms/server/contest/templates/contest.html
@@ -18,7 +18,7 @@ var LANGUAGES = {
 {% endfor %}
 };
 
-{% if current_user is none %}
+{% if participation is undefined %}
 var utils = new CMS.CWSUtils("{{ url() }}", "{{ contest_url() }}", "{{ contest.name }}",
                              0, 0, 0, 0, 0);
 {% else %}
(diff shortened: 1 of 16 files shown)


cms/server/contest/templates/contest.html
```diff
@@ -161,11 +165,11 @@ <h3 id="countdown_box">
...
 
                             <li{% if request.path == '/' %} class="active"{% endif %}>
-                                <a href="{{ contest_url() }}">{{ _("Overview") }}</a>
+                                <a href="{{ contest_url() }}">{% trans %}Overview{% endtrans %}</a>
                             </li>
                             <li{% if request.path == '/communication' %} class="active"{% endif %}>
-                                <a href="{{ contest_url("communication") }}">{{ _("Communication") }}
-                                    <span id="unread_count" class="label label-warning no_unread">{{ _("%d unread") % 0 }}</span>
+                                <a href="{{ contest_url("communication") }}">{% trans %}Communication{% endtrans %}
+                                    <span id="unread_count" class="label label-warning no_unread"></span>
```
stefano-maggiolo: Forgot something?

cms/server/contest/templates/contest.html
```diff
@@ -174,35 +178,35 @@ <h3 id="countdown_box">
...
                             </li>
         {% endif %}
     {% endif %}
                         </ul>
                     </div>
                     <span class="license_notice">
-                    <a href="http://github.com/cms-dev/cms/" rel="author noreferrer" target="_blank">{{ _("Contest Management System") }}</a>
-                    {{ _("is released under the") }}
-                    <a href="http://www.gnu.org/licenses/agpl" rel="license noreferrer" target="_blank">{{ _("GNU Affero General Public License") }}</a>
+                    <a href="http://github.com/cms-dev/cms/" rel="author noreferrer" target="_blank">{% trans %}Contest Management System{% endtrans %}</a>
```
stefano-maggiolo: Do we actually translate that?! It also probably needs to be translated in one piece instead of 3...

cms/server/contest/templates/overview.html
```diff
@@ -60,56 +78,56 @@ <h2>{{ _("General information") }}</h2>
...
+        {% trans %}You can see the detailed result of a submission by using a token on it.{% endtrans %}
+        {% trans %}Your score for each task will be the maximum among the tokened submissions and the last one.{% endtrans %}
         </p>
     {% else %}
         <p>
-        {% raw _("You have two types of tokens: a set of <em>contest-tokens</em> shared among all tasks and a distinct set of <em>task-tokens</em> for each task.") %}
+        {% trans %}You have two types of tokens: a set of <em>contest-tokens</em> shared among all tasks and a distinct set of <em>task-tokens</em> for each task.{% endtrans %}
         {{ contest|extract_token_params|format_token_rules(t_type="contest") }}
-        {{ _("You can find the rules for the %(type_pl)s on each task's description page.") % {"type_pl": _("task-tokens")} }}
+        {% trans type_pl=_("task-tokens") %}You can find the rules for the {{ type_pl }} on each task's description page.{% endtrans %}
```
stefano-maggiolo: Something weird going on here.

cms/server/contest/templates/overview.html
```diff
@@ -60,56 +78,56 @@ <h2>{{ _("General information") }}</h2>
...
-        {{ _("Your score for each task will be the maximum among the tokened submissions and the last one.") }}
+        {% trans %}You can see the detailed result of a submission by using a token on it.{% endtrans %}
+        {% trans %}Your score for each task will be the maximum among the tokened submissions and the last one.{% endtrans %}
         </p>
     {% elif tokens_contest == 2 %}
         <p>
-        {{ _("You have a distinct set of tokens for each task.") }}
-        {{ _("You can find the rules for the %(type_pl)s on each task's description page.") % {"type_pl": _("tokens")} }}
+        {% trans %}You have a distinct set of tokens for each task.{% endtrans %}
+        {% trans type_pl=_("tokens") %}You can find the rules for the {{ type_pl }} on each task's description page.{% endtrans %}
```
stefano-maggiolo: Something weird going on here.

cms/server/contest/templates/task_description.html
```diff
@@ -128,28 +128,28 @@ <h2>{{ _("Some details") }}</h2>
...
+        {% trans %}You have an infinite number of tokens.{% endtrans %}
         </p>
     {% elif tokens_contest == 2 %}
         <p>
         {{ task|extract_token_params|format_token_rules }}
         </p>
     {% elif tokens_tasks == 2 %}
         <p>
-        {% raw _("You can find the rules for the %(type_pl)s on the <a href=\"%(contest_root)s\">contest overview page</a>.") % {"type_pl": _("tokens"), "contest_root": contest_url()} %}
+        {% trans type_pl=_("tokens"), contest_root=contest_url() %}You can find the rules for the {{ type_pl }} on the <a href="{{ contest_root }}">contest overview page</a>.{% endtrans %}
```
stefano-maggiolo: Something weird going on here.

cms/server/contest/templates/task_description.html
```diff
@@ -128,28 +128,28 @@ <h2>{{ _("Some details") }}</h2>
...
     {% else %}
         <p>
         {{ task|extract_token_params|format_token_rules(t_type="task") }}
         </p>
 
         <p>
-        {{ _("Remember that to see the detailed result of a submission you need to use both a contest-token and a task-token.") }}
-        {% raw _("You can find the rules for the %(type_pl)s on the <a href=\"%(contest_root)s\">contest overview page</a>.") % {"type_pl": _("contest-tokens"), "contest_root": contest_url()} %}
+        {% trans %}Remember that to see the detailed result of a submission you need to use both a contest-token and a task-token.{% endtrans %}
+        {% trans type_pl=_("contest-tokens"), contest_root=contest_url() %}You can find the rules for the {{ type_pl }} on the <a href="{{ contest_root }}">contest overview page</a>.{% endtrans %}
```
stefano-maggiolo: Something weird going on here.

cms/server/contest/templates/task_submissions.html
```diff
@@ -168,33 +168,33 @@ <h2 style="margin: 40px 0 10px">{{ _("Previous submissions") }}</h2>
...
         {% endif %}
         {% if need_to_wait %}
             {% set t = tokens_info[2]|format_datetime_smart %}
-            {% raw _("But you have to wait until %(expiration_time)s to use them.") % {"expiration_time": t} %}
+            {% trans expiration_time=t %}But you have to wait until {{ expiration_time }} to use them.{% endtrans %}
         {% endif %}
         {% if tokens_info[1] is not none %}
             {% set t = tokens_info[1]|format_datetime_smart %}
-            {{ _("You will receive a new token at %(gen_time)s.") % {"gen_time": t} }}
+            {% trans gen_time=t %}You will receive a new token at {{ gen_time }}.{% endtrans %}
```
stefano-maggiolo: Can you avoid the double assignment? Also above

stefano-maggiolo: And below

cms/server/contest/templates/contest.html
lerks: Yes, I forgot to remove it earlier. Upon rendering, that message is invisible, because of the `no_unread` class. It is then JavaScript that updates its content and its class, loading the translated message from `div#translation_unread`. Hence this was useless.

cms/server/contest/templates/contest.html
lerks: Yes, and most locales even provide a translation. I don't know which are the other two places we localize it at, but if the messages are identical they appear as one in the .po files, which is OK.

cms/server/contest/templates/task_description.html
lerks: See above.

cms/server/contest/templates/task_description.html
lerks: See above.

cms/server/contest/templates/task_submissions.html
lerks: Done.

cms/server/contest/templates/overview.html
stefano-maggiolo: Yes, but shouldn't you get rid of _?

cms/server/contest/templates/task_submissions.html
stefano-maggiolo: Would be nice to wrap lines now that we can :P

cms/server/contest/templates/overview.html
lerks: Jinja's i18n extension provides the `_` symbol as an alias to the `gettext` function (that we provide, in the context). Hence this still works. (In fact, this whole commit wasn't technically necessary, as `{{ _("foo") }}` is pretty much the same as `{% trans %}foo{% endtrans %}`; the latter one is more idiomatic though).

cms/server/contest/templates/task_submissions.html
lerks: You mean like this?

stefano-maggiolo:
Yup, thank you, LGTM

lerks:
Merged into the `jinja` branch.

PR #847, opened 2018-02-20
Finish porting the CWS templates from Tornado to Jinja2

lerks:
This is the third of a series of pull requests that ports CWS's templates from the Tornado engine to Jinja2. This pull request fixes the last outstanding bugs and makes CWS usable again.

After this PR is approved the `jinja` branch will be merged into `master`. I have some further clean-up commits but I will submit them after that.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/847)

### DIFF
--- a/cms/grading/ScoreType.py
+++ b/cms/grading/ScoreType.py
@@ -43,9 +43,8 @@ import logging
 import re
 from abc import ABCMeta, abstractmethod
 
-from tornado.template import Template
-
 from cms.locale import DEFAULT_TRANSLATION
+from cms.server.contest.jinja2_toolbox import CWS_ENVIRONMENT
 
 
 logger = logging.getLogger(__name__)
@@ -79,6 +78,8 @@ class ScoreType(with_metaclass(ABCMeta, object)):
         # Preload the maximum possible scores.
         self.max_score, self.max_public_score, self.ranking_headers = \
             self.max_scores()
+
+        self.template = CWS_ENVIRONMENT.from_string(self.TEMPLATE)
 
     @staticmethod
     def format_score(score, max_score, unused_score_details,
@@ -118,14 +119,17 @@ class ScoreType(with_metaclass(ABCMeta, object)):
 
         """
         _ = translation.gettext
+        n_ = translation.ngettext
         if score_details is None:
             logger.error("Found a null score details string. "
                          "Try invalidating scores.")
             return _("Score details temporarily unavailable.")
         else:
-            return Template(self.TEMPLATE).generate(details=score_details,
-                                                    translation=translation,
-                                                    _=_)
+            # FIXME we should provide to the template all the variables
+            # of a typical CWS context as it's entitled to expect them.
+            return self.template.render(details=score_details,
+                                        translation=translation,
+                                        gettext=_, ngettext=n_)
 
     @abstractmethod
     def max_scores(self):
@@ -186,15 +190,14 @@ class ScoreTypeGroup(ScoreTypeAlone):
 
     """
     # Mark strings for localization.
-    N_("Subtask %d")
+    N_("Subtask %(index)s")
+    N_("#")
     N_("Outcome")
     N_("Details")
     N_("Execution time")
     N_("Memory used")
     N_("N/A")
     TEMPLATE = """\
-{% from cms.grading import format_status_text %}
-{% set idx = 0 %}
 {% for st in details %}
     {% if "score" in st and "max_score" in st %}
         {% if st["score"] >= st["max_score"] %}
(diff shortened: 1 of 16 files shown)


stefano-maggiolo:
Reviewed 3 of 3 files at r1, 5 of 5 files at r2, 1 of 1 files at r3, 4 of 4 files at r4, 9 of 9 files at r5, 2 of 2 files at r6.
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks broke.

---

*[cms/server/jinja2_toolbox.py, line 36 at r1](https://reviewable.io:443/reviews/cms-dev/cms/847#-L5rHQc927R_ZMJfCbQG:-L5rHQcAHD6CawTB8mI1:bb76tpr) ([raw file](https://github.com/cms-dev/cms/blob/287ed99a96452cf132cf08ad1a3cf1956299d3a6/cms/server/jinja2_toolbox.py#L36)):*
> ```Python
> 
> from jinja2 import Environment, StrictUndefined, contextfilter, contextfunction, \
>     environmentfunction
> ```

Wrap

---

*[cms/server/jinja2_toolbox.py, line 148 at r1](https://reviewable.io:443/reviews/cms-dev/cms/847#-L5rHUrd2rhc3jEkNyor:-L5rHUrd2rhc3jEkNyos:b4k2a57) ([raw file](https://github.com/cms-dev/cms/blob/287ed99a96452cf132cf08ad1a3cf1956299d3a6/cms/server/jinja2_toolbox.py#L148)):*
> ```Python
>     try:
>         return get_task_type(*args, **kwargs)
>     except Exception as err:
> ```

Can we narrow the exception, also below?

---

*[cms/server/jinja2_toolbox.py, line 149 at r1](https://reviewable.io:443/reviews/cms-dev/cms/847#-L5rHiu6-Dc9YIAMza3Z:-L5rHiu6-Dc9YIAMza3_:b-yzfyo8) ([raw file](https://github.com/cms-dev/cms/blob/287ed99a96452cf132cf08ad1a3cf1956299d3a6/cms/server/jinja2_toolbox.py#L149)):*
> ```Python
>         return get_task_type(*args, **kwargs)
>     except Exception as err:
>         return env.undefined("TaskType not found", exc=err)
> ```

Curious about this: will the exception be reported anywhere (since you're passing it to this function creating an undefined value, I assume)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/847)*

lerks:
Review status: 15 of 16 files reviewed at latest revision, 3 unresolved discussions, some commit checks broke.

---

*[cms/server/jinja2_toolbox.py, line 36 at r1](https://reviewable.io:443/reviews/cms-dev/cms/847#-L5rHQc927R_ZMJfCbQG:-L5uEaFcyG1k-Gcv2-l8:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/287ed99a96452cf132cf08ad1a3cf1956299d3a6/cms/server/jinja2_toolbox.py#L36)):*

Wrap
</blockquote></details>

Done.

---

*[cms/server/jinja2_toolbox.py, line 148 at r1](https://reviewable.io:443/reviews/cms-dev/cms/847#-L5rHUrd2rhc3jEkNyor:-L5uEaiNKcQBy6PXW1Fo:b-570lhp) ([raw file](https://github.com/cms-dev/cms/blob/287ed99a96452cf132cf08ad1a3cf1956299d3a6/cms/server/jinja2_toolbox.py#L148)):*

Can we narrow the exception, also below?
</blockquote></details>

No, because the task/score type's constructor is called, which may raise whatever exception it likes (in fact, score types raise StandardError in some cases when their parameters are malformed).

---

*[cms/server/jinja2_toolbox.py, line 149 at r1](https://reviewable.io:443/reviews/cms-dev/cms/847#-L5rHiu6-Dc9YIAMza3Z:-L5uGwoDo3nARGIHgD7X:b-yfmhjy) ([raw file](https://github.com/cms-dev/cms/blob/287ed99a96452cf132cf08ad1a3cf1956299d3a6/cms/server/jinja2_toolbox.py#L149)):*

Curious about this: will the exception be reported anywhere (since you're passing it to this function creating an undefined value, I assume)
</blockquote></details>

I actually though so, but since you asked I checked and that is actually not the case. The exc value should be a subclass of UndefinedError and it can be used to provide a custom exception. It is not used for detailed reporting. So I formatted the error into the message.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/847)*

stefano-maggiolo:
:lgtm:

---

Review status: 15 of 16 files reviewed at latest revision, all discussions resolved, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/847#-:-L5uKZtYc_d-TzSu3zX5:bnfp4nl)*

lerks:
Pushed.

PR #853, opened 2018-02-22
Batch: use a brand-new sandbox for the checking phase

wil93:
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/853)

### DIFF
--- a/cms/grading/Sandbox.py
+++ b/cms/grading/Sandbox.py
@@ -41,7 +41,6 @@ from functools import wraps, partial
 
 import gevent
 from gevent import subprocess
-#import gevent_subprocess as subprocess
 
 from cms import config
 from cms.io.GeventUtils import copyfileobj, rmtree
--- a/cms/grading/Sandbox.py
+++ b/cms/grading/Sandbox.py
@@ -200,20 +200,22 @@ class SandboxBase(with_metaclass(ABCMeta, object)):
     EXIT_SYSCALL = 'syscall'
     EXIT_NONZERO_RETURN = 'nonzero return'
 
-    def __init__(self, multithreaded, file_cacher, temp_prefix=None, temp_dir=None):
+    def __init__(self, multithreaded, file_cacher, name=None, temp_dir=None):
         """Initialization.
 
         multithreaded (boolean): whether the sandbox should allow
             multithreading.
         file_cacher (FileCacher): an instance of the FileCacher class
             (to interact with FS), if the sandbox needs it.
+        name (string|None): name of the sandbox, which might appear in the
+            path and in system logs.
         temp_dir (unicode|None): temporary directory to use; if None, use the
             default temporary directory specified in the configuration.
 
         """
         self.multithreaded = multithreaded
         self.file_cacher = file_cacher
-        self.temp_prefix = temp_prefix if temp_prefix is not None else "tmp"
+        self.name = name if name is not None else "unnamed"
         self.temp_dir = temp_dir if temp_dir is not None else config.temp_dir
 
         self.cmd_file = "commands.log"
@@ -869,13 +871,13 @@ class IsolateSandbox(SandboxBase):
     # on the current directory.
     SECURE_COMMANDS = ["/bin/cp", "/bin/mv", "/usr/bin/zip", "/usr/bin/unzip"]
 
-    def __init__(self, multithreaded, file_cacher, temp_prefix=None, temp_dir=None):
+    def __init__(self, multithreaded, file_cacher, name=None, temp_dir=None):
         """Initialization.
 
         For arguments documentation, see SandboxBase.__init__.
 
         """
-        SandboxBase.__init__(self, multithreaded, file_cacher, temp_prefix, temp_dir)
+        SandboxBase.__init__(self, multithreaded, file_cacher, name, temp_dir)
 
         # Isolate only accepts ids between 0 and 99. We assign the
         # range [(shard+1)*10, (shard+2)*10) to each Worker and keep
@@ -898,7 +900,9 @@ class IsolateSandbox(SandboxBase):
         # system to, which is why the outer directory exists with no read
         # permissions.
         self.inner_temp_dir = "/tmp"
-        self.outer_temp_dir = tempfile.mkdtemp(dir=self.temp_dir, prefix=self.temp_prefix)
+        self.outer_temp_dir = tempfile.mkdtemp(
+            dir=self.temp_dir,
+            prefix="CMS-%s-" % (self.name))
         # Don't use os.path.join here, because the absoluteness of /tmp will
         # bite you.
         self.path = self.outer_temp_dir + self.inner_temp_dir
--- a/cms/grading/Sandbox.py
+++ b/cms/grading/Sandbox.py
@@ -544,16 +544,18 @@ class StupidSandbox(SandboxBase):
 
     """
 
-    def __init__(self, multithreaded, file_cacher, temp_dir=None):
+    def __init__(self, multithreaded, file_cacher, name=None, temp_dir=None):
         """Initialization.
 
         For arguments documentation, see SandboxBase.__init__.
 
         """
-        SandboxBase.__init__(self, multithreaded, file_cacher, temp_dir)
+        SandboxBase.__init__(self, multithreaded, file_cacher, name, temp_dir)
 
         # Make box directory
-        self.path = tempfile.mkdtemp(dir=self.temp_dir)
+        self.path = tempfile.mkdtemp(
+            dir=self.temp_dir,
+            prefix="cms-%s-" % (self.name))
 
         self.exec_num = -1
         self.popen = None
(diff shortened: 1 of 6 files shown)


stefano-maggiolo:
Please also fix the style (there are nice recipes in the wiki to run linters every time you commit).

The file is in need of some refactoring to limit depth, but let's do it after this.

---

Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[cms/grading/tasktypes/Batch.py, line 316 at r2](https://reviewable.io:443/reviews/cms-dev/cms/853#-L5wjI8fFdxfcph_8HSb:-L5wjI8fFdxfcph_8HSc:b-6vpihr) ([raw file](https://github.com/cms-dev/cms/blob/8c14ceef357747d28c3302acbd0e4fcbe0fdbb78/cms/grading/tasktypes/Batch.py#L316)):*
> ```Python
> 
>                     # Create the checkbox: a brand-new sandbox, just for checking
>                     checkbox = create_sandbox(file_cacher, False)  # do we need multithreading?
> ```

No harm in enabling it, since it's all admin code, right?

---

*[cms/grading/tasktypes/Batch.py, line 317 at r2](https://reviewable.io:443/reviews/cms-dev/cms/853#-L5wkKHHZC75H8qx2OE3:-L5wkKHHZC75H8qx2OE4:bh4429p) ([raw file](https://github.com/cms-dev/cms/blob/8c14ceef357747d28c3302acbd0e4fcbe0fdbb78/cms/grading/tasktypes/Batch.py#L317)):*
> ```Python
>                     # Create the checkbox: a brand-new sandbox, just for checking
>                     checkbox = create_sandbox(file_cacher, False)  # do we need multithreading?
>                     checker_didnt_crash = True
> ```

`checker_success`?

---

*[cms/grading/tasktypes/Batch.py, line 330 at r2](https://reviewable.io:443/reviews/cms-dev/cms/853#-L5wk78DGVvCi0VtEa00:-L5wk78DGVvCi0VtEa01:b8f0mgg) ([raw file](https://github.com/cms-dev/cms/blob/8c14ceef357747d28c3302acbd0e4fcbe0fdbb78/cms/grading/tasktypes/Batch.py#L330)):*
> ```Python
> 
>                     # Put the user-produced output file into the checkbox
>                     shutil.copyfile(
> ```

If your goal was to avoid symlink attacks, this copy follows symlinks.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/853#-:-L5wklecfpZS1q9Z5IZn:ba70cvv)*

wil93:
For the style, apparently my IDE already has linting enabled on save (and it should use pylint by default) but it doesn't complain about the long lines... I think we would need to create a `pylintrc` file in the root of the repository with the rules to check. Who wants to do that? :stuck_out_tongue: 

---

Review status: 0 of 6 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[cms/grading/tasktypes/Batch.py, line 316 at r2](https://reviewable.io:443/reviews/cms-dev/cms/853#-L5wjI8fFdxfcph_8HSb:-L5wwKMXiFVK9F-ow-tT:b32w0x) ([raw file](https://github.com/cms-dev/cms/blob/8c14ceef357747d28c3302acbd0e4fcbe0fdbb78/cms/grading/tasktypes/Batch.py#L316)):*

No harm in enabling it, since it's all admin code, right?
</blockquote></details>

OK

---

*[cms/grading/tasktypes/Batch.py, line 317 at r2](https://reviewable.io:443/reviews/cms-dev/cms/853#-L5wkKHHZC75H8qx2OE3:-L5wwLkXFkI8qS3vq7d8:bnpjbjv) ([raw file](https://github.com/cms-dev/cms/blob/8c14ceef357747d28c3302acbd0e4fcbe0fdbb78/cms/grading/tasktypes/Batch.py#L317)):*

`checker_success`?
</blockquote></details>

I thought about it but I wanted to avoid any confusion with the actual "success" of the checking process (as in: true iff the contestant has solved this input)

---

*[cms/grading/tasktypes/Batch.py, line 330 at r2](https://reviewable.io:443/reviews/cms-dev/cms/853#-L5wk78DGVvCi0VtEa00:-L5wul79jwwO2SUV6w7X:b-7tbcca) ([raw file](https://github.com/cms-dev/cms/blob/8c14ceef357747d28c3302acbd0e4fcbe0fdbb78/cms/grading/tasktypes/Batch.py#L330)):*

If your goal was to avoid symlink attacks, this copy follows symlinks.
</blockquote></details>

If I understand correctly, here it's better to follow symlinks (so the copy will fail... I have to check this special case actually) because if I didn't follow symlinks then someone could produce a symlink to `res.txt`

---

*[cms/grading/tasktypes/Batch.py, line 321 at r4](https://reviewable.io:443/reviews/cms-dev/cms/853#-L5waqideRf_kWve_3qE-r4-321:-L5wwtT-eGEOQtOVCtgu:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/06f0997d037387078b16af601033b9f2d69dd5f3/cms/grading/tasktypes/Batch.py#L321)):*

Codacy Issue found: [Line too long (81/80)](https://app.codacy.com/app/s-maggiolo/cms/pullRequest?prid=1361148)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/853#-:-L5wv2rdmebtVb_0gxP1:bavl774)*

stefano-maggiolo:
PS: next time can you open a new PR for a new feature rather than adding it here?

---

Review status: 0 of 6 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[cms/grading/Sandbox.py, line 872 at r3](https://reviewable.io:443/reviews/cms-dev/cms/853#-L5xm5OuMgXlT137QTKd:-L5xm5OuMgXlT137QTKe:b-kw5jn2) ([raw file](https://github.com/cms-dev/cms/blob/1c9535c28dd99bd7188a74b6dfbe40174e51fa6d/cms/grading/Sandbox.py#L872)):*
> ```Python
>     SECURE_COMMANDS = ["/bin/cp", "/bin/mv", "/usr/bin/zip", "/usr/bin/unzip"]
> 
>     def __init__(self, multithreaded, file_cacher, temp_prefix=None, temp_dir=None):
> ```

Can you add to stupidsandbox too, so they keep the same interface?

---

*[cms/grading/TaskType.py, line 56 at r3](https://reviewable.io:443/reviews/cms-dev/cms/853#-L5xlQAoGttdDHzXkeet:-L5xlQAoGttdDHzXkeeu:b-c5tmpf) ([raw file](https://github.com/cms-dev/cms/blob/1c9535c28dd99bd7188a74b6dfbe40174e51fa6d/cms/grading/TaskType.py#L56)):*
> ```Python
> ## Sandbox lifecycle. ##
> 
> def create_sandbox(file_cacher, temp_prefix=None, multithreaded=False):
> ```

Add at the end, so that it doesn't hurt if somebody uses multithreaded without the keyword.

---

*[cms/grading/tasktypes/Batch.py, line 317 at r2](https://reviewable.io:443/reviews/cms-dev/cms/853#-L5wkKHHZC75H8qx2OE3:-L5xn8mb56WYTd1nJXAm:b-eg99t8) ([raw file](https://github.com/cms-dev/cms/blob/8c14ceef357747d28c3302acbd0e4fcbe0fdbb78/cms/grading/tasktypes/Batch.py#L317)):*

I thought about it but I wanted to avoid any confusion with the actual "success" of the checking process (as in: true iff the contestant has solved this input)
</blockquote></details>

We already use `success` from the point of view of CMS, not the contestant's, so it's safe to use here too - but mostly "crash" is not the only failure mode, so it's misleading to name it like that.

---

*[cms/grading/tasktypes/Batch.py, line 330 at r2](https://reviewable.io:443/reviews/cms-dev/cms/853#-L5wk78DGVvCi0VtEa00:-L5xok_Ms-QSoy2hwhyd:b-ner241) ([raw file](https://github.com/cms-dev/cms/blob/8c14ceef357747d28c3302acbd0e4fcbe0fdbb78/cms/grading/tasktypes/Batch.py#L330)):*

If I understand correctly, here it's better to follow symlinks (so the copy will fail... I have to check this special case actually) because if I didn't follow symlinks then someone could produce a symlink to `res.txt`
</blockquote></details>

Right, I wonder though if it makes sense to test that we're not following symlinks (https://docs.python.org/3/library/os.path.html#os.path.islink)

---

*[cms/grading/tasktypes/Batch.py, line 166 at r3](https://reviewable.io:443/reviews/cms-dev/cms/853#-L5xlaj9q76dFYtlXoZ4:-L5xlaj9q76dFYtlXoZ5:b8sucqj) ([raw file](https://github.com/cms-dev/cms/blob/1c9535c28dd99bd7188a74b6dfbe40174e51fa6d/cms/grading/tasktypes/Batch.py#L166)):*
> ```Python
>         sandbox = create_sandbox(
>             file_cacher,
>             temp_prefix="CMS-compile-",
> ```

Nit: I prefer lowercase (as we usually write cms in the filesystem)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/853#-:-L5xklHnpRG3i_Og-xbV:b59w6cc)*

stefano-maggiolo:
Re linting: any tool following PEP8 should give you warning for passing 80 columns, it's weird if it doesn't. With pylint we have so many positives that it's not very informative to run, flake8 usually works nicely.

stefano-maggiolo:
:lgtm:

---

Review status: 0 of 6 files reviewed at latest revision, 6 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/853#-:-L60S5SkbZ37LIu4yRE0:bnfp4nl)*

stefano-maggiolo:
Merged

PR #854, opened 2018-02-22
Respond with HTTP error 405 for unimplemented methods

### LINKED ISSUE #810: HTTP 500 on invalid GET /<contest>/login request
When CWS is running in multicontest mode, sending an (invalid) `GET /<contest>/login` request results in internal server error:

```
2017-12-03 18:57:27,577 - ERROR [Contest,0 5 web::log_exception] Uncaught exception GET /test/login (10.0.2.2)
HTTPServerRequest(protocol='http', host='localhost:8080', method='GET', uri='/test/login', version='HTTP/1.1', remote_ip='10.0.2.2', headers={…})
  File "/usr/lib/python2.7/dist-packages/tornado/web.py", line 1413, in _execute
    result = method(*self.path_args, **self.path_kwargs)
TypeError: get() takes exactly 1 argument (2 given)
2017-12-03 18:57:27,594 - ERROR [Contest,0 5 base::write_error] Uncaught exception (TypeError('get() takes exactly 1 argument (2 given)',)) while processing a request: Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/tornado/web.py", line 1413, in _execute
    result = method(*self.path_args, **self.path_kwargs)
TypeError: get() takes exactly 1 argument (2 given)

2017-12-03 18:57:27,604 - ERROR [Contest,0 5 web::log_request] 500 GET /test/login (10.0.2.2) 32.81ms
```

Probably this should return HTTP status code 405 Method Not Allowed or something else instead of 500 Internal Server Error.

I think this goes like this: `LoginHandler` -> `ContestHandler` -> `BaseHandler` -> `get()`, which doesn't have `@multicontest` decorator but is passed the contest name.

CC @wil93

wil93:
Fixes #810

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/854)

### DIFF
--- a/cms/server/contest/handlers/contest.py
+++ b/cms/server/contest/handlers/contest.py
@@ -97,6 +97,14 @@ class ContestHandler(BaseHandler):
         super(ContestHandler, self).__init__(*args, **kwargs)
         self.contest_url = None
 
+    def get(self, *args, **kwargs):
+        # This method should be implemented by child classes
+        raise tornado.web.HTTPError(405, "Method Not Allowed")
+
+    def post(self, *args, **kwargs):
+        # This method should be implemented by child classes
+        raise tornado.web.HTTPError(405, "Method Not Allowed")
+
     def prepare(self):
         self.choose_contest()
 


stefano-maggiolo:
Only nit: should it be a 404 to minimize information? Not that it matters much...

wil93:
¯\\\_(ツ)_/¯

lerks:
This is the wrong fix to the problem. Tornado already raises an error 405 when a method is undefined. The problem, here, is that a get method is defined on the BaseHandler, when it shouldn't. The correct fix is moving that method to its own handler.

stefano-maggiolo:
Agree, but this is still an improvement to the status quo, available now. Let's merge with a todo to move to the right structure.

If tornado does 405, let's use that.

stefano-maggiolo:
Merged

PR #856, opened 2018-02-22
Fix broken prerequisites.py introduced by Py2+3 code

wil93:
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/856)

### DIFF
--- a/prerequisites.py
+++ b/prerequisites.py
@@ -37,6 +37,17 @@ from __future__ import unicode_literals
 # the user will be using and in which the requirements were installed
 # (e.g. the site installation rather than a venv).
 
+### Start Python 2+3 compatibility
+
+# Equivalent to running: from future.builtins import input
+try:
+    input = raw_input
+    del raw_input
+except NameError:
+    pass
+
+### End Python 2+3 compatibility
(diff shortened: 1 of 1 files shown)


lerks:
Nice catch!

Though, I did the py2->py3 conversion and know that code in and out, and it still took me a bit to realize why this fix was necessary. Can you add a comment explaining it better?

Also, I have personally never seen an except for NameError, it feels a bit weird. I would prefer `if sys.version_info < (3.0): ...`, and maybe you can even put it just around the only instance where we use input.

wil93:
I think I've seen NameError being used in this way before (`statement` without side effect, followed by some action in the right Python version, and the wrong version of Python will throw a NameError) I even remembered it was a NameError without having to look it up... but if you personally prefer a different equivalent implementation you are free to change it :smile:

lerks:
Ok, fine, leave the code as it is but please add a more useful comment.

wil93:
Done

lerks:
Pushed (at last), as 289fbd94b693945418e6116fb27981f32d18257b.

PR #858, opened 2018-02-22
Improve cmstestsuite determinism and usability

bblackham:
These commits add the ability to run individual unit tests, and make the test ordering deterministic (not dependent on filesystem inode ordering). Unfortunately, this also exposes the fact that ImportTaskTest and ImportContestTest can fail unless ImportDatasetTest has run first. And even then, there are some weird dependency issues I haven't figured out yet. So this commit causes the unit tests to fail. But I'm hoping someone else can fix the failure who better understands what exploded! This reproduces it on a fresh database:

`$ TEST_SUITE=unittests cmsRunTests ImportTask
`

If you run the tests in reverse order, they all pass, but I'd prefer to get the failure fixed (presumably before this can be merged!) Or we can not merge the last commit in the sequence (that sorts the tests), but then it's still a ticking time bomb.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/858)

### DIFF
--- a/cmstestsuite/RunUnitTests.py
+++ b/cmstestsuite/RunUnitTests.py
@@ -174,7 +174,7 @@ def main():
 
     if args.dry_run:
         for t in test_list:
-            print(t[0].name, t[1])
+            print(t[0], t[1])
         return 0
 
     if args.retry_failed:
--- a/cmstestsuite/RunUnitTests.py
+++ b/cmstestsuite/RunUnitTests.py
@@ -115,7 +115,7 @@ def get_all_tests():
     tests = []
     files = sorted(os.walk(os.path.join("cmstestsuite", "unit_tests")))
     for path, _, names in files:
-        for name in names:
+        for name in sorted(names):
             if name.endswith(".py"):
                 tests.append((path, name))
     return tests
--- a/cmstestsuite/RunUnitTests.py
+++ b/cmstestsuite/RunUnitTests.py
@@ -29,6 +29,7 @@ import argparse
 import io
 import logging
 import os
+import re
 import sys
 import subprocess
 import datetime
@@ -171,6 +172,15 @@ def main():
         test_list = load_failed_tests()
     else:
         test_list = get_all_tests()
+
+    if args.regex:
+        # Require at least one regex to match to include it in the list.
+        filter_regexps = [re.compile(regex) for regex in args.regex]
+
+        def test_match(t):
+            return any([r.search(t) is not None for r in filter_regexps])
+
+        test_list = [t for t in test_list if test_match(' '.join(t))]
 
     if args.dry_run:
         for t in test_list:
--- a/cmstestsuite/RunUnitTests.py
+++ b/cmstestsuite/RunUnitTests.py
@@ -113,7 +113,8 @@ def load_test_list_from_file(filename):
 
 def get_all_tests():
     tests = []
-    for path, _, names in os.walk(os.path.join("cmstestsuite", "unit_tests")):
+    files = sorted(os.walk(os.path.join("cmstestsuite", "unit_tests")))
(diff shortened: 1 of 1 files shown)


bblackham:
Well apparently these tests do pass on Travis despite being in the same order that fails for me. So I've now no idea what is causing the tests to fail on my local machine (see attached). I'll file this as a separate issue.

[log.txt](https://github.com/cms-dev/cms/files/1748390/log.txt)

stefano-maggiolo:
Not sure what is happening in your logs, but it doesn't seem to be order-dependent.

It is caused by setting the task id to null when deleting all rows in all tables between unittests, but I don't understand why the task id should be set to null (I'd understand if it was a foreign key and you deleted the referred row before the referring).

stefano-maggiolo:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[cmstestsuite/RunUnitTests.py, line 176 at r2](https://reviewable.io:443/reviews/cms-dev/cms/858#-L67NcWSQczr1BapUdTS:-L67NcWSQczr1BapUdTT:b-512t32) ([raw file](https://github.com/cms-dev/cms/blob/d87b3b0f22689b5ca1f1411b2b6813268b43cbe5/cmstestsuite/RunUnitTests.py#L176)):*
> ```Python
>         test_list = get_all_tests()
> 
>     if args.regex:
> ```

Can you move the argument definition from the "unused" section and copy the help from the functional tests?

---

*[cmstestsuite/RunUnitTests.py, line 181 at r2](https://reviewable.io:443/reviews/cms-dev/cms/858#-L67NUUEPHz9eGsFigjo:-L67NUUEPHz9eGsFigjp:b-3zmdp5) ([raw file](https://github.com/cms-dev/cms/blob/d87b3b0f22689b5ca1f1411b2b6813268b43cbe5/cmstestsuite/RunUnitTests.py#L181)):*
> ```Python
> 
>         def test_match(t):
>             return any([r.search(t) is not None for r in filter_regexps])
> ```

You can get rid of the []

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/858)*

bblackham:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[cmstestsuite/RunUnitTests.py, line 176 at r2](https://reviewable.io:443/reviews/cms-dev/cms/858#-L67NcWSQczr1BapUdTS:-L67goIMHYoeKFzGPRv2:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/d87b3b0f22689b5ca1f1411b2b6813268b43cbe5/cmstestsuite/RunUnitTests.py#L176)):*

Can you move the argument definition from the "unused" section and copy the help from the functional tests?
</blockquote></details>

Done.

---

*[cmstestsuite/RunUnitTests.py, line 181 at r2](https://reviewable.io:443/reviews/cms-dev/cms/858#-L67NUUEPHz9eGsFigjo:-L67gon9GsTwGe2VeGEj:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/d87b3b0f22689b5ca1f1411b2b6813268b43cbe5/cmstestsuite/RunUnitTests.py#L181)):*

You can get rid of the []
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/858)*

stefano-maggiolo:
:lgtm:

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/858#-:-L6GwQa2Pq9Gmz3mJieK:bnfp4nl)*

stefano-maggiolo:
Merged!

PR #862, opened 2018-02-23
Add shared syntactical validation of task type parameters

stefano-maggiolo:
Added this while working on a refactoring of Batch (see https://github.com/stefano-maggiolo/cms/tree/split_batch for the final form).

AWS is still able to show the dataset page for tasks with invalid parameters (it will display default values instead). There is no indication that the underlying data is bad, but this isn't new.

### DIFF
--- a/cms/grading/ParameterTypes.py
+++ b/cms/grading/ParameterTypes.py
@@ -34,7 +34,7 @@ from __future__ import print_function
 from __future__ import unicode_literals
 from future.builtins.disabled import *
 from future.builtins import *
-from six import with_metaclass
+from six import integer_types, string_types, with_metaclass
 
 from abc import ABCMeta, abstractmethod
 
@@ -59,6 +59,17 @@ class ParameterType(with_metaclass(ABCMeta, object)):
         self.name = name
         self.short_name = short_name
         self.description = description
+
+    @abstractmethod
+    def validate(self, value):
+        """Validate that the passed value is syntactically appropriate
+
+        value (object): the value to test
+
+        raise (ValueError): if the value is malformed for this parameter.
+
+        """
+        pass
 
     @abstractmethod
     def parse_string(self, value):
@@ -94,6 +105,11 @@ class ParameterTypeString(ParameterType):
     TEMPLATE = "<input type=\"text\" name=\"{{parameter_name}}\" " \
         "value=\"{{parameter_value}}\" />"
 
+    def validate(self, value):
+        if not isinstance(value, string_types):
+            raise ValueError("Invalid value for string parameter %s" %
+                             self.name)
+
     def parse_string(self, value):
         """Returns the specified string.
         """
@@ -110,6 +126,11 @@ class ParameterTypeFloat(ParameterType):
 
     TEMPLATE = "<input type=\"text\" name=\"{{parameter_name}}\" " \
         "value=\"{{parameter_value}}\" />"
+
+    def validate(self, value):
+        if not isinstance(value, float):
+            raise ValueError("Invalid value for float parameter %s" %
+                             self.name)
 
     def parse_string(self, value):
         """Attempts to parse the specified string as a float and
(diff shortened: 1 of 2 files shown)


cms/grading/ParameterTypes.py
```diff
@@ -129,6 +150,10 @@ class ParameterTypeInt(ParameterType):
     TEMPLATE = "<input type=\"text\" name=\"{{parameter_name}}\" " \
         "value=\"{{parameter_value}}\" />"
 
+    def validate(self, value):
+        if not isinstance(value, integer_types):
```
lerks: Not really necessary:
```
In [1]: i = 1l

In [2]: isinstance(i, int)
Out[2]: False

In [3]: from future.builtins import *

In [4]: isinstance(i, int)
Out[4]: True
```

cms/grading/TaskType.py
```diff
@@ -152,6 +152,23 @@ def __init__(self, parameters):
...
         self.parameters = parameters
+        self.validate_parameters()
+
+    def validate_parameters(self):
+        """Validate the parameters syntactically.
+
+        raise (ValueError): if the parameters are malformed.
+
+        """
+        if len(self.parameters) != len(self.ACCEPTED_PARAMETERS):
```
lerks: Wanna add a check that `self.parameters` is a list?

cms/grading/ParameterTypes.py
```diff
@@ -94,6 +105,11 @@ class ParameterTypeString(ParameterType):
     TEMPLATE = "<input type=\"text\" name=\"{{parameter_name}}\" " \
         "value=\"{{parameter_value}}\" />"
 
+    def validate(self, value):
+        if not isinstance(value, string_types):
```
lerks: Do you know of anything that would break if you used `str` here (meaning that they had to be `unicode` in py2)?

stefano-maggiolo:
Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions.

---

*[cms/grading/ParameterTypes.py, line 109 at r1](https://reviewable.io:443/reviews/cms-dev/cms/862#-L63eegtH2Gl79PQVdUO-r1-109:-L63jylW19X71lRbMoZh:bmjutov) ([raw file](https://github.com/cms-dev/cms/blob/57cc00bb83b10745f19dbdb56b011ebcc2bb13e6/cms/grading/ParameterTypes.py#L109)):*

Do you know of anything that would break if you used `str` here (meaning that they had to be `unicode` in py2)?
</blockquote></details>

To be honest I was thinking at text_types, but I used string_types... To be on the safe side given this paths are not well-tested I'd leave the generic one.

---

*[cms/grading/ParameterTypes.py, line 154 at r1](https://reviewable.io:443/reviews/cms-dev/cms/862#-L63eegtH2Gl79PQVdUO-r1-154:-L63kao4Bb8BAqXOghY4:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/57cc00bb83b10745f19dbdb56b011ebcc2bb13e6/cms/grading/ParameterTypes.py#L154)):*

Not really necessary:
```
In [1]: i = 1l

In [2]: isinstance(i, int)
Out[2]: False

In [3]: from future.builtins import *

In [4]: isinstance(i, int)
Out[4]: True
```
</blockquote></details>

Done.

---

*[cms/grading/TaskType.py, line 163 at r1](https://reviewable.io:443/reviews/cms-dev/cms/862#-L63eegtH2Gl79PQVdUP-r1-163:-L63krh0swmrHcNhG31j:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/57cc00bb83b10745f19dbdb56b011ebcc2bb13e6/cms/grading/TaskType.py#L163)):*

Wanna add a check that `self.parameters` is a list?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/862)*

stefano-maggiolo:
One step at a time :)

PR #876, opened 2018-03-01
Some work around dumps

stefano-maggiolo:
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/876)

### DIFF
--- a/cmscontrib/DumpExporter.py
+++ b/cmscontrib/DumpExporter.py
@@ -32,7 +32,7 @@ from __future__ import print_function
 from __future__ import unicode_literals
 from future.builtins.disabled import *
 from future.builtins import *
-from six import itervalues, iteritems
+from six import PY3, itervalues, iteritems
 
 # We enable monkey patching to make many libraries gevent-friendly
 # (for instance, urllib3, used by requests)
@@ -252,9 +252,15 @@ class DumpExporter(object):
 
                 data["_version"] = model_version
 
-                with io.open(os.path.join(export_dir, "contest.json"),
-                             "wt", encoding="utf-8") as fout:
-                    json.dump(data, fout, indent=4, sort_keys=True)
+                if PY3:
+                    with io.open(os.path.join(export_dir, "contest.json"),
+                                 "wt", encoding="utf-8") as fout:
+                        json.dump(data, fout, indent=4, sort_keys=True)
+                else:
+                    with io.open(os.path.join(export_dir, "contest.json"),
+                                 "wb") as fout:
+                        json.dump(data, fout, indent=4, sort_keys=True,
+                                  encoding="utf-8")
 
         # If the admin requested export to file, we do that.
         if archive_info["write_mode"] != "":
--- a/cmscontrib/DumpExporter.py
+++ b/cmscontrib/DumpExporter.py
@@ -122,9 +122,8 @@ def encode_value(type_, value):
     """
     if value is None:
         return None
-    elif isinstance(type_, (Boolean, Integer, Float, Unicode, Enum, JSONB)):
-        return value
-    elif isinstance(type_, String):
+    elif isinstance(type_, (
+            Boolean, Integer, Float, String, Unicode, Enum, JSONB)):
         return value
     elif isinstance(type_, DateTime):
         return make_timestamp(value)
--- a/cmscontrib/DumpExporter.py
+++ b/cmscontrib/DumpExporter.py
@@ -252,15 +252,13 @@ class DumpExporter(object):
 
                 data["_version"] = model_version
 
+                destination = os.path.join(export_dir, "contest.json")
                 if PY3:
-                    with io.open(os.path.join(export_dir, "contest.json"),
-                                 "wt", encoding="utf-8") as fout:
+                    with io.open(destination, "wt", encoding="utf-8") as fout:
                         json.dump(data, fout, indent=4, sort_keys=True)
                 else:
-                    with io.open(os.path.join(export_dir, "contest.json"),
-                                 "wb") as fout:
-                        json.dump(data, fout, indent=4, sort_keys=True,
-                                  encoding="utf-8")
+                    with io.open(destination, "wb") as fout:
+                        json.dump(data, fout, indent=4, sort_keys=True)
 
         # If the admin requested export to file, we do that.
         if archive_info["write_mode"] != "":
--- a/cmscontrib/DumpExporter.py
+++ b/cmscontrib/DumpExporter.py
@@ -125,7 +125,7 @@ def encode_value(type_, value):
     elif isinstance(type_, (Boolean, Integer, Float, Unicode, Enum, JSONB)):
         return value
     elif isinstance(type_, String):
-        return value.decode('latin1')
+        return value
     elif isinstance(type_, DateTime):
         return make_timestamp(value)
     elif isinstance(type_, Interval):
@@ -253,8 +253,8 @@ class DumpExporter(object):
 
                 data["_version"] = model_version
 
-                with io.open(os.path.join(export_dir,
-                                          "contest.json"), "wb") as fout:
+                with io.open(os.path.join(export_dir, "contest.json"),
+                             "wt", encoding="utf-8") as fout:
                     json.dump(data, fout, indent=4, sort_keys=True)
 
         # If the admin requested export to file, we do that.
--- a/cmscontrib/DumpImporter.py
+++ b/cmscontrib/DumpImporter.py
@@ -108,9 +108,8 @@ def decode_value(type_, value):
     """
     if value is None:
         return None
-    elif isinstance(type_, (Boolean, Integer, Float, Unicode, Enum, JSONB)):
-        return value
-    elif isinstance(type_, String):
+    elif isinstance(type_, (
+            Boolean, Integer, Float, String, Unicode, Enum, JSONB)):
         return value
     elif isinstance(type_, DateTime):
         return make_datetime(value)
--- a/cmscontrib/DumpImporter.py
+++ b/cmscontrib/DumpImporter.py
@@ -111,7 +111,7 @@ def decode_value(type_, value):
     elif isinstance(type_, (Boolean, Integer, Float, Unicode, Enum, JSONB)):
         return value
     elif isinstance(type_, String):
-        return value.encode('latin1')
+        return value
     elif isinstance(type_, DateTime):
         return make_datetime(value)
     elif isinstance(type_, Interval):
--- a/cmscontrib/updaters/update_11.py
+++ b/cmscontrib/updaters/update_11.py
@@ -32,6 +32,7 @@ from __future__ import print_function
 from __future__ import unicode_literals
 from future.builtins.disabled import *
 from future.builtins import *
+from six import iteritems
 
 
 class Updater(object):
--- a/cmstestsuite/unit_tests/cmscontrib/DumpExporterTest.py
+++ b/cmstestsuite/unit_tests/cmscontrib/DumpExporterTest.py
@@ -56,12 +56,17 @@ class TestDumpExporter(TestCaseWithDatabase):
         # Add a file to be used as a statement.
         self.st_content = b"statement"
         self.st_digest = "bbb930cc426507ed3f6b7c343c75dd0e041494b7"
-        self.add_file(self.st_digest, self.st_content)
+        self.add_fsobject(self.st_digest, self.st_content)
+
+        # Add a file to be used as a submission source.
+        self.file_content = b"source"
+        self.file_digest = "828d338a9b04221c9cbe286f50cd389f68de4ecf"
+        self.add_fsobject(self.file_digest, self.file_content)
 
         # Add a file to be used as an executable.
         self.exe_content = b"executable"
         self.exe_digest = "0d5602f9eacdb79b54a675b2e91433cd13ed246e"
-        self.add_file(self.exe_digest, self.exe_content)
+        self.add_fsobject(self.exe_digest, self.exe_content)
 
         self.contest = self.add_contest(description="你好")
         self.participation = self.add_participation(contest=self.contest)
@@ -72,6 +77,8 @@ class TestDumpExporter(TestCaseWithDatabase):
         self.dataset = self.add_dataset(task=self.task)
         self.task.active_dataset = self.task.datasets[0]
         self.submission = self.add_submission(self.task, self.participation)
+        self.file = self.add_file(
+            submission=self.submission, digest=self.file_digest)
 
         # Add the executable to the submission
         self.submission_result = self.add_submission_result(
@@ -190,6 +197,7 @@ class TestDumpExporter(TestCaseWithDatabase):
 
         self.assertFileInDump(self.st_digest, self.st_content)
         self.assertFileInDump(self.exe_digest, self.exe_content)
+        self.assertFileInDump(self.file_digest, self.file_content)
 
         # Root objects are the contests, the users, and unattached tasks.
         assertCountEqual(self, self.dump["_objects"],
@@ -211,6 +219,7 @@ class TestDumpExporter(TestCaseWithDatabase):
 
         self.assertFileInDump(self.st_digest, self.st_content)
         self.assertFileInDump(self.exe_digest, self.exe_content)
+        self.assertFileInDump(self.file_digest, self.file_content)
 
     def test_export_single_contest_no_files(self):
         """Test exporting a contest does not export files of other contests."""
@@ -226,6 +235,7 @@ class TestDumpExporter(TestCaseWithDatabase):
 
         self.assertFileNotInDump(self.st_digest)
         self.assertFileNotInDump(self.exe_digest)
+        self.assertFileNotInDump(self.file_digest)
 
     def test_skip_files(self):
         """Test skipping files, generated or original."""
@@ -235,6 +245,7 @@ class TestDumpExporter(TestCaseWithDatabase):
         self.assertInDump(Executable, digest=self.exe_digest)
         self.assertFileNotInDump(self.st_digest)
         self.assertFileNotInDump(self.exe_digest)
+        self.assertFileNotInDump(self.file_digest)
 
     def test_skip_submissions(self):
         """Test skipping submissions.
@@ -251,6 +262,7 @@ class TestDumpExporter(TestCaseWithDatabase):
         self.assertNotInDump(Submission)
         self.assertNotInDump(SubmissionResult)
         self.assertFileNotInDump(self.exe_digest)
+        self.assertFileNotInDump(self.file_digest)
 
     def test_skip_generated(self):
         """Test skipping generated file.
@@ -263,6 +275,8 @@ class TestDumpExporter(TestCaseWithDatabase):
 
         self.assertInDump(Statement, digest=self.st_digest)
         self.assertFileInDump(self.st_digest, self.st_content)
+        self.assertInDump(Submission)
+        self.assertFileInDump(self.file_digest, self.file_content)
 
         self.assertNotInDump(SubmissionResult)
         self.assertFileNotInDump(self.exe_digest)
--- a/cmstestsuite/unit_tests/cmscontrib/DumpExporterTest.py
+++ b/cmstestsuite/unit_tests/cmscontrib/DumpExporterTest.py
@@ -128,7 +128,7 @@ class TestDumpExporter(TestCaseWithDatabase,
                     all(obj[k] == v for k, v in iteritems(kwargs)):
                 return key
         raise AssertionError("Cannot find object of class %s with fields %s" %
-                             (cls.__name__, str(kwargs)))
+                             (cls.__name__, kwargs))
 
     def assertNotInDump(self, cls, **kwargs):
         """Assert that an object is not in the dump.
@@ -144,7 +144,7 @@ class TestDumpExporter(TestCaseWithDatabase,
                     all(obj[k] == v for k, v in iteritems(kwargs)):
                 raise AssertionError("Object of class %s with fields %s "
                                      "should not appear in the dump" %
-                                     (cls.__name__, str(kwargs)))
+                                     (cls.__name__, kwargs))
 
     def assertFileInDump(self, digest, content):
         path = os.path.join(self.base, "files", digest)
--- a/cmstestsuite/unit_tests/cmscontrib/DumpExporterTest.py
+++ b/cmstestsuite/unit_tests/cmscontrib/DumpExporterTest.py
@@ -96,7 +96,7 @@ class TestDumpExporter(TestCaseWithDatabase):
 
     def tearDown(self):
         self.delete_data()
-        shutil.rmtree(self.base)
+        shutil.rmtree(os.path.dirname(self.base))
         super(TestDumpExporter, self).tearDown()
 
     def do_export(self, contest_ids, dump_files=True, skip_generated=False,
--- a/cmstestsuite/unit_tests/cmscontrib/DumpExporterTest.py
+++ b/cmstestsuite/unit_tests/cmscontrib/DumpExporterTest.py
@@ -30,9 +30,8 @@ from six import assertCountEqual, iteritems, itervalues
 import json
 import io
 import os
+import tempfile
 import unittest
-
-from pyfakefs import fake_filesystem_unittest
 
 # Needs to be first to allow for monkey patching the DB connection string.
 from cmstestsuite.unit_tests.testdbgenerator import TestCaseWithDatabase
@@ -44,16 +43,14 @@ from cms.db import Contest, Executable, Participation, Statement, Submission, \
 from cmscontrib.DumpExporter import DumpExporter
 
 
-class TestDumpExporter(TestCaseWithDatabase,
-                       fake_filesystem_unittest.TestCase):
+class TestDumpExporter(TestCaseWithDatabase):
 
     def setUp(self):
         super(TestDumpExporter, self).setUp()
-        self.setUpPyfakefs()
         if not os.path.exists(config.temp_dir):
             os.makedirs(config.temp_dir)
 
-        self.base = "/tmp/target"
+        self.base = os.path.join(tempfile.mkdtemp(), "target")
         self.dump = None
 
         # Add a file to be used as a statement.
@@ -104,9 +101,10 @@ class TestDumpExporter(TestCaseWithDatabase,
             skip_generated=skip_generated,
             skip_submissions=skip_submissions,
             skip_user_tests=False).do_export()
+        dump_path = os.path.join(self.base, "contest.json")
         try:
-            self.dump = json.load(
-                io.open("/tmp/target/contest.json", "rt", encoding="utf-8"))
+            with io.open(dump_path, "rt", encoding="utf-8") as f:
+                self.dump = json.load(f)
         except Exception:
             self.dump = None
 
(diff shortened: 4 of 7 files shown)


lerks:
Reviewed 2 of 2 files at r1, 5 of 5 files at r2.
Review status: all files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---

*[a discussion](https://reviewable.io:443/reviews/cms-dev/cms/876#-L6_vZC363q1qJj9LzUt:-L6_vZC363q1qJj9LzUu:bjoj9p) (no related file):*
I don't understand the benefits of using pyfakefs, but I see many risks in it.

In this particular case, I feel that that one could remove all pyfakefs stuff, replace `self.base = "/tmp/target"` with `self.base = tempfile.mkdtemp()`, add a removal in the `tearDown` method and get the same results. Even performance-wise they should be similar: in normal conditions tempfile returns a path in `/tmp`, which is mounted as `tmpfs` and thus is in-memory.

The risks I see are the usual ones associated with monkey-patching, which already bit us when we did this for gevent. My understanding is that files opened at import-time (and, oh my, does a lot of stuff happen at import time!) live on the real filesystem, whereas all that is opened during runtime lives on the fake fs. Since I don't know by heart which is which, I believe this is already a problem. From where is the config loaded (real fs)? Where does the file cacher live (fake fs)? Where do the coverage files end (real fs)?

Plus there are my concerns with the fact that the implementation of pyfakefs could differ subtly from a real filesystem (those bugs, that would only occur in tests, would be a nightmare to debug!). And my concerns that third-party low-level libraries would not see the files we expect them to see (I can't name any off the top of my head, but that doesn't mean I'm not worried).

I add that my gut feeling is that if a module requires pyfakefs in order to be tested reasonably well, that module would benefit from being refactored to accept parameters specifying where to store files (not just from a point of view of testability, really from a code quality standpoint).

Please argue in favor of pyfakefs.

---

*[cmscontrib/DumpExporter.py, line 128 at r1](https://reviewable.io:443/reviews/cms-dev/cms/876#-L6Yn3QGbUmDgIKU4iwA:-L6Yn3QGbUmDgIKU4iwB:b1p7d83) ([raw file](https://github.com/cms-dev/cms/blob/cf4cdf0039ba45d21058790cf19cf8516a55adb3/cmscontrib/DumpExporter.py#L128)):*
> ```Python
>         return value
>     elif isinstance(type_, String):
>         return value.decode('latin1')
> ```

My understanding was that we needed this because some of our fields were binary (i.e., str in py2 but bytes in py3), which is a type that cannot be encoded in JSON. We hence transformed them to unicode by decoding them as latin1 (which has the property that any string is valid). [Alternatives would have been base64 or hex, but they wouldn't have been as efficient].

Does this change mean that I was wrong and that we don't have any binary field? In that case, though, please add `String` to the tuple in the previous case of the if. And wouldn't it make sense to change the fields in cms.db to use only one of `String` or `Unicode`, rather than mix them, since they are the same?

---

*[cmstestsuite/unit_tests/testdbgenerator.py, line 178 at r2](https://reviewable.io:443/reviews/cms-dev/cms/876#-L6YpKifWmg9MSYj39q3:-L6YpKifWmg9MSYj39q4:bvok5vn) ([raw file](https://github.com/cms-dev/cms/blob/18e28403e3190f72a3ef4803f44ae3a500eaa1c2/cmstestsuite/unit_tests/testdbgenerator.py#L178)):*
> ```Python
>         return task
> 
>     def add_statement(self, task=None, **kwargs):
> ```

For other classes we have both a `get_` and an `add_` method, consider doing the same here.

---

*[cmstestsuite/unit_tests/testdbgenerator.py, line 268 at r2](https://reviewable.io:443/reviews/cms-dev/cms/876#-L6YpXcleOm_qDw2BsUS:-L6YpXcleOm_qDw2BsUT:buo9nmt) ([raw file](https://github.com/cms-dev/cms/blob/18e28403e3190f72a3ef4803f44ae3a500eaa1c2/cmstestsuite/unit_tests/testdbgenerator.py#L268)):*
> ```Python
>         return submission_result
> 
>     def add_executable(self, submission_result=None, **kwargs):
> ```

Same here

---

*[cmstestsuite/unit_tests/cmscontrib/DumpExporterTest.py, line 77 at r2](https://reviewable.io:443/reviews/cms-dev/cms/876#-L6YtxqDd8E6GmVZw8Jz:-L6YtxqDd8E6GmVZw8K-:bd1w3bx) ([raw file](https://github.com/cms-dev/cms/blob/18e28403e3190f72a3ef4803f44ae3a500eaa1c2/cmstestsuite/unit_tests/cmscontrib/DumpExporterTest.py#L77)):*
> ```Python
>         self.dataset = self.add_dataset(task=self.task)
>         self.task.active_dataset = self.task.datasets[0]
>         self.submission = self.add_submission(self.task, self.participation)
> ```

The submission has no files attached to it, correct? Maybe we would like it to have some, to make sure they are not dropped?

---

*[cmstestsuite/unit_tests/cmscontrib/DumpExporterTest.py, line 131 at r2](https://reviewable.io:443/reviews/cms-dev/cms/876#-L6YumSUjnkowT9U-1XA:-L6YumSUjnkowT9U-1XB:b-9ii4eb) ([raw file](https://github.com/cms-dev/cms/blob/18e28403e3190f72a3ef4803f44ae3a500eaa1c2/cmstestsuite/unit_tests/cmscontrib/DumpExporterTest.py#L131)):*
> ```Python
>                 return key
>         raise AssertionError("Cannot find object of class %s with fields %s" %
>                              (cls.__name__, str(kwargs)))
> ```

Isn't `str` here redundant?

---

*[cmstestsuite/unit_tests/cmscontrib/DumpExporterTest.py, line 147 at r2](https://reviewable.io:443/reviews/cms-dev/cms/876#-L6YuosGg5kzT9L6_aSB:-L6YuosGg5kzT9L6_aSC:buo9nmt) ([raw file](https://github.com/cms-dev/cms/blob/18e28403e3190f72a3ef4803f44ae3a500eaa1c2/cmstestsuite/unit_tests/cmscontrib/DumpExporterTest.py#L147)):*
> ```Python
>                 raise AssertionError("Object of class %s with fields %s "
>                                      "should not appear in the dump" %
>                                      (cls.__name__, str(kwargs)))
> ```

Same here

---

*[cmstestsuite/unit_tests/cmscontrib/DumpExporterTest.py, line 159 at r2](https://reviewable.io:443/reviews/cms-dev/cms/876#-L6_nJ0LSH_yEIcObqCL:-L6_nJ0LSH_yEIcObqCM:b4iep4e) ([raw file](https://github.com/cms-dev/cms/blob/18e28403e3190f72a3ef4803f44ae3a500eaa1c2/cmstestsuite/unit_tests/cmscontrib/DumpExporterTest.py#L159)):*
> ```Python
>         self.assertFalse(os.path.exists(path))
> 
>     def test_dont_overwrite(self):
> ```

`doesnt`? or maybe `tests`?

---

*[cmstestsuite/unit_tests/cmscontrib/DumpImporterTest.py, line 56 at r2](https://reviewable.io:443/reviews/cms-dev/cms/876#-L6_pZFRpRmKZ9L46fK_:-L6_pZFRpRmKZ9L46fKa:b-a0vig2) ([raw file](https://github.com/cms-dev/cms/blob/18e28403e3190f72a3ef4803f44ae3a500eaa1c2/cmstestsuite/unit_tests/cmscontrib/DumpImporterTest.py#L56)):*
> ```Python
>             "tasks": ["task_key"],
>         },
>         "task_key": {
> ```

Doesn't this need a `contest` key?

---

*[cmstestsuite/unit_tests/cmscontrib/DumpImporterTest.py, line 162 at r2](https://reviewable.io:443/reviews/cms-dev/cms/876#-L6_qoHdgbYdz4cagMdP:-L6_qoHdgbYdz4cagMdQ:bs8nxpq) ([raw file](https://github.com/cms-dev/cms/blob/18e28403e3190f72a3ef4803f44ae3a500eaa1c2/cmstestsuite/unit_tests/cmscontrib/DumpImporterTest.py#L162)):*
> ```Python
>         os.makedirs(d_path)
>         for digest, (desc, content) in iteritems(data):
>             with io.open(os.path.join(d_path, digest), "wt", encoding="utf-8")\
> ```

Nit: could you wrap inside the `open` call?

---

*[cmstestsuite/unit_tests/cmscontrib/DumpImporterTest.py, line 175 at r2](https://reviewable.io:443/reviews/cms-dev/cms/876#-L6_r76RMdxEAAHKXOZi:-L6_r76RMdxEAAHKXOZj:b-jc9y92) ([raw file](https://github.com/cms-dev/cms/blob/18e28403e3190f72a3ef4803f44ae3a500eaa1c2/cmstestsuite/unit_tests/cmscontrib/DumpImporterTest.py#L175)):*
> ```Python
>         a brand new session.
> 
>         From contest_id on, parameters are checked only if not None.
> ```

What's `contest_id`?

---

*[cmstestsuite/unit_tests/cmscontrib/DumpImporterTest.py, line 178 at r2](https://reviewable.io:443/reviews/cms-dev/cms/876#-L6_rD5o96_nhlQ3Ny0k:-L6_rD5o96_nhlQ3Ny0l:b-vbffmr) ([raw file](https://github.com/cms-dev/cms/blob/18e28403e3190f72a3ef4803f44ae3a500eaa1c2/cmstestsuite/unit_tests/cmscontrib/DumpImporterTest.py#L178)):*
> ```Python
> 
>         """
>         db_contests = self.session.query(Contest) \
> ```

Nit: no space before `\`?

---

*[cmstestsuite/unit_tests/cmscontrib/DumpImporterTest.py, line 192 at r2](https://reviewable.io:443/reviews/cms-dev/cms/876#-L6_rXtWLW0xCRl1tOAa:-L6_rXtWLW0xCRl1tOAb:buo9nmt) ([raw file](https://github.com/cms-dev/cms/blob/18e28403e3190f72a3ef4803f44ae3a500eaa1c2/cmstestsuite/unit_tests/cmscontrib/DumpImporterTest.py#L192)):*
> ```Python
>     def assertContestNotInDb(self, name):
>         """Assert that the contest with the given name is not in the DB."""
>         db_contests = self.session.query(Contest) \
> ```

Same here

---

*[cmstestsuite/unit_tests/cmscontrib/DumpImporterTest.py, line 245 at r2](https://reviewable.io:443/reviews/cms-dev/cms/876#-L6_s9IcuQ8HUwmQ9LCz:-L6_s9IcuQ8HUwmQ9LD-:b-tq5twa) ([raw file](https://github.com/cms-dev/cms/blob/18e28403e3190f72a3ef4803f44ae3a500eaa1c2/cmstestsuite/unit_tests/cmscontrib/DumpImporterTest.py#L245)):*
> ```Python
>             "040f06fd774092478d450774f5ba30c5da78acc8", "desc", b"content")
> 
>     def test_import_skip_generated(self):
> ```

This is essentially testing whether it's skipping files. Can it be made more precise?

---

*[cmstestsuite/unit_tests/cmscontrib/DumpImporterTest.py, line 280 at r2](https://reviewable.io:443/reviews/cms-dev/cms/876#-L6_sTJSMlRG3wLYYkzv:-L6_sTJSMlRG3wLYYkzw:b-5edi0d) ([raw file](https://github.com/cms-dev/cms/blob/18e28403e3190f72a3ef4803f44ae3a500eaa1c2/cmstestsuite/unit_tests/cmscontrib/DumpImporterTest.py#L280)):*
> ```Python
>                 "tasks": ["task_key"],
>             },
>             "task_key": {
> ```

Again missing `contest` key?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/876)*

stefano-maggiolo:
Review status: 1 of 6 files reviewed at latest revision, 15 unresolved discussions.

---

*[a discussion](https://reviewable.io:443/reviews/cms-dev/cms/876#-L6_vZC363q1qJj9LzUt:-L6aV_jFQyx6Tyis-iwV:budcktd) (no related file):*

I don't understand the benefits of using pyfakefs, but I see many risks in it.

In this particular case, I feel that that one could remove all pyfakefs stuff, replace `self.base = "/tmp/target"` with `self.base = tempfile.mkdtemp()`, add a removal in the `tearDown` method and get the same results. Even performance-wise they should be similar: in normal conditions tempfile returns a path in `/tmp`, which is mounted as `tmpfs` and thus is in-memory.

The risks I see are the usual ones associated with monkey-patching, which already bit us when we did this for gevent. My understanding is that files opened at import-time (and, oh my, does a lot of stuff happen at import time!) live on the real filesystem, whereas all that is opened during runtime lives on the fake fs. Since I don't know by heart which is which, I believe this is already a problem. From where is the config loaded (real fs)? Where does the file cacher live (fake fs)? Where do the coverage files end (real fs)?

Plus there are my concerns with the fact that the implementation of pyfakefs could differ subtly from a real filesystem (those bugs, that would only occur in tests, would be a nightmare to debug!). And my concerns that third-party low-level libraries would not see the files we expect them to see (I can't name any off the top of my head, but that doesn't mean I'm not worried).

I add that my gut feeling is that if a module requires pyfakefs in order to be tested reasonably well, that module would benefit from being refactored to accept parameters specifying where to store files (not just from a point of view of testability, really from a code quality standpoint).

Please argue in favor of pyfakefs.
</blockquote></details>

No, I won't argue because it's not something that deserve this time. I will award your essay with a removal of pyfakefs.

---

*[cmscontrib/DumpExporter.py, line 128 at r1](https://reviewable.io:443/reviews/cms-dev/cms/876#-L6Yn3QGbUmDgIKU4iwA:-L6aVm4Rrmk699QUJ0Cn:b-t2fn91) ([raw file](https://github.com/cms-dev/cms/blob/cf4cdf0039ba45d21058790cf19cf8516a55adb3/cmscontrib/DumpExporter.py#L128)):*

My understanding was that we needed this because some of our fields were binary (i.e., str in py2 but bytes in py3), which is a type that cannot be encoded in JSON. We hence transformed them to unicode by decoding them as latin1 (which has the property that any string is valid). [Alternatives would have been base64 or hex, but they wouldn't have been as efficient].

Does this change mean that I was wrong and that we don't have any binary field? In that case, though, please add `String` to the tuple in the previous case of the if. And wouldn't it make sense to change the fields in cms.db to use only one of `String` or `Unicode`, rather than mix them, since they are the same?
</blockquote></details>

Did some digging in SQLAlchemy's source code (https://github.com/zzzeek/sqlalchemy/blob/master/lib/sqlalchemy/sql/sqltypes.py). Unicode is derived from String, is just a String with `convert_unicode=True` and `_warn_on_bytestring=True`. 

I didn't get how it computes the returned value, but the property `python_type` returns `str` even when `convert_unicode` is false. So my conclusion, which could probably have been inferred from the documentation alone, is that String is still a string type, but left in the default encoding of the DB, so in our case is probably equivalent to Unicode. For binary data we should use... LargeBinary?! :) 

I'll leave fields as they are now, since it doesn't really matter. I'll move this to the if above.

---

*[cmstestsuite/unit_tests/testdbgenerator.py, line 178 at r2](https://reviewable.io:443/reviews/cms-dev/cms/876#-L6YpKifWmg9MSYj39q3:-L6bQfQrCf01E_S2KBcY:b-hvinyo) ([raw file](https://github.com/cms-dev/cms/blob/18e28403e3190f72a3ef4803f44ae3a500eaa1c2/cmstestsuite/unit_tests/testdbgenerator.py#L178)):*

For other classes we have both a `get_` and an `add_` method, consider doing the same here.
</blockquote></details>

We can do it when we need it. Not building a generic library here.

---

*[cmstestsuite/unit_tests/testdbgenerator.py, line 268 at r2](https://reviewable.io:443/reviews/cms-dev/cms/876#-L6YpXcleOm_qDw2BsUS:-L6bmE3STWalgewFBduv:bb73xfv) ([raw file](https://github.com/cms-dev/cms/blob/18e28403e3190f72a3ef4803f44ae3a500eaa1c2/cmstestsuite/unit_tests/testdbgenerator.py#L268)):*

Same here
</blockquote></details>

Same

---

*[cmstestsuite/unit_tests/cmscontrib/DumpExporterTest.py, line 77 at r2](https://reviewable.io:443/reviews/cms-dev/cms/876#-L6YtxqDd8E6GmVZw8Jz:-L6bqa5VxMpb86aCJSu3:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/18e28403e3190f72a3ef4803f44ae3a500eaa1c2/cmstestsuite/unit_tests/cmscontrib/DumpExporterTest.py#L77)):*

The submission has no files attached to it, correct? Maybe we would like it to have some, to make sure they are not dropped?
</blockquote></details>

Done.

---

*[cmstestsuite/unit_tests/cmscontrib/DumpExporterTest.py, line 131 at r2](https://reviewable.io:443/reviews/cms-dev/cms/876#-L6YumSUjnkowT9U-1XA:-L6bQnqykOCmzxh6AqTW:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/18e28403e3190f72a3ef4803f44ae3a500eaa1c2/cmstestsuite/unit_tests/cmscontrib/DumpExporterTest.py#L131)):*

Isn't `str` here redundant?
</blockquote></details>

Done.

---

*[cmstestsuite/unit_tests/cmscontrib/DumpExporterTest.py, line 147 at r2](https://reviewable.io:443/reviews/cms-dev/cms/876#-L6YuosGg5kzT9L6_aSB:-L6bQoIp3aXZHOgHHIRV:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/18e28403e3190f72a3ef4803f44ae3a500eaa1c2/cmstestsuite/unit_tests/cmscontrib/DumpExporterTest.py#L147)):*

Same here
</blockquote></details>

Done.

---

*[cmstestsuite/unit_tests/cmscontrib/DumpExporterTest.py, line 159 at r2](https://reviewable.io:443/reviews/cms-dev/cms/876#-L6_nJ0LSH_yEIcObqCL:-L6bQy4PEefiuu-aRHj2:bo0til4) ([raw file](https://github.com/cms-dev/cms/blob/18e28403e3190f72a3ef4803f44ae3a500eaa1c2/cmstestsuite/unit_tests/cmscontrib/DumpExporterTest.py#L159)):*

`doesnt`? or maybe `tests`?
</blockquote></details>

Or maybe assume it's imperative, like the others.

---

*[cmstestsuite/unit_tests/cmscontrib/DumpImporterTest.py, line 56 at r2](https://reviewable.io:443/reviews/cms-dev/cms/876#-L6_pZFRpRmKZ9L46fK_:-L6bS3i376U-MnHIS4r1:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/18e28403e3190f72a3ef4803f44ae3a500eaa1c2/cmstestsuite/unit_tests/cmscontrib/DumpImporterTest.py#L56)):*

Doesn't this need a `contest` key?
</blockquote></details>

Done.

---

*[cmstestsuite/unit_tests/cmscontrib/DumpImporterTest.py, line 162 at r2](https://reviewable.io:443/reviews/cms-dev/cms/876#-L6_qoHdgbYdz4cagMdP:-L6bRDpnoGzQ6jPYqUU9:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/18e28403e3190f72a3ef4803f44ae3a500eaa1c2/cmstestsuite/unit_tests/cmscontrib/DumpImporterTest.py#L162)):*

Nit: could you wrap inside the `open` call?
</blockquote></details>

Done.

---

*[cmstestsuite/unit_tests/cmscontrib/DumpImporterTest.py, line 175 at r2](https://reviewable.io:443/reviews/cms-dev/cms/876#-L6_r76RMdxEAAHKXOZi:-L6bRM9O8vIJNPHdwzQ9:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/18e28403e3190f72a3ef4803f44ae3a500eaa1c2/cmstestsuite/unit_tests/cmscontrib/DumpImporterTest.py#L175)):*

What's `contest_id`?
</blockquote></details>

Done.

---

*[cmstestsuite/unit_tests/cmscontrib/DumpImporterTest.py, line 178 at r2](https://reviewable.io:443/reviews/cms-dev/cms/876#-L6_rD5o96_nhlQ3Ny0k:-L6bRQPu7wS065yXFhWE:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/18e28403e3190f72a3ef4803f44ae3a500eaa1c2/cmstestsuite/unit_tests/cmscontrib/DumpImporterTest.py#L178)):*

Nit: no space before `\`?
</blockquote></details>

Done.

---

*[cmstestsuite/unit_tests/cmscontrib/DumpImporterTest.py, line 192 at r2](https://reviewable.io:443/reviews/cms-dev/cms/876#-L6_rXtWLW0xCRl1tOAa:-L6bRR1j1j2jm-8TRL6T:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/18e28403e3190f72a3ef4803f44ae3a500eaa1c2/cmstestsuite/unit_tests/cmscontrib/DumpImporterTest.py#L192)):*

Same here
</blockquote></details>

Done.

---

*[cmstestsuite/unit_tests/cmscontrib/DumpImporterTest.py, line 245 at r2](https://reviewable.io:443/reviews/cms-dev/cms/876#-L6_s9IcuQ8HUwmQ9LCz:-L6btEp9i4_VJRqq4pOk:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/18e28403e3190f72a3ef4803f44ae3a500eaa1c2/cmstestsuite/unit_tests/cmscontrib/DumpImporterTest.py#L245)):*

This is essentially testing whether it's skipping files. Can it be made more precise?
</blockquote></details>

Done.

---

*[cmstestsuite/unit_tests/cmscontrib/DumpImporterTest.py, line 280 at r2](https://reviewable.io:443/reviews/cms-dev/cms/876#-L6_sTJSMlRG3wLYYkzv:-L6bS26LrG8C2JlGwPZC:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/18e28403e3190f72a3ef4803f44ae3a500eaa1c2/cmstestsuite/unit_tests/cmscontrib/DumpImporterTest.py#L280)):*

Again missing `contest` key?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/876)*

lerks:
Reviewed 2 of 2 files at r3, 2 of 2 files at r4, 3 of 3 files at r5, 3 of 3 files at r6.

---

*[cmscontrib/DumpExporter.py, line 128 at r1](https://reviewable.io:443/reviews/cms-dev/cms/876#-L6Yn3QGbUmDgIKU4iwA:-L6cE88Ya-B6DWa1Ww4N:b-gbl3gk) ([raw file](https://github.com/cms-dev/cms/blob/cf4cdf0039ba45d21058790cf19cf8516a55adb3/cmscontrib/DumpExporter.py#L128)):*

Did some digging in SQLAlchemy's source code (https://github.com/zzzeek/sqlalchemy/blob/master/lib/sqlalchemy/sql/sqltypes.py). Unicode is derived from String, is just a String with `convert_unicode=True` and `_warn_on_bytestring=True`. 

I didn't get how it computes the returned value, but the property `python_type` returns `str` even when `convert_unicode` is false. So my conclusion, which could probably have been inferred from the documentation alone, is that String is still a string type, but left in the default encoding of the DB, so in our case is probably equivalent to Unicode. For binary data we should use... LargeBinary?! :) 

I'll leave fields as they are now, since it doesn't really matter. I'll move this to the if above.

</blockquote></details>

I see. So the reason this was failing in py3 is that we were calling `decode` on a `str`?

---

*[cmstestsuite/unit_tests/cmscontrib/DumpExporterTest.py, line 159 at r2](https://reviewable.io:443/reviews/cms-dev/cms/876#-L6_nJ0LSH_yEIcObqCL:-L6cEuuplwvzqMLmI13M:bsby1ic) ([raw file](https://github.com/cms-dev/cms/blob/18e28403e3190f72a3ef4803f44ae3a500eaa1c2/cmstestsuite/unit_tests/cmscontrib/DumpExporterTest.py#L159)):*

Or maybe assume it's imperative, like the others.
</blockquote></details>

Ooh... :)

---

*[cmstestsuite/unit_tests/cmscontrib/DumpExporterTest.py, line 53 at r5](https://reviewable.io:443/reviews/cms-dev/cms/876#-L6cF10X0Fq9Cnsd9hc5:-L6cF10YA6mYs99nC40Z:bgn4mbr) ([raw file](https://github.com/cms-dev/cms/blob/14083aedbf322d6b1181b2a35cfc6259cb1318b5/cmstestsuite/unit_tests/cmscontrib/DumpExporterTest.py#L53)):*
> ```Python
>             os.makedirs(config.temp_dir)
> 
>         self.base = os.path.join(tempfile.mkdtemp(), "target")
> ```

Shouldn't you remove it in the `tearDown` / with an `addCleanUp`?

---

*[cmstestsuite/unit_tests/cmscontrib/DumpImporterTest.py, line 127 at r6](https://reviewable.io:443/reviews/cms-dev/cms/876#-L6cGjPFhLYd-yft-GRo:-L6cGjPFhLYd-yft-GRp:buo9nmt) ([raw file](https://github.com/cms-dev/cms/blob/0c78f93f3ad53a9857230e3fce88da7dc9bd7489/cmstestsuite/unit_tests/cmscontrib/DumpImporterTest.py#L127)):*
> ```Python
>             os.makedirs(config.temp_dir)
> 
>         self.base = tempfile.mkdtemp()
> ```

Same here

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/876)*

stefano-maggiolo:
Review status: 3 of 6 files reviewed at latest revision, 2 unresolved discussions.

---

*[cmscontrib/DumpExporter.py, line 128 at r1](https://reviewable.io:443/reviews/cms-dev/cms/876#-L6Yn3QGbUmDgIKU4iwA:-L6cZ99p0VT-CPY3eyw_:b7u03hh) ([raw file](https://github.com/cms-dev/cms/blob/cf4cdf0039ba45d21058790cf19cf8516a55adb3/cmscontrib/DumpExporter.py#L128)):*

I see. So the reason this was failing in py3 is that we were calling `decode` on a `str`?
</blockquote></details>

Correct

---

*[cmstestsuite/unit_tests/cmscontrib/DumpExporterTest.py, line 53 at r5](https://reviewable.io:443/reviews/cms-dev/cms/876#-L6cF10X0Fq9Cnsd9hc5:-L6cZgOrg4qFdU-07Wta:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/14083aedbf322d6b1181b2a35cfc6259cb1318b5/cmstestsuite/unit_tests/cmscontrib/DumpExporterTest.py#L53)):*

Shouldn't you remove it in the `tearDown` / with an `addCleanUp`?
</blockquote></details>

Done.

---

*[cmstestsuite/unit_tests/cmscontrib/DumpImporterTest.py, line 127 at r6](https://reviewable.io:443/reviews/cms-dev/cms/876#-L6cGjPFhLYd-yft-GRo:-L6cZhFJGsRl_A3EkFK4:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/0c78f93f3ad53a9857230e3fce88da7dc9bd7489/cmstestsuite/unit_tests/cmscontrib/DumpImporterTest.py#L127)):*

Same here
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/876)*

lerks:
:lgtm:

---

Reviewed 1 of 1 files at r7, 2 of 2 files at r8, 3 of 3 files at r9.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/876#-:-L6coOj9CeobxPAiZBXP:bnfp4nl)*

stefano-maggiolo:
Merged

Issue #848, opened 2018-02-21
DumpImporter fails on 32-bit systems because timestamps are too large


lerks:
The contests we create for testing have an end date in the year 2100. When this is exported, the end date is stored as a floating-point UNIX timestamp (number of seconds since 1970-01-01 00:00:00 UTC) whose integral part doesn't fit into a 32-bit signed integer. Upon reimport the datetime library fails to parse these timestamps back into a datetime object.

I'm not sure how this can be fixed. Either we provide our own converter that works on 32-bit systems (it will be slower though). Or, for a start, we avoid using dates so far into the future. Or we declare CMS incompatible with 32-bit systems.

stefano-maggiolo:
FWIW we don't use 2100 anymore. I think we only used it when importing a contest with --test, which I sneakily deleted recently.

wil93:
> FWIW we don't use 2100 anymore

There are some occurrences here and there

https://github.com/cms-dev/cms/blob/a369a245f992e144072b6b60dd233086be288c83/cms/db/contest.py#L193L215

stefano-maggiolo:
Oh you're right, my git-grep-fu was bad. I'll remove them.

stefano-maggiolo:
Fixed

PR #877, opened 2018-03-01
Fix the trouble with the dates past Y2k38

stefano-maggiolo:
Only last commit is useful, but injecting the other two to have tests.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/877)

### DIFF
--- a/cms/db/contest.py
+++ b/cms/db/contest.py
@@ -200,7 +200,7 @@ class Contest(Base):
     stop = Column(
         DateTime,
         nullable=False,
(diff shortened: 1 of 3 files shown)


stefano-maggiolo:
Fixes #848

lerks:
If you didn't include the two commits of #876 in here too you would have a LGTM by now :P

stefano-maggiolo:
Rebasing to master to see reports...

stefano-maggiolo:
Merged.

PR #879, opened 2018-03-02
Codecov integration

stefano-maggiolo:
Some related improvements in the other commits

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/879)


wil93:
neat

stefano-maggiolo:
Also, depressing, but that's the point.

lerks:
Re first commit: I disagree that you renamed it to the format used by all others. The format, I guess, is appending `Test` if the name is `CamelCase`, and appending either `_test` or `test` if it's `underscore_separated`. You seemed to choose the latter but forgot to rename all the tests the root, in `db` and in `io`. I would actually prefer to use the former (and rename all the files in `service`.

---

Reviewed 3 of 3 files at r1, 4 of 4 files at r2, 4 of 4 files at r3.

---

*[cmstestsuite/__init__.py, line 117 at r3](https://reviewable.io:443/reviews/cms-dev/cms/879#-L6cBwwwBAmQhN7fBvYq:-L6cBwwxUEU_Xjyd23-X:b2nifei) ([raw file](https://github.com/cms-dev/cms/blob/93f90201c7f8ddf6154ecbd36e7429e29e7ee2be/cmstestsuite/__init__.py#L117)):*
> ```Python
>         logger.info("Sending coverage results to codecov for flag %s." % flag)
>         subprocess.call(
>             "bash -c 'bash <(curl -s https://codecov.io/bash) -c -F %s'" %
> ```

wow

---

*[cmstestsuite/RunFunctionalTests.py, line 175 at r3](https://reviewable.io:443/reviews/cms-dev/cms/879#-L6cCdpjlrk9xf4jeNbg:-L6cCdpjlrk9xf4jeNbh:bcbio40) ([raw file](https://github.com/cms-dev/cms/blob/93f90201c7f8ddf6154ecbd36e7429e29e7ee2be/cmstestsuite/RunFunctionalTests.py#L175)):*
> ```Python
>         "-v", "--verbose", action="count", default=0,
>         help="print debug information (use multiple times for more)")
>     parser.add_argument(
> ```

Is there any scenario you imagine where it would make sense for someone to call this for a run on their local machine? If not it may make sense to add the codecov reporting function to be called "externally", in the Travis config.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/879#-:-L6cA-f6yEcPNETccdwI:b9xh5qj)*

stefano-maggiolo:
Discussed offline - the files in the root are not tests but utilities. Decided to keep *Test for camelcase and *_test for not, and use init_test for tests of __init__.

---

---

*[cmstestsuite/__init__.py, line 117 at r3](https://reviewable.io:443/reviews/cms-dev/cms/879#-L6cBwwwBAmQhN7fBvYq:-L6ccya9acpZ823W7M-A:bwbwp7l) ([raw file](https://github.com/cms-dev/cms/blob/93f90201c7f8ddf6154ecbd36e7429e29e7ee2be/cmstestsuite/__init__.py#L117)):*

wow
</blockquote></details>

Eh. The double bash or the curl? The curl there is no way around, if you have suggestions for the double bash, please.

---

*[cmstestsuite/RunFunctionalTests.py, line 175 at r3](https://reviewable.io:443/reviews/cms-dev/cms/879#-L6cCdpjlrk9xf4jeNbg:-L6cd5yzSCO4t5XG-UxZ:b976baz) ([raw file](https://github.com/cms-dev/cms/blob/93f90201c7f8ddf6154ecbd36e7429e29e7ee2be/cmstestsuite/RunFunctionalTests.py#L175)):*

Is there any scenario you imagine where it would make sense for someone to call this for a run on their local machine? If not it may make sense to add the codecov reporting function to be called "externally", in the Travis config.
</blockquote></details>

I think it's more complicated because either: 
- you must call Unit and Functional separately
- you must have them generate coverage in different files, but being aware that each is already a combination of many independent coverage files, so it's a mess.

Also, if this is useless in a local machine you just have to restrain yourself from adding it to your command line :P (it would fail anyway because you'd need to provide a token to identify the build).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/879#-:-L6c_UIIPpcnwsbs6Qz2:b-dtrg5e)*

lerks:
Just rename `localetest` to `locale_test`.

---

Reviewed 11 of 11 files at r4, 1 of 1 files at r5.

---

*[cmstestsuite/__init__.py, line 117 at r3](https://reviewable.io:443/reviews/cms-dev/cms/879#-L6cBwwwBAmQhN7fBvYq:-L6cBtY0ihiYme8EciX6:b-1hbsaf) ([raw file](https://github.com/cms-dev/cms/blob/93f90201c7f8ddf6154ecbd36e7429e29e7ee2be/cmstestsuite/__init__.py#L117)):*

Eh. The double bash or the curl? The curl there is no way around, if you have suggestions for the double bash, please.
</blockquote></details>

The whole thing. But we discussed this in private and I'm ok with keeping it.

---

*[cmstestsuite/RunFunctionalTests.py, line 175 at r3](https://reviewable.io:443/reviews/cms-dev/cms/879#-L6cCdpjlrk9xf4jeNbg:-L6cpHGlvLZB9fBpNqT8:b-4tsole) ([raw file](https://github.com/cms-dev/cms/blob/93f90201c7f8ddf6154ecbd36e7429e29e7ee2be/cmstestsuite/RunFunctionalTests.py#L175)):*

I think it's more complicated because either: 
- you must call Unit and Functional separately
- you must have them generate coverage in different files, but being aware that each is already a combination of many independent coverage files, so it's a mess.

Also, if this is useless in a local machine you just have to restrain yourself from adding it to your command line :P (it would fail anyway because you'd need to provide a token to identify the build).
</blockquote></details>

I was thinking of your first option (calling Unit and Functional tests separately, and calling the codecov magic in between) but I see the merits of this approach and I'm ok with keeping it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/879#-:-L6cpUACRKokVrXgknwR:b2kynip)*

lerks:
:lgtm:

---

Reviewed 1 of 4 files at r1, 2 of 2 files at r6.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/879#-:-L6cpwxPdi-bx5dfT3QU:bnfp4nl)*

stefano-maggiolo:
Merged.

PR #882, opened 2018-03-03
Add a WSGI middleware to serve files from the file cacher by digest

### LINKED ISSUE #870: Issue between FileHandler and AdminWebServer causes the whole file to be read in memory by AWS
To reproduce, import a task with a single 1GB testcase and then try to download said testcase from AWS.

With @bblackham I investigated this a bit, we tried using `.flush()` after `.write()` but it wasn't enough. We tried adding `Transfer-Encoding: chunked` but it appears it's better to not use that (it's automatically added)

lerks:
The Tornado WSGI adapter buffers the whole response before starting to
send it to the client. This causes unexpected spikes in memory usage and
could be a vector for a denial-of-service attack. The Tornado devs have
no intention of fixing this, as their WSGI support is sub-standard by
design.

The most reasonable solution was to implement file serving as a "native"
WSGI application and insert it as a middleware into our stack. The
Tornado handlers that want to serve files have to issue a redirect to
this application. Files are accessed by digest, which means that
potentially every user can access any file, without restrictions (no
autentication, no authorization, ...). Yet, given the large space
digests live in, the likeliness of this happening is so small that we
don't believe it to be a security concern. Users are also able to
specify the file name and the MIME type they want the response to be
served as, though these are validated.

The new WSGI middleware replaces the classes obtained by using
file_handler_gen (and, as such, reduces dynamic class generation magic
and improves static analysis) and also replaces a similar handler in
AWS. It supports conditional requests (based on the ETag header) and
partial requests (based on the Range header).

Unit tests are introduced too.

Fixes #870.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/882)

### DIFF
--- a/cms/io/web_service.py
+++ b/cms/io/web_service.py
@@ -38,6 +38,9 @@ from gevent.pywsgi import WSGIServer
 from werkzeug.wsgi import DispatcherMiddleware, SharedDataMiddleware
 from werkzeug.contrib.fixers import ProxyFix
 
+from cms.db.filecacher import FileCacher
+from cms.server.file_middleware import FileByDigestMiddleware
+
 from .service import Service
 from .web_rpc import RPCMiddleware
 
@@ -74,6 +77,10 @@ class WebService(Service):
                 self.wsgi_app, {"/static": entry},
                 cache=True, cache_timeout=SECONDS_IN_A_YEAR,
                 fallback_mimetype="application/octet-stream")
+
+        self.file_cacher = FileCacher(self)
+        self.wsgi_app = DispatcherMiddleware(
+            self.wsgi_app, {"/file": FileByDigestMiddleware(self.file_cacher)})
 
         if rpc_enabled:
             self.wsgi_app = DispatcherMiddleware(
--- a/cms/server/__init__.py
+++ b/cms/server/__init__.py
@@ -26,14 +26,11 @@ from future.builtins import *
 
 from .util import \
     CommonRequestHandler, actual_phase_required, compute_actual_phase, \
-    file_handler_gen, filter_ascii, \
-    create_url_builder, multi_contest
+    filter_ascii, create_url_builder, multi_contest
 
 
 __all__ = [
     # util
     "CommonRequestHandler", "actual_phase_required", "compute_actual_phase",
-    "file_handler_gen", "filter_ascii",
-    "create_url_builder",
-    "multi_contest",
+    "filter_ascii", "create_url_builder", "multi_contest",
 ]
--- a/cms/server/admin/handlers/__init__.py
+++ b/cms/server/admin/handlers/__init__.py
@@ -28,7 +28,6 @@ from future.builtins.disabled import *
 from future.builtins import *
 
 from .base import \
-    FileFromDigestHandler, \
     SimpleHandler, \
     SimpleContestHandler
 from .main import \
@@ -119,7 +118,6 @@ HANDLERS = [
     (r"/resources/([0-9]+|all)", ResourcesHandler),
     (r"/resources/([0-9]+|all)/([0-9]+)", ResourcesHandler),
     (r"/notifications", NotificationsHandler),
-    (r"/file/([a-f0-9]+)/([a-zA-Z0-9_.-]+)", FileFromDigestHandler),
 
     # Contest
 
--- a/cms/server/admin/handlers/base.py
+++ b/cms/server/admin/handlers/base.py
@@ -54,7 +54,7 @@ from cms.db import Admin, Contest, Participation, Question, \
     UserTest
 from cms.grading.scoretypes import get_score_type_class
 from cms.grading.tasktypes import get_task_type_class
-from cms.server import CommonRequestHandler, file_handler_gen
+from cms.server import CommonRequestHandler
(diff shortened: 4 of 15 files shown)


wil93:
But what if in some contest (I'm thinking about online rounds mainly) some admin leaks a URL in some way? It's true that hash values are very sparse but the contestant could get to know a URL in some way and then visit without the system requiring authentication..

lerks:
You're right. Though that possibility feels a bit remote, and all solutions we came up with to limit it felt too complicated. If you have some suggestion let's hear it.

stefano-maggiolo:
Thank you! I would be waaay more comfortable though with a setup like this:

- tornado check access, issues a token which could be HMAC(secret_key, timestamp||participation_id||digest);
- tornado redirects to file/digest/filename, passing also token, and timestamp;
- werkzeug receives the request, retrieves participation_id (this can be done easily?) , validates the token is correct, checks timestamp is not more than X minutes ago (5?), and if so gives the file

The two things I'm trying to avoid: sharing URLs, giving indefinite access to resources that might be protected later.

---

Review status: 0 of 15 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[cms/server/file_middleware.py, line 68 at r1](https://reviewable.io:443/reviews/cms-dev/cms/882#-L6fSpVmBW8ZRvhFS9IR:-L6fSpVmBW8ZRvhFS9IS:b-tl4ape) ([raw file](https://github.com/cms-dev/cms/blob/11231a3716b20e019d5ac402abbee06ecb858045/cms/server/file_middleware.py#L68)):*
> ```Python
>         """
>         self.file_cacher = file_cacher
>         self.url_map = Map([Rule("/<digest:digest>/<filename:filename>",
> ```

I would have passed the filename as an argument, rather than part of the url - any reason not to?

---

*[cms/server/admin/handlers/submission.py, line 97 at r1](https://reviewable.io:443/reviews/cms-dev/cms/882#-L6fTnuOTCGvo-JkctS-:-L6fTnuOTCGvo-JkctS0:bynwvoz) ([raw file](https://github.com/cms-dev/cms/blob/11231a3716b20e019d5ac402abbee06ecb858045/cms/server/admin/handlers/submission.py#L97)):*
> ```Python
> 
>         self.redirect(self.url("file", digest, real_filename,
>                                mimetype="text/plain"),
> ```

Should you let the middleware detect the mimetype? Could be a source file, after all.

---

*[cms/server/admin/handlers/submission.py, line 98 at r1](https://reviewable.io:443/reviews/cms-dev/cms/882#-L6fTdgjZJgbRx3_vgau:-L6fTdgjZJgbRx3_vgav:b5pqy11) ([raw file](https://github.com/cms-dev/cms/blob/11231a3716b20e019d5ac402abbee06ecb858045/cms/server/admin/handlers/submission.py#L98)):*
> ```Python
>         self.redirect(self.url("file", digest, real_filename,
>                                mimetype="text/plain"),
>                       permanent=False, status=303)
> ```

Method for this pattern in the base?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/882#-:-L6fWOsPjCOUtmWRSIAi:b-jymgsb)*

stefano-maggiolo:
Another situation similar to the one @wil93 describes is sharing of digests between contestants, which seems easier than sharing of files.

fagu:
Wouldn't this also allow direct communication between contestants? As far as I understand, it would be possible to check whether a specific file is present on the server. Say the contestants agree on a specific set of files F1,...,Fn before the contest. During the contest, every contestant repeatedly tries to download the files corresponding to the digests of F1,...,Fn.
If a contestant wants to broadcast a number 1<=i<=n, he uploads file Fi as a solution. From that point on, everybody's attempts to download the file with digest of Fi will succeed, so they know the number i.

For example, the contestants could agree that submitting the file with contents "A: segment tree" means that problem A can be solved using a segment tree. Or "B: easy" could mean that problem B is easy. They could repeatedly try to access the files with digest(A: segment tree), digest(A: bipartite matching), digest(A: 1), digest(B: 1), ..., until they know which algorithms to use on each problem.

Or someone could even use that strategy to transmit the digests of their solution files, "digit by digit". Then, anybody else can find and download the solution files.

lerks:
Yes, very good remark, I hadn't thought of that, thanks!

I think this definitely kills this PR in the currently proposed form. I was already working on implementing some type of access control (for example the one @stefano-maggiolo proposed). I'll submit a new PR when I'm done, I'm keeping this one open in the meantime as a reminder.

lerks:
I've had an idea that while a bit ugly should effectively solve all the issues reported here.

When some Tornado handler wants to send a file back to the user it instead sends a response with no body but with a custom header whose value is the digest of the file (say, X-CMS-File-Digest). Then we install a middleware around Tornado that, whenever it sees that a response is being served that contains such a header, it discards that request and replaces it with one that instead includes the contents of the file corresponding to that digest.

This will not expose an endpoint where users can get files by digest, thus the current access logic is kept (no redirect + second request, no tokens/cookies/query parameters, no second authentication, ...).

Does this solution seem convincing to all the people who weighted in in this discussion?

lerks:
This PR has been superseded by #984; discussion moves there.

PR #884, opened 2018-03-09
Fix single submission rejudge

akmohtashami:
This fixes rejudge of single submissions in AWS by preventing task_id to be automatically filled when submission_id is set. Otherwise an exception in raised here:

https://github.com/cms-dev/cms/blob/2b1d76b8f9ddd5bce06773f6255e10ec75b488d2/cms/service/__init__.py#L79

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/884)


lerks:
I fixed the issue reported by Codacy and pushed this as 0cb514e204abb5ef3374624955f3fe414aed35bf. Thanks!

PR #885, opened 2018-03-09
Fix error on importing testcases from zip file

akmohtashami:
Currently, uploading a zip file in AWS fails with the following error:
```
TypeError(u"Testcase.__init__() got a '<type 'str'>' for keyword argument 'codename', which requires a '<type 'unicode'>'",)
```
This fixes the problem, however I am not sure if it does it the right way.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/885)

### DIFF
--- a/cmscommon/importers.py
+++ b/cmscommon/importers.py
@@ -112,7 +112,10 @@ def import_testcases_from_zipfile(
                         output, "Testcase output for task %s" % task_name)
                 except Exception:
                     raise Exception("Testcase storage failed")
-
+                try:
+                    codename = codename.decode("utf-8")
+                except AttributeError:
+                    pass
                 testcase = Testcase(codename, public, input_digest,
                                     output_digest, dataset=dataset)
                 session.add(testcase)


lerks:
So, I investigated a bit and the issue seems to be that `zipfile.ZipFile.namelist()` returns (bytes) `str`s in Python 2 for filename that are ASCII-only (it does return `unicode` when they aren't, though). In Python 3 (unicode) `str`s are returned all the time. I think a smaller fix would be to just cast the filename to `str`, and to do so as soon as we get it (i.e., on line 62) so that all the code can assume it is in unicode form. Could you do that? Also please add a comment explaining it.

lerks:
This might need to be backported to v1.3.

lerks:
Fixed and pushed as 89548be300164b95c79554aa2339163aec460956.

Also, no need to backport to v1.3, `Base.__init__` was more lenient back then:
https://github.com/cms-dev/cms/blob/3ec2262a36bdc7c7a9bd58ac594e97f48004eabe/cms/db/base.py#L47
Now:
https://github.com/cms-dev/cms/blob/89548be300164b95c79554aa2339163aec460956/cms/db/base.py#L61

PR #899, opened 2018-03-17
Compilation command improvements

andreyv:
- C11/C++11:
  - Enable GNU extensions (default GCC mode).
- Pascal:
  - Reorder flags to follow (preprocessor, compiler, linker) order.
  - Enable symbol stripping.
- Java:
  - Replace `mv` calls with direct output names.
  - Use `/bin/sh` instead of `/bin/bash` for portability.
  - Drop unneeded full paths inside shell commands.
  - Define `eval` system property for evaluation, similar to `-DEVAL`. Can be read with `Boolean.getBoolean("eval")` in the code. Possible alternative name: `cms.eval`.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/899)

### DIFF
--- a/cms/grading/languages/c++11_g++.py
+++ b/cms/grading/languages/c++11_g++.py
@@ -65,7 +65,7 @@ class Cpp11Gpp(CompiledLanguage):
         command = ["/usr/bin/g++"]
         if for_evaluation:
             command += ["-DEVAL"]
-        command += ["-std=c++11", "-O2", "-pipe", "-static",
+        command += ["-std=gnu++11", "-O2", "-pipe", "-static",
                     "-s", "-o", executable_filename]
         command += source_filenames
         return [command]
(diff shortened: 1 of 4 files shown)


andreyv:
One more possible improvement: escape `executable_filename` in shell commands with `pipes.quote()`/`shlex.quote()`.

stefano-maggiolo:
Thanks, looks good. Couple of comments:
- can you split the first commit in 3 (C/C++, Java, Pascal) since they are pretty much unrelated?
- using quote applies to any language really, right? Am I not seeing a reason for doing it only in Java?

andreyv:
1. Done.
2. Only names that are passed to shell need to be escaped. So far there are only two such commands, both in this file.

stefano-maggiolo:
Right, makes sense. Thank you! :lgtm: 

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/899#-:-L7urBKj-GOsUwlZxXV_:beufe9y)*

stefano-maggiolo:
Merged

PR #886, opened 2018-03-11
Start porting AWS to Jinja2

lerks:
The is the first PR that ports AWS to Jinja2. It's along the same lines as the one that ported CWS, so even though it's many commits I hope it will go smoothly.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/886)

PR #900, opened 2018-03-17
Finish porting AWS to Jinja2

lerks:
See #886 for the first part.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/900)


cms/server/admin/handlers/base.py
```diff
@@ -305,6 +305,9 @@ def render_params(self):
         params["contest"] = self.contest
         params["url"] = self.url
         params["xsrf_form_html"] = self.xsrf_form_html()
+        # FIXME these are cheating
+        params["config"] = config
+        params["handler"] = self
```
stefano-maggiolo: Is it me or the commit title doesn't agree much with what the commit does?

lerks: The first line of the commit message is correct (config and handler are values that, under Tornado, were either automatically injected by Tornado or imported by templates). But, yes, the rest of the message has nothing to do with it and appears to be duplicated from the following commit. Not sure how that happened.

cms/server/util.py
```diff
@@ -375,29 +375,57 @@ def get_url_root(request_path):
         return "."
 
 
-def create_url_builder(url_root):
-    """Return a function that builds an URL relative to the given root.
+class Url(object):
```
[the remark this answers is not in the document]
lerks: Done

stefano-maggiolo:
LGTM

lerks:
All pushed to master!

PR #905, opened 2018-03-19
Move some CWS workflows out of the handlers and into dedicated modules

lerks:
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/905)

### DIFF
--- a/cms/db/user.py
+++ b/cms/db/user.py
@@ -152,7 +152,10 @@ class Participation(Base):
         Integer,
         primary_key=True)
 
-    # The user can log in CWS only from this IP address or subnet.
+    # If the IP lock is enabled the user can log into CWS only if their
+    # requests come from an IP address that belongs to any of these
+    # subnetworks. An empty list prevents the user from logging in,
+    # None disables the IP lock for the user.
     ip = Column(
         CastingArray(CIDR),
         nullable=True)
--- a/cms/server/__init__.py
+++ b/cms/server/__init__.py
@@ -25,12 +25,12 @@ from future.builtins.disabled import *  # noqa
 from future.builtins import *  # noqa
 
 from .util import \
-    CommonRequestHandler, actual_phase_required, compute_actual_phase, \
-    file_handler_gen, filter_ascii, Url, multi_contest
+    CommonRequestHandler, file_handler_gen, filter_ascii, Url, \
+    multi_contest
 
 
 __all__ = [
     # util
-    "CommonRequestHandler", "actual_phase_required", "compute_actual_phase",
-    "file_handler_gen", "filter_ascii", "Url", "multi_contest",
+    "CommonRequestHandler", "file_handler_gen", "filter_ascii", "Url",
+    "multi_contest",
 ]
--- a/cms/server/contest/authentication.py
+++ b/cms/server/contest/authentication.py
@@ -1,0 +1,376 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+# Contest Management System - http://cms-dev.github.io/
+# Copyright © 2010-2014 Giovanni Mascellani <mascellani@poisson.phc.unipi.it>
+# Copyright © 2010-2016 Stefano Maggiolo <s.maggiolo@gmail.com>
+# Copyright © 2010-2012 Matteo Boscariol <boscarim@hotmail.com>
+# Copyright © 2012-2018 Luca Wehrstedt <luca.wehrstedt@gmail.com>
+# Copyright © 2013 Bernard Blackham <bernard@largestprime.net>
+# Copyright © 2014 Artem Iglikov <artem.iglikov@gmail.com>
+# Copyright © 2014 Fabian Gundlach <320pointsguy@gmail.com>
+# Copyright © 2015-2016 William Di Luigi <williamdiluigi@gmail.com>
+# Copyright © 2016 Myungwoo Chun <mc.tamaki@gmail.com>
+# Copyright © 2016 Amir Keivan Mohtashami <akmohtashami97@gmail.com>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+from future.builtins.disabled import *  # noqa
+from future.builtins import *  # noqa
+
+import ipaddress
+import logging
+import pickle
+from datetime import timedelta
+
+from sqlalchemy.orm import contains_eager
+
+from cms import config
+from cms.db import Participation, User
+from cms.server import filter_ascii
+from cmscommon.crypto import validate_password
+from cmscommon.datetime import make_datetime, make_timestamp
+
+logger = logging.getLogger(__name__)
+
+
+def check_ip(address, networks):
+    """Return if client IP belongs to one of the accepted networks.
+
+    address (bytes): IP address to verify.
+    networks ([ipaddress.IPv4Network|ipaddress.IPv6Network]): IP
+        networks (addresses w/ subnets) to check against.
+
+    return (bool): whether the address belongs to one of the networks.
+
+    """
+    try:
+        address = ipaddress.ip_address(str(address))
+    except ValueError:
+        return False
+
+    for network in networks:
+        if address in network:
+            return True
+
+    return False
+
+
+def validate_first_login(
+        sql_session, contest, timestamp, username, password, ip_address):
+    """Authenticate a user logging in, with username and password.
+
+    Given the information the user provided (the username and the
+    password) and some context information (contest, to determine which
+    users are allowed to log in, how and with which restrictions;
+    timestamp for cookie creation; IP address to check against) try to
+    authenticate the user and return its participation and the cookie
+    to set to help authenticate future visits.
+
+    After finding the participation, IP login and hidden users
+    restrictions are checked.
+
+    sql_session (Session): the SQLAlchemy database session used to
+        execute queries.
+    contest (Contest): the contest the user is trying to access.
+    timestamp (datetime): the date and the time of the request.
+    username (str): the username the user provided.
+    password (str): the password the user provided.
+    ip_address (str): the IP address the request came from.
+
+    return ((Participation|None), (bytes|None)): if the user couldn't
+        be authenticated then return None, otherwise return the
+        participation that they wanted to authenticate as; if a cookie
+        has to be set return it as well, otherwise return None.
+
+    """
+    filtered_user = filter_ascii(username)
+
+    def log_failed_attempt(msg, *args):
+        logger.info("Unsuccessful login attempt from IP address %s, as user "
+                    "%s, on contest %s, at %s: " + msg, ip_address,
+                    filtered_user, contest.name, timestamp, *args)
+
+    if not contest.allow_password_authentication:
+        log_failed_attempt("password authentication not allowed")
+        return None, None
+
+    participation = sql_session.query(Participation) \
+        .join(Participation.user) \
+        .options(contains_eager(Participation.user)) \
+        .filter(Participation.contest == contest)\
+        .filter(User.username == username)\
+        .first()
+
+    if participation is None:
+        log_failed_attempt("user not registered to contest")
+        return None, None
+
+    # If a contest-specific password is defined, use that. If it's
+    # not, use the user's main password.
+    if participation.password is None:
+        correct_password = participation.user.password
+    else:
+        correct_password = participation.password
+
+    try:
+        password_valid = validate_password(correct_password, password)
+    except ValueError:
+        # This is either a programming or a configuration error.
+        logger.warning("Invalid password stored in database for user %s in "
+                       "contest %s", filtered_user, contest.name)
+        return None, None
+
+    if not password_valid:
+        log_failed_attempt("wrong password")
+        return None, None
+
+    if contest.ip_restriction and participation.ip is not None \
+            and not check_ip(ip_address, participation.ip):
+        log_failed_attempt("unauthorized IP address")
+        return None, None
+
+    if contest.block_hidden_participations and participation.hidden:
+        log_failed_attempt("participation is hidden and unauthorized")
+        return None, None
+
+    logger.info("Successful login attempt from IP address %s, as user %s, on "
+                "contest %s, at %s", ip_address, filtered_user, contest.name,
+                timestamp)
+
+    return (participation,
+            pickle.dumps((username, password, make_timestamp(timestamp))))
+
+
+def validate_returning_login(
+        sql_session, contest, timestamp, cookie, ip_address):
+    """Authenticate a user returning to the site, with a cookie.
+
+    Given the information the user's browser provided (the cookie) and
+    some context information (contest, to determine which users are
+    allowed to log in, how and with which restrictions; timestamp for
+    cookie validation/creation, IP address to either do autologin or to
+    check against) try to authenticate the user and return its
+    participation and the cookie to reset to help authenticate future
+    visits.
+
+    There are two way a user can authenticate:
+    - if IP autologin is enabled, we look for a participation whose IP
+      address matches the remote IP address; if a match is found, the
+      user is authenticated as that participation;
+    - if username/password authentication is enabled, and the cookie
+      is valid, the corresponding participation is returned, and the
+      cookie is refreshed.
+
+    After finding the participation, IP login and hidden users
+    restrictions are checked.
+
+    In case of any error, or of a login by other sources, the cookie is
+    deleted.
+
+    sql_session (Session): the SQLAlchemy database session used to
+        execute queries.
+    contest (Contest): the contest the user is trying to access.
+    timestamp (datetime): the date and the time of the request.
+    cookie (bytes): the cookie the user's browser provided in the
+        request.
+    ip_address (str): the IP address the request came from.
+
+    return ((Participation|None), (bytes|None)): if the user couldn't
+        be authenticated then return None, otherwise return the
+        participation that they wanted to authenticate as; if a cookie
+        has to be set return it as well, otherwise return None.
+
+    """
+    participation = None
+
+    if contest.ip_autologin:
+        try:
+            participation = authenticate_by_ip_address(
+                sql_session, contest, ip_address)
+            # If the login is IP-based, we delete previous cookies.
+            if participation is not None:
+                cookie = None
+        except RuntimeError:
+            return None, None
+
+    if participation is None \
+            and contest.allow_password_authentication:
+        participation, cookie = authenticate_from_cookie(
+            sql_session, contest, timestamp, cookie)
+
+    if participation is None:
+        return None, None
+
+    # Check if user is using the right IP (or is on the right subnet),
+    # and that is not hidden if hidden users are blocked.
+    ip_login_restricted = \
+        contest.ip_restriction and participation.ip is not None \
+        and not check_ip(ip_address, participation.ip)
+    hidden_user_restricted = \
+        contest.block_hidden_participations and participation.hidden
+    if ip_login_restricted or hidden_user_restricted:
+        return None, None
+
+    return participation, cookie
+
+
+def authenticate_by_ip_address(sql_session, contest, ip_address):
+    """Return the current participation based on the IP address.
+
+    sql_session (Session): the SQLAlchemy database session used to
+        execute queries.
+    contest (Contest): the contest the user is trying to access.
+    ip_address (str): the IP address the request came from.
+
+    return (Participation|None): the only participation that is allowed
+        to connect from the given IP address, or None if not found.
+
+    raise (RuntimeError): if there is more than one participation
+        matching the remote IP address.
+
+    """
+    try:
+        remote_ip = ipaddress.ip_address(ip_address)
+        # We encode it as a network (i.e., we assign it a /32 or
+        # /128 mask) since we're comparing it for equality with
+        # other networks.
+        remote_ip_net = ipaddress.ip_network((remote_ip,
+                                              remote_ip.max_prefixlen))
+    except ValueError:
+        # This is a programming error.
+        logger.warning("Invalid IP address received from web server: %s",
+                       ip_address)
+        return None
+
+    participations = sql_session.query(Participation) \
+        .filter(Participation.contest == contest) \
+        .filter(Participation.ip.any(remote_ip_net))
+
+    # If hidden users are blocked we ignore them completely.
+    if contest.block_hidden_participations:
+        participations = participations \
+            .filter(Participation.hidden.is_(False))
+
+    participations = participations.all()
+
+    if len(participations) == 1:
+        participation = participations[0]
+        logger.info(
+            "Successful IP authentication from IP address %s, as user %s, on "
+            "contest %s", ip_address, participation.user.username, contest.name)
+        return participation
+
+    # Having more than participation with the same IP, is a mistake and
+    # should not happen. In such case, we disallow login for that IP
+    # completely, in order to make sure the problem is noticed.
+    if len(participations) > 1:
+        # This is a configuration error.
+        logger.warning(
+            "Ambiguous IP address %s, assigned to %d participations.",
+            ip_address, len(participations))
+        raise RuntimeError("More than one participation with the same IP.")
+
+    logger.info(
+        "Unsuccessful IP authentication from IP address %s, on contest %s: no "
+        "user matches the IP address", ip_address, contest.name)
+    return None
+
+
+def authenticate_from_cookie(sql_session, contest, timestamp, cookie):
+    """Return the current participation based on the cookie.
+
+    If a participation can be extracted, the cookie is refreshed.
+
+    sql_session (Session): the SQLAlchemy database session used to
+        execute queries.
+    contest (Contest): the contest the user is trying to access.
+    timestamp (datetime): the date and the time of the request.
+    cookie (bytes): the cookie the user's browser provided in the
+        request.
+
+    return ((Participation|None), (bytes|None)): the participation
+        extracted from the cookie and the cookie to set/refresh, or
+        None in case of errors.
+
+    """
+    if cookie is None:
+        logger.info("Unsuccessful cookie authentication: no cookie provided")
+        return None, None
+
+    # Parse cookie.
+    try:
+        cookie = pickle.loads(cookie)
+        username = cookie[0]
+        password = cookie[1]
+        last_update = make_datetime(cookie[2])
+    except Exception as e:
+        # Cookies are stored securely and thus cannot be tampered with:
+        # this is either a programming or a configuration error.
+        logger.warning("Invalid cookie (%s): %s", e, cookie)
+        return None, None
+
+    filtered_user = filter_ascii(username)
+
+    def log_failed_attempt(msg, *args):
+        logger.info("Unsuccessful cookie authentication as %s, returning from "
+                    "%s, at %s: " + msg, filtered_user, last_update, timestamp,
+                    *args)
+
+    # Check if the cookie is expired.
+    if timestamp - last_update > timedelta(seconds=config.cookie_duration):
+        log_failed_attempt("cookie expired (lasts %d seconds)",
+                           config.cookie_duration)
+        return None, None
+
+    # Load participation from DB and make sure it exists.
+    participation = sql_session.query(Participation) \
+        .join(Participation.user) \
+        .options(contains_eager(Participation.user)) \
+        .filter(Participation.contest == contest) \
+        .filter(User.username == username) \
+        .first()
+    if participation is None:
+        log_failed_attempt("user not registered to contest")
+        return None, None
+
+    # Check that the password is correct (if a contest-specific
+    # password is defined, use that instead of the user password).
+    if participation.password is None:
+        correct_password = participation.user.password
+    else:
+        correct_password = participation.password
+
+    try:
+        password_valid = validate_password(correct_password, password)
+    except ValueError:
+        # This is either a programming or a configuration error.
+        logger.warning("Invalid password stored in database for user %s in "
+                       "contest %s", filtered_user, contest.name)
+        return None, None
+
+    if not password_valid:
+        log_failed_attempt("wrong password")
+        return None, None
+
+    logger.info("Successful cookie authentication as user %s, on contest %s, "
+                "returning from %s, at %s", filtered_user, contest.name,
+                last_update, timestamp)
+
+    return (participation,
+            pickle.dumps((username, password, make_timestamp(timestamp))))
--- a/cms/server/contest/authentication.py
+++ b/cms/server/contest/authentication.py
@@ -110,10 +110,10 @@ def validate_first_login(
 
     try:
         password_valid = validate_password(correct_password, password)
-    except ValueError:
+    except ValueError as e:
         # This is either a programming or a configuration error.
         logger.warning("Invalid password stored in database for user %s in "
-                       "contest %s", filtered_user, contest.name)
+                       "contest %s: %s", filtered_user, contest.name, e)
         return None, None
 
     if not password_valid:
@@ -331,10 +331,10 @@ def authenticate_from_cookie(sql_session, contest, timestamp, cookie):
 
     try:
         password_valid = validate_password(correct_password, password)
-    except ValueError:
+    except ValueError as e:
         # This is either a programming or a configuration error.
         logger.warning("Invalid password stored in database for user %s in "
-                       "contest %s", filtered_user, contest.name)
+                       "contest %s: %s", filtered_user, contest.name, e)
         return None, None
 
     if not password_valid:
--- a/cms/server/contest/authentication.py
+++ b/cms/server/contest/authentication.py
@@ -201,14 +201,21 @@ def authenticate_request(
     if participation is None:
         return None, None
 
-    # Check if user is using the right IP (or is on the right subnet),
-    # and that is not hidden if hidden users are blocked.
-    ip_login_restricted = \
-        contest.ip_restriction and participation.ip is not None \
-        and not any(ip_address in network for network in participation.ip)
-    hidden_user_restricted = \
-        contest.block_hidden_participations and participation.hidden
-    if ip_login_restricted or hidden_user_restricted:
+    # Check if user is using the right IP (or is on the right subnet).
+    if contest.ip_restriction and participation.ip is not None \
+            and not any(ip_address in network for network in participation.ip):
+        logger.info(
+            "Unsuccessful authentication from IP address %s, on contest %s, "
+            "as %s, at %s: unauthorized IP address",
+            ip_address, contest.name, participation.user.username, timestamp)
+        return None, None
+
+    # Check that the user is not hidden if hidden users are blocked.
+    if contest.block_hidden_participations and participation.hidden:
+        logger.info(
+            "Unsuccessful authentication from IP address %s, on contest %s, "
+            "as %s, at %s: participation is hidden and unauthorized",
+            ip_address, contest.name, participation.user.username, timestamp)
         return None, None
 
     return participation, cookie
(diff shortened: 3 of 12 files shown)


stefano-maggiolo:
I only looked at the second commit (I looked at first but assuming the content was the same, which I haven't checked, if it was then it's fine).

---

Review status: 0 of 11 files reviewed at latest revision, 24 unresolved discussions, some commit checks failed.

---

*[cms/server/contest/authentication.py, line 48 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L84S3jWrfUoUXFaVziX:-L84S3jWrfUoUXFaVziY:b-e4ts3z) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L48)):*
> ```Python
> from cmscommon.crypto import validate_password
> from cmscommon.datetime import make_datetime, make_timestamp
> ```

Nit: two lines

---

*[cms/server/contest/authentication.py, line 55 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L84SLU0v797NNDHpOFL:-L84SLU0v797NNDHpOFM:bqzeg5j) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L55)):*
> ```Python
>     """Return if client IP belongs to one of the accepted networks.
> 
>     address (bytes): IP address to verify.
> ```

Is this in b"192.168.0.1" or b"\xC0\xA8\x00\x01"?

---

*[cms/server/contest/authentication.py, line 94 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L84SV6UFpW9qKoLJaXN:-L84SV6UFpW9qKoLJaXO:b-4kdk6v) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L94)):*
> ```Python
>     username (str): the username the user provided.
>     password (str): the password the user provided.
>     ip_address (str): the IP address the request came from.
> ```

Same, format?

---

*[cms/server/contest/authentication.py, line 96 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L86xSam_AdD4ErPFBjW:-L86xSam_AdD4ErPFBjX:b-4v74w7) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L96)):*
> ```Python
>     ip_address (str): the IP address the request came from.
> 
>     return ((Participation|None), (bytes|None)): if the user couldn't
> ```

might be simpler as `((Participation, bytes)|(None, None))`? Also a nit is that this function doesn't know if the cookie needs to be set or not, just returns an authentication cookie and the caller decides what to do.

---

*[cms/server/contest/authentication.py, line 136 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L86wHanssmsbyXNQSFL:-L86wHanssmsbyXNQSFM:b-gvc0h2) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L136)):*
> ```Python
>         # This is either a programming or a configuration error.
>         logger.warning("Invalid password stored in database for user %s in "
>                        "contest %s", filtered_user, contest.name)
> ```

Logging the message of the exception would give info on which of the two problem this is

---

*[cms/server/contest/authentication.py, line 160 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L874OeL3uWQoFy9lsiy:-L874OeL3uWQoFy9lsiz:b-w0o4sj) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L160)):*
> ```Python
> 
> 
> def validate_returning_login(
> ```

I'm confused by the names and scopes of these two methods - mainly I started commenting on the other that it was more like "authenticate_from_password" but then I realized "login" might be synonym for (username, password). But now you are using login for either (username, password) (via cookie) or ip address, so I'm confused.

So maybe I'll make them more different, like `validate_login` and `authenticate_request`? Or `validate_login` and `authenticate_implicit` (there is another word better than implicit, but at the moment I cannot remember it - what I want to say is "background" "without user interaction").

---

*[cms/server/contest/authentication.py, line 184 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L876B0S-u4Ds7XMEmCt:-L876B0S-u4Ds7XMEmCu:b-5o1odb) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L184)):*
> ```Python
> 
>     In case of any error, or of a login by other sources, the cookie is
>     deleted.
> ```

Decision of the caller, not something that happens here.

---

*[cms/server/contest/authentication.py, line 192 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L876F_ja2AKla0Gq_41:-L876F_ja2AKla0Gq_42:b-j60dct) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L192)):*
> ```Python
>     cookie (bytes): the cookie the user's browser provided in the
>         request.
>     ip_address (str): the IP address the request came from.
> ```

Format?

---

*[cms/server/contest/authentication.py, line 194 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L876DaqTxAzpNmXJDAM:-L876DaqTxAzpNmXJDAN:bf68xcn) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L194)):*
> ```Python
>     ip_address (str): the IP address the request came from.
> 
>     return ((Participation|None), (bytes|None)): if the user couldn't
> ```

Same comment as before

---

*[cms/server/contest/authentication.py, line 206 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L876LFhXaCG-kjZQTUW:-L876LFhXaCG-kjZQTUX:b702owa) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L206)):*
> ```Python
>             participation = authenticate_by_ip_address(
>                 sql_session, contest, ip_address)
>             # If the login is IP-based, we delete previous cookies.
> ```

Decision of the caller.

---

*[cms/server/contest/authentication.py, line 224 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L87E18kWs9lTNCabiw3:-L87E18l6pGjOUbi7Uza:bchsha2) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L224)):*
> ```Python
>     ip_login_restricted = \
>         contest.ip_restriction and participation.ip is not None \
>         and not check_ip(ip_address, participation.ip)
> ```

Sorry again for something not your fault: can you add a comment to user.py about the meaning of null participation.ip?

---

*[cms/server/contest/authentication.py, line 233 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L87oi3BeloPS4QdAOoo:-L87oi3BeloPS4QdAOop:b-vedmkb) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L233)):*
> ```Python
> 
> 
> def authenticate_by_ip_address(sql_session, contest, ip_address):
> ```

This and the cookie one should be private to the module, right? Can you use __all__ or underscore these?

---

*[cms/server/contest/authentication.py, line 239 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L87EQYhV5tVm1_LGnO7:-L87EQYhV5tVm1_LGnO8:b-89a4l7) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L239)):*
> ```Python
>         execute queries.
>     contest (Contest): the contest the user is trying to access.
>     ip_address (str): the IP address the request came from.
> ```

Ditto

---

*[cms/server/contest/authentication.py, line 244 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L876WQXd16hDWD6xiwq:-L876WQXd16hDWD6xiwr:bw1du56) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L244)):*
> ```Python
>         to connect from the given IP address, or None if not found.
> 
>     raise (RuntimeError): if there is more than one participation
> ```

Not your fault, but maybe there is a better exception type? LookupError? 

---

*[cms/server/contest/authentication.py, line 263 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L87E_0GJCNBNqTzjRJu:-L87E_0GJCNBNqTzjRJv:b9amq4a) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L263)):*
> ```Python
>     participations = sql_session.query(Participation) \
>         .filter(Participation.contest == contest) \
>         .filter(Participation.ip.any(remote_ip_net))
> ```

Want to eagerload user here as well?

---

*[cms/server/contest/authentication.py, line 268 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L87ElzvXIjqW3gX_NPw:-L87ElzvXIjqW3gX_NPx:b-o3ub6h) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L268)):*
> ```Python
>     if contest.block_hidden_participations:
>         participations = participations \
>             .filter(Participation.hidden.is_(False))
> ```

Not sure about this: I think we should filter post-facto because we still want to highlight if there is a hidden participation with the same ip?

---

*[cms/server/contest/authentication.py, line 291 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L87M4L5boYwC16Mi3kz:-L87M4L5boYwC16Mi3l-:b-3efo46) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L291)):*
> ```Python
>     logger.info(
>         "Unsuccessful IP authentication from IP address %s, on contest %s: no "
>         "user matches the IP address", ip_address, contest.name)
> ```

Can we reformat this as all other methods: first the "early"-return problematic conditions, then the default path?

---

*[cms/server/contest/authentication.py, line 312 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L87M_xuXZn0OCsg2F9e:-L87M_xuXZn0OCsg2F9f:b-fz4d8j) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L312)):*
> ```Python
> 
>     """
>     if cookie is None:
> ```

If it can be none, say it in the docstring

---

*[cms/server/contest/authentication.py, line 369 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L87MrLSZV32g_nr0frh:-L87MrLSZV32g_nr0fri:bx760fk) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L369)):*
> ```Python
>     if not password_valid:
>         log_failed_attempt("wrong password")
>         return None, None
> ```

Can we extract the last 15-ish lines to a `safe_check_password(participation)`? It is duplicated at least twice.

---

*[cmstestsuite/unit_tests/server/contest/authentication_test.py, line 77 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L86xmfKj-kZqjGS4o3Z:-L86xmfKj-kZqjGS4o3_:bhie4xz) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cmstestsuite/unit_tests/server/contest/authentication_test.py#L77)):*
> ```Python
> 
>     def test_successful_login(self):
>         self.contest.allow_password_authentication = True
> ```

Want to put this on top and amend when necessary?

---

*[cmstestsuite/unit_tests/server/contest/authentication_test.py, line 128 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L86y-eGet96zTaRBn41:-L86y-eGet96zTaRBn42:b4hkfud) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cmstestsuite/unit_tests/server/contest/authentication_test.py#L128)):*
> ```Python
> 
>         # Mainly checks that no exception is raised.
>         self.assertFailure("myuser", "mypass", "127.0.0.1")
> ```

Nit: "myotherpass"

---

*[cmstestsuite/unit_tests/server/contest/authentication_test.py, line 139 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L86yEJ3TvNTfR4cYJKv:-L86yEJ3TvNTfR4cYJKw:be9wdpx) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cmstestsuite/unit_tests/server/contest/authentication_test.py#L139)):*
> ```Python
>         self.assertFailure("myuser", "mypass", "10.0.1.1")
> 
>         # Corner cases.
> ```

Can you try also when there are two ip networks, and you match the second, and possibly when you match a single IP (can do both in the same check).

---

*[cmstestsuite/unit_tests/server/contest/authentication_test.py, line 168 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L87O0S-t-Hfz25Axh4M:-L87O0S-t-Hfz25Axh4N:b-g5749q) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cmstestsuite/unit_tests/server/contest/authentication_test.py#L168)):*
> ```Python
>             "mypass", "10.0.0.1")
> 
>     def attempt_authentication(self, **kwargs):
> ```

This style of using kwargs is ... weird? Just wanted to avoid the verbosity of passing ts, cookie, ip to each?

---

*[cmstestsuite/unit_tests/server/contest/authentication_test.py, line 182 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L87mSd70tXSxz0jkaVm:-L87mSd70tXSxz0jkaVn:b-l4n1dt) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cmstestsuite/unit_tests/server/contest/authentication_test.py#L182)):*
> ```Python
>         return cookie
> 
>     def assertSuccessWithCookie(self, **kwargs):
> ```

Can you please write a short description of what each of these assert method does?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/905#-:-L87ouGlACpvDEOneT9m:by76ni7)*

lerks:
Yes, the first commit only moves the code without modifications

I fixed most (but not all) remarks. PTAL, and after we settle the discussions I will do the rest.

---

Review status: 0 of 12 files reviewed at latest revision, 24 unresolved discussions.

---

*[cms/server/contest/authentication.py, line 48 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L84S3jWrfUoUXFaVziX:-L8DObm5TAgSsEAhugjG:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L48)):*

Nit: two lines
</blockquote></details>

Done.

---

*[cms/server/contest/authentication.py, line 55 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L84SLU0v797NNDHpOFL:-L8DTw3eJAZCtUfLrTzB:b-2u0ve7) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L55)):*

Is this in b"192.168.0.1" or b"\xC0\xA8\x00\x01"?
</blockquote></details>

Right, this is confusing. Let's have the user of this module (i.e., CWS) decode the IP address and provide it as an object. That simplifies this function to the point that it can be inlined in the two places where it's called.

---

*[cms/server/contest/authentication.py, line 94 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L84SV6UFpW9qKoLJaXN:-L8DUI-xNskfa6RcnAVw:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L94)):*

Same, format?
</blockquote></details>

Done.

---

*[cms/server/contest/authentication.py, line 96 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L86xSam_AdD4ErPFBjW:-L8DWBYuYDBFjtKtJHGE:bgjkcof) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L96)):*

might be simpler as `((Participation, bytes)|(None, None))`? Also a nit is that this function doesn't know if the cookie needs to be set or not, just returns an authentication cookie and the caller decides what to do.
</blockquote></details>

Done:
* I used your suggestion for the return type across the board, but it had to become a tad more complex for validate_returning_login (yet, now, it captures the fact that cookie is None => participation is None)
* I reworded some of the language used for cookies. The "has to be set/refreshed" phrases seem fine to me as they describe the intent the cookie is returned for.

---

*[cms/server/contest/authentication.py, line 136 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L86wHanssmsbyXNQSFL:-L8DX2tbz07wnwSJVrIH:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L136)):*

Logging the message of the exception would give info on which of the two problem this is
</blockquote></details>

Done.

---

*[cms/server/contest/authentication.py, line 160 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L874OeL3uWQoFy9lsiy:-L8DaxkzCfr8GRwngCuM:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L160)):*

I'm confused by the names and scopes of these two methods - mainly I started commenting on the other that it was more like "authenticate_from_password" but then I realized "login" might be synonym for (username, password). But now you are using login for either (username, password) (via cookie) or ip address, so I'm confused.

So maybe I'll make them more different, like `validate_login` and `authenticate_request`? Or `validate_login` and `authenticate_implicit` (there is another word better than implicit, but at the moment I cannot remember it - what I want to say is "background" "without user interaction").
</blockquote></details>

Done.

---

*[cms/server/contest/authentication.py, line 184 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L876B0S-u4Ds7XMEmCt:-L8DXg599lI0dLh_werq:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L184)):*

Decision of the caller, not something that happens here.
</blockquote></details>

Done.

---

*[cms/server/contest/authentication.py, line 192 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L876F_ja2AKla0Gq_41:-L8DXglxrj9Kr5uwjFtf:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L192)):*

Format?
</blockquote></details>

Done.

---

*[cms/server/contest/authentication.py, line 194 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L876DaqTxAzpNmXJDAM:-L8DXhIn3EJ7ym19wmG_:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L194)):*

Same comment as before
</blockquote></details>

Done.

---

*[cms/server/contest/authentication.py, line 206 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L876LFhXaCG-kjZQTUW:-L8DXwIh9M5dc81zCHw3:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L206)):*

Decision of the caller.
</blockquote></details>

Done.

---

*[cms/server/contest/authentication.py, line 224 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L87E18kWs9lTNCabiw3:-L8DfQ_ffuiWXjCWXYdF:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L224)):*

Sorry again for something not your fault: can you add a comment to user.py about the meaning of null participation.ip?
</blockquote></details>

Done.

---

*[cms/server/contest/authentication.py, line 233 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L87oi3BeloPS4QdAOoo:-L8Db0WwHZmSKYBEYYD2:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L233)):*

This and the cookie one should be private to the module, right? Can you use __all__ or underscore these?
</blockquote></details>

Done.

---

*[cms/server/contest/authentication.py, line 239 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L87EQYhV5tVm1_LGnO7:-L8DY0UXQ41psmF6ZhOC:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L239)):*

Ditto
</blockquote></details>

Done.

---

*[cms/server/contest/authentication.py, line 244 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L876WQXd16hDWD6xiwq:-L8DZi8Sd1ZsTHKsEjdY:b-rh5w9n) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L244)):*

Not your fault, but maybe there is a better exception type? LookupError? 
</blockquote></details>

I propose to remove this entirely. Right now an exception is raised to signal to the caller that the user shouldn't be allowed to log in *in any way*, not even if they have a valid cookie. I don't really understand why. If IP autologin is ambiguous for this user we log the misconfiguration for the admin to see, we prevent the user from using it. Banning the user feels punitive and may create issues/appeals in real contests.

---

*[cms/server/contest/authentication.py, line 263 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L87E_0GJCNBNqTzjRJu:-L8DYQd1pdxZ0tdRE-c1:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L263)):*

Want to eagerload user here as well?
</blockquote></details>

Done.

---

*[cms/server/contest/authentication.py, line 268 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L87ElzvXIjqW3gX_NPw:-L8DZSLzhQmx7u8Esn3a:b-nsxaju) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L268)):*

Not sure about this: I think we should filter post-facto because we still want to highlight if there is a hidden participation with the same ip?
</blockquote></details>

The comment hints to the fact that this was a conscious design decision. I didn't change this even if I don't really get what scenario the author had in mind. Do you still think we should change this?

---

*[cms/server/contest/authentication.py, line 291 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L87M4L5boYwC16Mi3kz:-L8DbUDbdjCkxef2McuL:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L291)):*

Can we reformat this as all other methods: first the "early"-return problematic conditions, then the default path?
</blockquote></details>

Done.

---

*[cms/server/contest/authentication.py, line 312 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L87M_xuXZn0OCsg2F9e:-L8DYk8-qdjgGbtRjbE8:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L312)):*

If it can be none, say it in the docstring
</blockquote></details>

Done.

---

*[cms/server/contest/authentication.py, line 369 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L87MrLSZV32g_nr0frh:-L8DkmBdViPGtVJ0j3HG:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L369)):*

Can we extract the last 15-ish lines to a `safe_check_password(participation)`? It is duplicated at least twice.
</blockquote></details>

Done.

---

*[cmstestsuite/unit_tests/server/contest/authentication_test.py, line 128 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L86y-eGet96zTaRBn41:-L8EVmv3rD05-cu0s5H3:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cmstestsuite/unit_tests/server/contest/authentication_test.py#L128)):*

Nit: "myotherpass"
</blockquote></details>

Done.

---

*[cmstestsuite/unit_tests/server/contest/authentication_test.py, line 139 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L86yEJ3TvNTfR4cYJKv:-L8EWU2ysvgdFI3p2-J9:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cmstestsuite/unit_tests/server/contest/authentication_test.py#L139)):*

Can you try also when there are two ip networks, and you match the second, and possibly when you match a single IP (can do both in the same check).
</blockquote></details>

Done.

---

*[cmstestsuite/unit_tests/server/contest/authentication_test.py, line 168 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L87O0S-t-Hfz25Axh4M:-L8EWffODqlqN0FD4Lqc:b-adxhy5) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cmstestsuite/unit_tests/server/contest/authentication_test.py#L168)):*

This style of using kwargs is ... weird? Just wanted to avoid the verbosity of passing ts, cookie, ip to each?
</blockquote></details>

There is that but, more pragmatically, I need to be able to explicitly pass `cookie=None` without that being interpreted as "no cookie provided" (and hence triggering the use of  the default value.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/905#-:-L8DON9cl00ovszlnP-X:b-wjjb3e)*

stefano-maggiolo:
Review status: 0 of 12 files reviewed at latest revision, 5 unresolved discussions.

---

*[cms/server/contest/authentication.py, line 194 at r2](https://beta.reviewable.io/reviews/cms-dev/cms/905#-L876DaqTxAzpNmXJDAM:-L8EfxPxA_QBqoyMwgLv:b-kcxpmd) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L194)):*

Done.
</blockquote></details>

Fancy :)

---

*[cms/server/contest/authentication.py, line 244 at r2](https://beta.reviewable.io/reviews/cms-dev/cms/905#-L876WQXd16hDWD6xiwq:-L8EgGv427vHQFzTWBlB:b-da5vv4) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L244)):*

I propose to remove this entirely. Right now an exception is raised to signal to the caller that the user shouldn't be allowed to log in *in any way*, not even if they have a valid cookie. I don't really understand why. If IP autologin is ambiguous for this user we log the misconfiguration for the admin to see, we prevent the user from using it. Banning the user feels punitive and may create issues/appeals in real contests.
</blockquote></details>

I think the rationale is that if there is such a misconfiguration *you don't know* which of the 2+ users with that IP is using the interface. You're saying "if they have a cookie, let them in" but it's difficult to understand how they would have gotten a cookie in the first place.

I don't want to argue too much because we're talking about something that changes behaviour only for very unlikely situations. If you feel strongly that a cookie should let them in, I won't object. In case, maybe in a different commit?

---

*[cms/server/contest/authentication.py, line 268 at r2](https://beta.reviewable.io/reviews/cms-dev/cms/905#-L87ElzvXIjqW3gX_NPw:-L8EhezJ6YoGnSEc736L:b-g0wwwt) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L268)):*

The comment hints to the fact that this was a conscious design decision. I didn't change this even if I don't really get what scenario the author had in mind. Do you still think we should change this?
</blockquote></details>

I guess this is even more remote than the situation in the other discussion... let's decide what to do in that one first.

---

*[cms/server/contest/authentication.py, line 68 at r19](https://beta.reviewable.io/reviews/cms-dev/cms/905#-L8Ef47i25zFSiNg0aFu:-L8Ef47i25zFSiNg0aFv:b-v836ov) ([raw file](https://github.com/cms-dev/cms/blob/d9920040f9f661a707fbb8b24a0d9e329eb0b9ec/cms/server/contest/authentication.py#L68)):*
> ```Python
> 
>     return (bool): whether the password matches the expected one.
>     """
> ```

Nit: empty line

---

*[cms/server/contest/authentication.py, line 257 at r19](https://beta.reviewable.io/reviews/cms-dev/cms/905#-L8EgzaG5sxgXxUGrSJO:-L8EgzaG5sxgXxUGrSJP:b-rs2ayg) ([raw file](https://github.com/cms-dev/cms/blob/d9920040f9f661a707fbb8b24a0d9e329eb0b9ec/cms/server/contest/authentication.py#L257)):*
> ```Python
>     # We encode it as a network (i.e., we assign it a /32 or /128 mask)
>     # since we're comparing it for equality with other networks.
>     ip_network = ipaddress.ip_network((ip_address, ip_address.max_prefixlen))
> ```

Technically, ip_network can raise ValueError. Technically, it shouldn't accept this format too... so yeah, leave it as it is.

---

*[cmstestsuite/unit_tests/server/contest/authentication_test.py, line 48 at r19](https://beta.reviewable.io/reviews/cms-dev/cms/905#-L8EiPZ3E_uPYE1zL74t:-L8EiPZ3E_uPYE1zL74u:b-ec4vm2) ([raw file](https://github.com/cms-dev/cms/blob/d9920040f9f661a707fbb8b24a0d9e329eb0b9ec/cmstestsuite/unit_tests/server/contest/authentication_test.py#L48)):*
> ```Python
> 
> 
> class TestValidateFirstLogin(DatabaseMixin, unittest.TestCase):
> ```

You should change these names too

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cms-dev/cms/905)*

lerks:
All done, PTAL

---

Review status: 0 of 12 files reviewed at latest revision, 5 unresolved discussions.

---

*[cms/server/contest/authentication.py, line 244 at r2](https://beta.reviewable.io/reviews/cms-dev/cms/905#-L876WQXd16hDWD6xiwq:-L8ExNi-5SKayoZ58Trt:bsymlkz) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L244)):*

I think the rationale is that if there is such a misconfiguration *you don't know* which of the 2+ users with that IP is using the interface. You're saying "if they have a cookie, let them in" but it's difficult to understand how they would have gotten a cookie in the first place.

I don't want to argue too much because we're talking about something that changes behaviour only for very unlikely situations. If you feel strongly that a cookie should let them in, I won't object. In case, maybe in a different commit?
</blockquote></details>

I don't have a strong opinion in either direction. Let's keep it like it was and, as you said, if we decide to change the semantics we do it in another dedicated commit.

I have defined a dedicated exception to replace RuntimeError. It seemed the cleanest solution.

---

*[cms/server/contest/authentication.py, line 268 at r2](https://beta.reviewable.io/reviews/cms-dev/cms/905#-L87ElzvXIjqW3gX_NPw:-L8ExaZF6qbssKRd1K-8:b-n7o2dh) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L268)):*

I guess this is even more remote than the situation in the other discussion... let's decide what to do in that one first.
</blockquote></details>

As above: let's not change semantics here.

---

*[cms/server/contest/authentication.py, line 68 at r19](https://beta.reviewable.io/reviews/cms-dev/cms/905#-L8Ef47i25zFSiNg0aFu:-L8EutEhEEFT64CwhICZ:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/d9920040f9f661a707fbb8b24a0d9e329eb0b9ec/cms/server/contest/authentication.py#L68)):*

Nit: empty line
</blockquote></details>

Done.

---

*[cmstestsuite/unit_tests/server/contest/authentication_test.py, line 77 at r2](https://beta.reviewable.io/reviews/cms-dev/cms/905#-L86xmfKj-kZqjGS4o3Z:-L8EyqNUC6BCgQKEt2z-:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cmstestsuite/unit_tests/server/contest/authentication_test.py#L77)):*

Want to put this on top and amend when necessary?
</blockquote></details>

Done.

---

*[cmstestsuite/unit_tests/server/contest/authentication_test.py, line 182 at r2](https://beta.reviewable.io/reviews/cms-dev/cms/905#-L87mSd70tXSxz0jkaVm:-L8ExL40BN4zl2D-lLsA:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cmstestsuite/unit_tests/server/contest/authentication_test.py#L182)):*

Can you please write a short description of what each of these assert method does?
</blockquote></details>

Done.

---

*[cmstestsuite/unit_tests/server/contest/authentication_test.py, line 48 at r19](https://beta.reviewable.io/reviews/cms-dev/cms/905#-L8EiPZ3E_uPYE1zL74t:-L8ExKJT49PleJv7ou4C:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/d9920040f9f661a707fbb8b24a0d9e329eb0b9ec/cmstestsuite/unit_tests/server/contest/authentication_test.py#L48)):*

You should change these names too
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cms-dev/cms/905#-:-L8EytHp-JOsoYBIrI2G:b78c8o9)*

stefano-maggiolo:
:lgtm:

---

Review status: 0 of 12 files reviewed at latest revision, 1 unresolved discussion.

---

*[cms/server/contest/authentication.py, line 244 at r2](https://reviewable.io:443/reviews/cms-dev/cms/905#-L876WQXd16hDWD6xiwq:-L8GmjUljuKNOMu3TEdK:b-zcjvte) ([raw file](https://github.com/cms-dev/cms/blob/a9462a38ffa2a1ee63a45f3b7da0a2fb9253a13f/cms/server/contest/authentication.py#L244)):*

I don't have a strong opinion in either direction. Let's keep it like it was and, as you said, if we decide to change the semantics we do it in another dedicated commit.

I have defined a dedicated exception to replace RuntimeError. It seemed the cleanest solution.
</blockquote></details>

Thanks

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cms-dev/cms/905#-:-L8GmxXrelgq69uNzhWO:bnfp4nl)*

lerks:
Merged

PR #918, opened 2018-03-27
Use MB instead of MiB throughout the code

lerks:
In general, prefer powers of 10 over powers of 2.

Moreover store the memory limit in bytes, rather than in MiB, to avoid any ambiguity.

This removes the inconsistencies of having the memory limit displayed in MiB in AWS and in MB in CWS.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/918)

### DIFF
--- a/cms/conf.py
+++ b/cms/conf.py
@@ -117,7 +117,7 @@ class Config(object):
         self.sandbox_implementation = 'isolate'
 
         # Sandbox.
-        self.max_file_size = 1048576
+        self.max_file_size = 1000 * 1000 * 1000
 
         # WebServers.
         self.secret_key_default = "8e045a51e4b102ea803c06f92841a1fb"
--- a/cms/db/__init__.py
+++ b/cms/db/__init__.py
@@ -89,7 +89,7 @@ __all__ = [
 
 # Instantiate or import these objects.
 
-version = 30
+version = 31
 
 engine = create_engine(config.database, echo=config.database_debug,
                        pool_timeout=60, pool_recycle=120)
--- a/cms/grading/Sandbox.py
+++ b/cms/grading/Sandbox.py
@@ -253,7 +253,7 @@ class SandboxBase(with_metaclass(ABCMeta, object)):
             time_str = "(time unknown)"
         memory_used = self.get_memory_used()
         if memory_used is not None:
-            mem_str = "%.2f MB" % (memory_used / (1024 * 1024))
+            mem_str = "%.2f MB" % (memory_used / (1000 * 1000))
         else:
             mem_str = "(memory usage unknown)"
         return "[%s - %s]" % (time_str, mem_str)
@@ -404,7 +404,7 @@ class SandboxBase(with_metaclass(ABCMeta, object)):
             file_ = Truncator(file_, trunc_len)
         return file_
 
-    def get_file_to_string(self, path, maxlen=1024):
+    def get_file_to_string(self, path, maxlen=1000):
         """Return the content of a file in the sandbox given its
         relative path.
 
@@ -729,12 +729,12 @@ class StupidSandbox(SandboxBase):
                                    (rlimit_cpu, rlimit_cpu))
 
             if self.address_space:
-                rlimit_data = int(self.address_space * 1024)
+                rlimit_data = self.address_space
                 resource.setrlimit(resource.RLIMIT_DATA,
                                    (rlimit_data, rlimit_data))
 
             if self.stack_space:
-                rlimit_stack = int(self.stack_space * 1024)
+                rlimit_stack = self.stack_space
                 resource.setrlimit(resource.RLIMIT_STACK,
                                    (rlimit_stack, rlimit_stack))
 
@@ -1059,12 +1059,14 @@ class IsolateSandbox(SandboxBase):
         if self.stdin_file is not None:
             res += ["--stdin=%s" % self.inner_absolute_path(self.stdin_file)]
         if self.stack_space is not None:
-            res += ["--stack=%d" % self.stack_space]
+            # Isolate wants memory values as KiB.
+            res += ["--stack=%d" % (self.stack_space // 1024)]
         if self.address_space is not None:
+            # Isolate wants memory values as KiB.
             if self.cgroup:
-                res += ["--cg-mem=%d" % self.address_space]
+                res += ["--cg-mem=%d" % (self.address_space // 1024)]
             else:
-                res += ["--mem=%d" % self.address_space]
+                res += ["--mem=%d" % (self.address_space // 1024)]
         if self.stdout_file is not None:
             res += ["--stdout=%s" % self.inner_absolute_path(self.stdout_file)]
         if self.max_processes is not None:
@@ -1141,6 +1143,7 @@ class IsolateSandbox(SandboxBase):
 
         """
         if 'cg-mem' in self.log:
+            # Isolate returns memory measurements in KiB.
             return int(self.log['cg-mem'][0]) * 1024
         return None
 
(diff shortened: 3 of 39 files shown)


stefano-maggiolo:
There are a couple of 1048576 too

---

Reviewed 38 of 38 files at r1.

---

*[cms/grading/tasktypes/Communication.py, line 262 at r1](https://beta.reviewable.io/reviews/cms-dev/cms/918#-L8d3VCa2a6tPl9gy8ox:-L8d3VCa2a6tPl9gy8oy:b-3ii1bf) ([raw file](https://github.com/cms-dev/cms/blob/a5c9c81be4dca1c64af95f3b4828470dec3f332b/cms/grading/tasktypes/Communication.py#L262)):*
> ```Python
>             if len(commands) > 1:
>                 evaluation_step(sandbox_user[i], commands[:-1], 10,
>                                 250 * 1000 * 1000)
> ```

Apparently in python3 you can do 250_000_000 :)

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cms-dev/cms/918#-:-L8d4kAXDMPuXeC8xr_7:bvyxej4)*

lerks:
Fixed the 1048576.

---

Review status: 37 of 40 files reviewed at latest revision, all discussions resolved.

---

*[cms/grading/tasktypes/Communication.py, line 262 at r1](https://beta.reviewable.io/reviews/cms-dev/cms/918#-L8d3VCa2a6tPl9gy8ox:-L8dB7CaE9_goBlJ7GcE:b8f28mv) ([raw file](https://github.com/cms-dev/cms/blob/a5c9c81be4dca1c64af95f3b4828470dec3f332b/cms/grading/tasktypes/Communication.py#L262)):*

Apparently in python3 you can do 250_000_000 :)
</blockquote></details>

Yes, but that is only true for 3.6+ and apparently it hasn't been backported to 2 (not even as a `__future__` import). I don't think putting comments everywhere is helpful. One day, after we will drop py2, we will remember about this and do it all at once.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cms-dev/cms/918#-:-L8dB4W20k_V8jYQgHEv:b7hhxr1)*

stefano-maggiolo:
Reviewed 2 of 3 files at r2.
Review status: 39 of 40 files reviewed at latest revision, all discussions resolved.

---

*[cms/conf.py, line 120 at r2](https://beta.reviewable.io/reviews/cms-dev/cms/918#-L8dCMH39myq-8ZQCJ1Q:-L8dCMH39myq-8ZQCJ1R:b7xw4oy) ([raw file](https://github.com/cms-dev/cms/blob/5a24c58fdf5fdf947e326d0319a9c55e10c12f4a/cms/conf.py#L120)):*
> ```Python
> 
>         # Sandbox.
>         self.max_file_size = 1000 * 1000 * 1000
> ```

The change of unit from k to non-k is a bit more dangerous because we don't have updater for cms.conf. Do you really dislike keeping it as KB?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cms-dev/cms/918)*

wil93:
Is there a reason not to keep MiB everywhere? 

IMHO the MB notation is not unambiguous, there are some cases where by MB you actually mean MiB, so it's better to use an unambiguous notation (which also happens to be more natural in computer science, because of powers of two)

@Dariost want to chime in?

stefano-maggiolo:
Judging from a couple of comments I've received over the years, at least a
couple of people were confused by MiB, or were worried that contestants
could get confused not knowing what the unit means.

dariost:
I agree that the inconsistencies between AWS and CWS of MiB vs MB should be resolved, but why use powers of 10 instead of powers of 2?
Most systems use the latter, people are used to the power of 2 meaning and K, M, G prefixes are ambiguous, since multiple standards (SI and JEDEC) define them in different ways.
Wouldn't be better to change the occurrences of KB, MB and GB to KiB, MiB and GiB to remove inconsistencies and be unambiguous about their meaning rather than using powers of 10?

wil93:
I never heard of a contestant confused by this. And we always used MiB in cmsocial (over 5k users and no one was confused)

Worst case, some user will interpret MiB as MB. But even when that happens, it's ok because they will just get slightly more memory and they will be happier

stefano-maggiolo:
I honestly don't have strong opinions myself, apart from using the same
consistently, I was just reporting the comments I received (which are far
from many). I'm happy with -bibytes too, maybe with a tooltip in cws to be
safe.

lerks:
> Is there a reason not to keep MiB everywhere?

Yes: Babel doesn't support it.

We can't assume all admins and contestants know the difference between MB and MiB. Even if we, who know it, managed to consistently use the correct one (spoiler: so far we didn't) we would have that people who know the difference between MiB and MB would be fine either way and people who don't would either not understand or misinterpret MiB. I think sticking to MB is a safer option.

Also, is there any real reason to use for bytes a different system of prefixes than for all other units?

wil93:
> Yes: Babel doesn't support it.

So we make a design choice based on a limitation of one of our dependencies? Surely there is a way around it (we could patch and send a PR to babel)

> Also, is there any real reason to use for bytes a different system of prefixes than for all other units?

I guess this is one of those fundamental questions like "why bytes are 8 bits and not 10?" :laughing: 

Anyway I think sticking to MB is not safe for the simple reason of not knowing, when you read `Memory limit 256 MB`, if the memory limit is 256 * 1000 * 1000 bytes or 256 * 1024 * 1000 bytes or 256 * 1024 * 1024 bytes (all are valid in some context)

wil93:
> I'm happy with -bibytes too, maybe with a tooltip in cws to be
safe.

@stefano-maggiolo I like the tooltip idea, and for the record this is the tooltip we use (and that successfully didn't confuse our users) :smile: 

screenshot from 2018-03-28 00-14-05

lerks:
> So we make a design choice based on a limitation of one of our dependencies? Surely there is a way around it (we could patch and send a PR to babel)

I wanted to open an issue, but if you feel like patching it go for it! (BTW, you need to patch CLDR, that's where Babel gets its data from). If you manage then, as soon as Ubuntu 20.04 comes out, we can switch back to MiB. :P

wil93:
> you need to patch CLDR, that's where Babel gets its data from). If you manage then, as soon as Ubuntu 20.04 comes out, we can switch back to MiB. :P

You give up too easily :P we can definitely find a workaround... The point is: is there a reason for *designing* the system to think in MB and not MiB? I can't really see it

andreyv:
> We would have that people who know the difference between MiB and MB would be fine either way and people who don't would either not understand or misinterpret MiB.

I think this goes backwards. MiB is unambiguous, meanwhile for MB even power users are never sure what does it mean, since so much software uses "MB" incorrectly.

> Also, is there any real reason to use for bytes a different system of prefixes than for all other units?

Yes, many other constants in computing are powers of two. Integer sizes, RAM bank size, virtual memory page size. Using the binary prefix everywhere allows for easier comparisons and estimations.

On the other hand, I just checked Codeforces, Codechef, COCI and SPOJ, and they all use decimal prefixes (of course, I don't know whether they really are decimal).

andreyv:
It would also be interesting to know what happens when the memory limit is not divisible by page size. Does the user get slightly more memory?

CC @gollux

gollux:
> It would also be interesting to know what happens when the memory limit is not divisible by page size. Does the user get slightly more memory?

It depends on rounding in the kernel. The current mainline kernel rounds down,
so the process gets slightly less memory.

lerks:
The unification of the MB-in-CWS vs. MiB-in-AWS issue has been achieved in #1025, which I just merged. The plan to use of powers of ten instead of powers of 2 has been abandoned. I'm still considering storing the memory limit in bytes rather than MiB but after the feedback received here I will do it in a different way. Thus this PR is of no use anymore.

PR #922, opened 2018-04-03
Complete the transition to using the Dataset's cached task & score type objects

lerks:
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/922)


stefano-maggiolo:
:lgtm:

---

Reviewed 10 of 10 files at r1, 1 of 1 files at r2, 3 of 3 files at r3.

---

*[cms/db/task.py, line 462 at r1](https://beta.reviewable.io/reviews/cms-dev/cms/922#-L9MJJ7B-fFCJ5aJlehQ:-L9MJJ7B-fFCJ5aJlehR:b-sc23vs) ([raw file](https://github.com/cms-dev/cms/blob/db8aba17d7a22da06ffc1477a60617c59346c687/cms/db/task.py#L462)):*
> ```Python
>     @property
>     def score_type_object(self):
>         public_testcases = {k: tc.public for k, tc in iteritems(self.testcases)}
> ```

It's pretty remote, but I think you want to make sure the order doesn't change too.

---

*[cms/server/jinja2_toolbox.py, line 182 at r1](https://beta.reviewable.io/reviews/cms-dev/cms/922#-L9MJrUUB097blI7cCez:-L9MJrUUB097blI7cCf-:ba9jwp6) ([raw file](https://github.com/cms-dev/cms/blob/db8aba17d7a22da06ffc1477a60617c59346c687/cms/server/jinja2_toolbox.py#L182)):*
> ```Python
>         return dataset.task_type_object
>     # The task type's constructor is called, which may raise any
>     # arbitrary exception, hence we stay as general al possible.
> ```

Fix typo as you're at it?

---

*[cms/server/jinja2_toolbox.py, line 194 at r1](https://beta.reviewable.io/reviews/cms-dev/cms/922#-L9MJuvZEQ_JJOadL_y6:-L9MJuvZEQ_JJOadL_y7:b-89a4l7) ([raw file](https://github.com/cms-dev/cms/blob/db8aba17d7a22da06ffc1477a60617c59346c687/cms/server/jinja2_toolbox.py#L194)):*
> ```Python
>         return dataset.score_type_object
>     # The score type's constructor is called, which may raise any
>     # arbitrary exception, hence we stay as general al possible.
> ```

Ditto

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cms-dev/cms/922#-:-L9MPzkJ16DAZHfufIr9:bnfp4nl)*

lerks:
Merged.

PR #852, opened 2018-02-22
[IOI2017] merge back changes related to TPS

wil93:
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/852)

PR #935, opened 2018-04-12
TPS importer

stefano-maggiolo:
Supersedes #852. FYI @wil93.
Tried on two tasks kindly provided by @akmohtashami and it imports correctly.

Main changes introduced by me:
- fixed pep8 & change of names within the existing commits (this required the extraction of the long "task_type_parameter_<name>" strings, the other changes are trivial);
- included at the end the update from the removal of submission format element.

Note for reviewer (@lerks): more subtle code style decisions should belong to TPS author, IMHO. 

The last remaining issue is that the importer wants to use the task types Batch2017 and Communication2017, which is not very future-proof. I guess the question for @akmohtashami is whether the tasks exported from TPS explicitly require these task types, or they can work with the regular version too. In the latter case I'd remove the 2017 suffix, in the former we need to find a solution to ensure compatibility.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/935)

### DIFF
--- a/cmscontrib/loaders/__init__.py
+++ b/cmscontrib/loaders/__init__.py
@@ -26,10 +26,14 @@ from six import itervalues, iteritems
 
 from .italy_yaml import YamlLoader
 from .polygon import PolygonTaskLoader, PolygonUserLoader, PolygonContestLoader
+from .tps import TpsTaskLoader
+
 
 LOADERS = dict(
     (loader_class.short_name, loader_class) for loader_class in [
-        YamlLoader, PolygonTaskLoader, PolygonUserLoader, PolygonContestLoader
+        YamlLoader,
+        PolygonTaskLoader, PolygonUserLoader, PolygonContestLoader,
+        TpsTaskLoader
     ]
 )
 
--- a/cmscontrib/loaders/tps.py
+++ b/cmscontrib/loaders/tps.py
@@ -31,6 +31,7 @@ import json
 import logging
 import os
 import re
+import subprocess
 
 from datetime import timedelta
 
@@ -239,8 +240,10 @@ class TpsTaskLoader(TaskLoader):
         if os.path.exists(checker_src):
             logger.info("Checker found, compiling")
             checker_exe = os.path.join(checker_dir, "checker")
-            os.system("g++ -x c++ -std=gnu++14 -O2 -static -o %s %s" %
-                      (checker_exe, checker_src))
+            subprocess.call([
+                "g++", "-x", "c++", "-std=gnu++14", "-O2", "-static",
+                "-o", checker_exe, checker_src
+            ])
             digest = self.file_cacher.put_file_from_path(
                 checker_exe,
                 "Manager for task %s" % name)
@@ -290,9 +293,10 @@ class TpsTaskLoader(TaskLoader):
         if os.path.exists(manager_src):
             logger.info("Manager found, compiling")
             manager_exe = os.path.join(graders_dir, "manager")
-            os.system("cat %s | \
-                            g++ -x c++ -O2 -static -o %s -" %
-                      (manager_src, manager_exe))
+            subprocess.call([
+                "g++", "-x", "c++", "-O2", "-static",
+                "-o", manager_exe, manager_src
+            ])
             digest = self.file_cacher.put_file_from_path(
                 manager_exe,
                 "Manager for task %s" % name)
--- a/cmscontrib/loaders/tps.py
+++ b/cmscontrib/loaders/tps.py
@@ -102,7 +102,6 @@ class TpsTaskLoader(TaskLoader):
                 [task_type_parameters[par_input],
                  task_type_parameters[par_output]],
                 evaluation_param,
-                task_type_parameters[par_user_managers]
             ]
 
         if task_type == 'Communication':
@@ -253,10 +252,9 @@ class TpsTaskLoader(TaskLoader):
             logger.info("Checker not found, using diff if necessary")
             evaluation_param = "diff"
 
+        # Note that the original TPS worked with custom task type Batch2017
+        # and Communication2017 instead of Batch and Communication.
         args["task_type"] = data['task_type']
-        if data['task_type'] != 'OutputOnly' \
-                and data['task_type'] != 'Notice':
-            args["task_type"] += '2017'
         args["task_type_parameters"] = \
             self._get_task_type_parameters(
                 data, data['task_type'], evaluation_param)
@@ -285,6 +283,9 @@ class TpsTaskLoader(TaskLoader):
             digest = self.file_cacher.put_file_from_path(
                 grader_src,
                 "Manager for task %s" % name)
+            if data['task_type'] == 'Communication' \
+                    and os.path.splitext(grader_name)[0] == 'grader':
+                grader_name = 'stub' + os.path.splitext(grader_name)[1]
             args["managers"][grader_name] = Manager(grader_name, digest)
 
         # Manager
--- a/cmscontrib/loaders/tps.py
+++ b/cmscontrib/loaders/tps.py
@@ -3,6 +3,7 @@
 
 # Programming contest management system
 # Copyright © 2017 Kiarash Golezardi <kiarashgolezardi@gmail.com>
+# Copyright © 2017 Amir Keivan Mohtashami <akmohtashami97@gmail.com>
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as
@@ -28,10 +29,12 @@ import io
 import json
 import logging
 import os
+import re
 
 from datetime import timedelta
 
-from cms.db import Task, SubmissionFormatElement, Dataset, Manager, Testcase
+from cms.db import Task, SubmissionFormatElement, Dataset, Manager, \
+    Testcase, Attachment
 
 from .base_loader import TaskLoader
 
@@ -51,7 +54,6 @@ class TpsTaskLoader(TaskLoader):
     short_name = 'tps_task'
     description = 'TPS task format'
 
-    # FIXME: stay?
     @staticmethod
     def detect(path):
         """See docstring in class Loader.
@@ -76,7 +78,7 @@ class TpsTaskLoader(TaskLoader):
             par_input = '%s_io_0_inputfile' % par_prefix
             par_output = '%s_io_1_outputfile' % par_prefix
             if par_compilation not in task_type_parameters:
-                task_type_parameters[par_compilation] = 'alone'
+                task_type_parameters[par_compilation] = 'grader'
             if par_input not in task_type_parameters:
                 task_type_parameters[par_input] = ''
             if par_output not in task_type_parameters:
@@ -115,7 +117,32 @@ class TpsTaskLoader(TaskLoader):
 
         # TODO: import statements
 
-        args["submission_format"] = [SubmissionFormatElement("%s.%%l" % name)]
+        # Setting the submission format
+        # Obtaining testcases' codename
+        testcases_dir = os.path.join(self.path, 'tests')
+        testcase_codenames = [filename[:-3]
+                              for filename in os.listdir(testcases_dir)
+                              if filename[-3:] == '.in']
+        if data["task_type"] == 'OutputOnly':
+            args["submission_format"] = list()
+            for codename in testcase_codenames:
+                args["submission_format"] += \
+                    SubmissionFormatElement("output_%s.txt" % codename)
+        elif data["task_type"] == 'Notice':
+            args["submission_format"] = list()
+        else:
+            args["submission_format"] = [
+                SubmissionFormatElement("%s.%%l" % name)]
+
+        # Attachments
+        args["attachments"] = dict()
+        attachments_dir = os.path.join(self.path, 'attachments')
+        if os.path.exists(attachments_dir):
+            for filename in os.listdir(attachments_dir):
+                digest = self.file_cacher.put_file_from_path(
+                    os.path.join(attachments_dir, filename),
+                    "Attachment %s for task %s" % (filename, name))
+                args["attachments"][filename] = Attachment(filename, digest)
 
         # These options cannot be configured in the TPS format.
         # Uncomment the following to set specific values for them.
@@ -130,6 +157,15 @@ class TpsTaskLoader(TaskLoader):
         # args['token_gen_number'] = 1
         # args['token_gen_interval'] = make_timedelta(1800)
         # args['token_gen_max'] = 2
+
+        args['max_submission_number'] = 50
+        args['max_user_test_number'] = 50
+        if data["task_type"] == 'OutputOnly':
+            args['max_submission_number'] = 100
+            args['max_user_test_number'] = 100
+
+        args['min_submission_interval'] = make_timedelta(60)
+        args['min_user_test_interval'] = make_timedelta(60)
 
         # TODO: additional cms config files
 
@@ -153,9 +189,8 @@ class TpsTaskLoader(TaskLoader):
         if os.path.exists(checker_src):
             logger.info("Checker found, compiling")
             checker_exe = os.path.join(checker_dir, "checker")
-            os.system("cat %s | \
-                g++ -x c++ -O2 -static -o %s -" %
-                      (checker_src, checker_exe))
+            os.system("g++ -x c++ -O2 -static -o %s %s" %
+                      (checker_exe, checker_src))
             digest = self.file_cacher.put_file_from_path(
                 checker_exe,
                 "Manager for task %s" % name)
@@ -166,6 +201,9 @@ class TpsTaskLoader(TaskLoader):
             evaluation_param = "diff"
 
         args["task_type"] = data['task_type']
+        if data['task_type'] != 'OutputOnly' \
+                and data['task_type'] != 'Notice':
+            args["task_type"] += '2017'
         args["task_type_parameters"] = \
             self._get_task_type_parameters(
                 data, data['task_type'], evaluation_param)
@@ -195,14 +233,9 @@ class TpsTaskLoader(TaskLoader):
             digest = self.file_cacher.put_file_from_path(
                 manager_exe,
                 "Manager for task %s" % name)
-            args["managers"] += [Manager("manager", digest)]
+            args["managers"]["manager"] = Manager("manager", digest)
 
         # Testcases
-        testcases_dir = os.path.join(self.path, 'tests')
-        testcase_codenames = [filename[:-3]
-                              for filename in os.listdir(testcases_dir)
-                              if filename[-3:] == '.in']
-
         args["testcases"] = {}
 
         for codename in testcase_codenames:
@@ -226,7 +259,7 @@ class TpsTaskLoader(TaskLoader):
 
         # Score Type
         subtasks_dir = os.path.join(self.path, 'subtasks')
-        subtasks = os.listdir(subtasks_dir)
+        subtasks = sorted(os.listdir(subtasks_dir))
 
         if len(subtasks) == 0:
             number_tests = len(testcase_codenames)
@@ -234,7 +267,18 @@ class TpsTaskLoader(TaskLoader):
             args["score_type_parameters"] = str(100 / number_tests)
         else:
             args["score_type"] = "GroupMin"
-            # FIXME: create the score_type_parameters
+            parsed_data = []
+            for subtask in subtasks:
+                with io.open(os.path.join(subtasks_dir, subtask), 'rt',
+                             encoding='utf-8') as subtask_json:
+                    subtask_data = json.load(subtask_json)
+                    score = int(subtask_data["score"])
+                    testcases = "|".join(
+                        re.escape(testcase)
+                        for testcase in subtask_data["testcases"]
+                    )
+                    parsed_data.append([score, testcases])
+            args["score_type_parameters"] = json.dumps(parsed_data)
 
         dataset = Dataset(**args)
         task.active_dataset = dataset
--- a/cmscontrib/loaders/tps.py
+++ b/cmscontrib/loaders/tps.py
@@ -34,7 +34,7 @@ import re
 from datetime import timedelta
 
 from cms.db import Task, SubmissionFormatElement, Dataset, Manager, \
-    Testcase, Attachment
+    Testcase, Attachment, Statement
 
 from .base_loader import TaskLoader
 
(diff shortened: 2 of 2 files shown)


lerks:
I'm against keeping the Notice and the Foo2017 TaskTypes.

---

Reviewed 2 of 2 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, 1 of 1 files at r5.

---

*[cmscontrib/loaders/__init__.py, line 30 at r1](https://beta.reviewable.io/reviews/cms-dev/cms/935#-LADdps-C2Gr1m-Xrdhk:-LADdps-C2Gr1m-Xrdhl:b-fa7d52) ([raw file](https://github.com/cms-dev/cms/blob/dd8e67f4baa9f84a35dd22d02178a8e04c6d3920/cmscontrib/loaders/__init__.py#L30)):*
> ```Python
> from .polygon import PolygonTaskLoader, PolygonUserLoader, PolygonContestLoader
> from .tps import TpsTaskLoader
> ```

Nit: two blank lines?

---

*[cmscontrib/loaders/tps.py, line 1 at r1](https://beta.reviewable.io/reviews/cms-dev/cms/935#-LADdvg_38A8f0Gf2mZm:-LADdvga3AVmjB2Lscjb:bibt9jt) ([raw file](https://github.com/cms-dev/cms/blob/dd8e67f4baa9f84a35dd22d02178a8e04c6d3920/cmscontrib/loaders/tps.py#L1)):*
> ```Python
> #!/usr/bin/env python2
> ```

s/python2/python/

---

*[cmscontrib/loaders/tps.py, line 22 at r1](https://beta.reviewable.io/reviews/cms-dev/cms/935#-LADdy8qEIPRF_97MvYd:-LADdy8qEIPRF_97MvYe:b-6ony3h) ([raw file](https://github.com/cms-dev/cms/blob/dd8e67f4baa9f84a35dd22d02178a8e04c6d3920/cmscontrib/loaders/tps.py#L22)):*
> ```Python
> from __future__ import absolute_import
> from __future__ import print_function
> from __future__ import unicode_literals
> ```

Add `division` (keep sorted lexicographically) and `from future... import *  # noqa` imports

---

*[cmscontrib/loaders/tps.py, line 32 at r1](https://beta.reviewable.io/reviews/cms-dev/cms/935#-LADnKwH76otDZp4Q4LJ:-LADnKwH76otDZp4Q4LK:b-gr68sq) ([raw file](https://github.com/cms-dev/cms/blob/dd8e67f4baa9f84a35dd22d02178a8e04c6d3920/cmscontrib/loaders/tps.py#L32)):*
> ```Python
> from cms.db import Task, SubmissionFormatElement, Dataset, Manager, Testcase
> from cmscontrib import touch
> ```

Keep blank line

---

*[cmscontrib/loaders/tps.py, line 68 at r1](https://beta.reviewable.io/reviews/cms-dev/cms/935#-LADkZLgDWVYeUI6ZBKi:-LADkZLgDWVYeUI6ZBKj:b-gyz6y3) ([raw file](https://github.com/cms-dev/cms/blob/dd8e67f4baa9f84a35dd22d02178a8e04c6d3920/cmscontrib/loaders/tps.py#L68)):*
> ```Python
>         parameters_str = data['task_type_params']
>         if parameters_str is None or \
>                 parameters_str == '':
> ```

No need to break the line.

---

*[cmscontrib/loaders/tps.py, line 82 at r1](https://beta.reviewable.io/reviews/cms-dev/cms/935#-LADlDUv-HeRnRqiMpKE:-LADlDUv-HeRnRqiMpKF:bpzkye9) ([raw file](https://github.com/cms-dev/cms/blob/dd8e67f4baa9f84a35dd22d02178a8e04c6d3920/cmscontrib/loaders/tps.py#L82)):*
> ```Python
>             if par_output not in task_type_parameters:
>                 task_type_parameters[par_output] = ''
>             return '["%s", ["%s", "%s"], "%s"]' % \
> ```

Encoding as a string is now wrong, since the DB stores JSON (also x2 below)

---

*[cmscontrib/loaders/tps.py, line 101 at r1](https://beta.reviewable.io/reviews/cms-dev/cms/935#-LADfuYdD1Z1T4UXdWfL:-LADfuYdD1Z1T4UXdWfM:bksic04) ([raw file](https://github.com/cms-dev/cms/blob/dd8e67f4baa9f84a35dd22d02178a8e04c6d3920/cmscontrib/loaders/tps.py#L101)):*
> ```Python
>         json_src = os.path.join(self.path, 'problem.json')
>         if not os.path.exists(json_src):
>             logger.error('No task found.')
> ```

Also return/raise exception? Otherwise the following open call will fail anyways with a not found message and this will be useless...

---

*[cmscontrib/loaders/tps.py, line 104 at r1](https://beta.reviewable.io/reviews/cms-dev/cms/935#-LADfXTp7CBxVqPXJ_cz:-LADfXTp7CBxVqPXJ_d-:b-9kmi15) ([raw file](https://github.com/cms-dev/cms/blob/dd8e67f4baa9f84a35dd22d02178a8e04c6d3920/cmscontrib/loaders/tps.py#L104)):*
> ```Python
> 
>         json_src = os.path.join(self.path, 'problem.json')
>         with open(json_src) as json_file:
> ```

Use `io.open`, mode `rt` and encoding utf-8.

---

*[cmscontrib/loaders/tps.py, line 115 at r1](https://beta.reviewable.io/reviews/cms-dev/cms/935#-LADhCV82WoILCBl84RT:-LADhCV82WoILCBl84RU:b-i7pg9u) ([raw file](https://github.com/cms-dev/cms/blob/dd8e67f4baa9f84a35dd22d02178a8e04c6d3920/cmscontrib/loaders/tps.py#L115)):*
> ```Python
>         touch(os.path.join(self.path, ".itime"))
>         # If this file is not deleted, then the import failed.
>         touch(os.path.join(self.path, ".import_error"))
> ```

Why are these two needed? I see that italy_yaml uses them and I guess this was just copied-and-pasted (polygon seems to have done it as well).

---

*[cmscontrib/loaders/tps.py, line 167 at r1](https://beta.reviewable.io/reviews/cms-dev/cms/935#-LADjs63ETisQrvPRwWF:-LADjs63ETisQrvPRwWG:b3p70ie) ([raw file](https://github.com/cms-dev/cms/blob/dd8e67f4baa9f84a35dd22d02178a8e04c6d3920/cmscontrib/loaders/tps.py#L167)):*
> ```Python
>             os.system("cat %s | \
>                 g++ -x c++ -O2 -static -o %s -" %
>                       (checker_src, checker_exe))
> ```

`os.system` is sort-of deprecated in favor of `subprocess`. Also, using a list of tokens avoids the need for the shell to parse all of this (hence removing the risk induced by the lack of escaping).

https://docs.python.org/3/library/subprocess.html#replacing-os-system

Same below for manager

---

*[cmscontrib/loaders/tps.py, line 251 at r1](https://beta.reviewable.io/reviews/cms-dev/cms/935#-LADmyro3BNltIWF554e:-LADmyro3BNltIWF554f:b-q64pkt) ([raw file](https://github.com/cms-dev/cms/blob/dd8e67f4baa9f84a35dd22d02178a8e04c6d3920/cmscontrib/loaders/tps.py#L251)):*
> ```Python
>         task.active_dataset = dataset
> 
>         os.remove(os.path.join(self.path, ".import_error"))
> ```

See above.

---

*[cmscontrib/loaders/tps.py, line 28 at r2](https://beta.reviewable.io/reviews/cms-dev/cms/935#-LADnD1F6cGqebaTxbTF:-LADnD1F6cGqebaTxbTG:bbl23vf) ([raw file](https://github.com/cms-dev/cms/blob/e2aff7387773f3acc6636e09c2f6e7553ce7bc31/cmscontrib/loaders/tps.py#L28)):*
> ```Python
> import logging
> import os
> import re
> ```

Keep sorted.

---

*[cmscontrib/loaders/tps.py, line 153 at r2](https://beta.reviewable.io/reviews/cms-dev/cms/935#-LADo3RBBwqwNns91Aug:-LADo3RBBwqwNns91Auh:b4oam0s) ([raw file](https://github.com/cms-dev/cms/blob/e2aff7387773f3acc6636e09c2f6e7553ce7bc31/cmscontrib/loaders/tps.py#L153)):*
> ```Python
>         # Uncomment the following to set specific values for them.
> 
>         # args['max_submission_number'] = 100
> ```

Remove these four since now they're actually set below.

---

*[cmscontrib/loaders/tps.py, line 240 at r2](https://beta.reviewable.io/reviews/cms-dev/cms/935#-LADoKnc4ES3yXYIvav1:-LADoKnc4ES3yXYIvav2:b-qjfekp) ([raw file](https://github.com/cms-dev/cms/blob/e2aff7387773f3acc6636e09c2f6e7553ce7bc31/cmscontrib/loaders/tps.py#L240)):*
> ```Python
>             os.system("cat %s | \
>                             g++ -x c++ -O2 -static -o %s -" %
>                       (manager_src, manager_exe))
> ```

Moreover, the filename can be passed as an argument to g++, as has been done above for the checker.

---

*[cmscontrib/loaders/tps.py, line 280 at r2](https://beta.reviewable.io/reviews/cms-dev/cms/935#-LADonCKEaM0ZmqQTzP7:-LADonCKEaM0ZmqQTzP8:brmshfm) ([raw file](https://github.com/cms-dev/cms/blob/e2aff7387773f3acc6636e09c2f6e7553ce7bc31/cmscontrib/loaders/tps.py#L280)):*
> ```Python
>             parsed_data = []
>             for subtask in subtasks:
>                 with open(os.path.join(subtasks_dir, subtask)) \
> ```

Use `io.open` + args

---

*[cmscontrib/loaders/tps.py, line 289 at r2](https://beta.reviewable.io/reviews/cms-dev/cms/935#-LADpJQV7sGIXxEeztqd:-LADpJQV7sGIXxEeztqe:blmuuzm) ([raw file](https://github.com/cms-dev/cms/blob/e2aff7387773f3acc6636e09c2f6e7553ce7bc31/cmscontrib/loaders/tps.py#L289)):*
> ```Python
>                     )
>                     parsed_data.append([score, testcases])
>             args["score_type_parameters"] = json.dumps(parsed_data)
> ```

Again, no need to dump to JSON.

---

*[cmscontrib/loaders/tps.py, line 152 at r3](https://beta.reviewable.io/reviews/cms-dev/cms/935#-LADqMT52PQFvON97_qT:-LADqMT52PQFvON97_qU:bq1vz3g) ([raw file](https://github.com/cms-dev/cms/blob/6a5d2810813d799de45cfa3a2a6df94cde68c455/cmscontrib/loaders/tps.py#L152)):*
> ```Python
>                     language = statement[:-4]
>                     if language == "en_US":
>                         args["primary_statements"] = '["en_US"]'
> ```

No need to JSON-encode: can keep as list-of-strings

---

*[cmscontrib/loaders/tps.py, line 313 at r3](https://beta.reviewable.io/reviews/cms-dev/cms/935#-LADrOSf9PQLdhP_dZkF:-LADrOSf9PQLdhP_dZkG:bqnfmqs) ([raw file](https://github.com/cms-dev/cms/blob/6a5d2810813d799de45cfa3a2a6df94cde68c455/cmscontrib/loaders/tps.py#L313)):*
> ```Python
>             if not os.path.exists(outfile):
>                 logger.critical(
>                     'Could not file the output file for testcase %s'
> ```

s/file/find/

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cms-dev/cms/935#-:-LADoWm6731VENj8KM-3:bhluetl)*

lerks:
Reviewed 2 of 2 files at r6, 1 of 1 files at r7, 1 of 1 files at r8, 1 of 1 files at r9, 1 of 1 files at r10, 1 of 1 files at r11, 1 of 1 files at r12.
Review status: all files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[cmscontrib/loaders/tps.py, line 32 at r1](https://beta.reviewable.io/reviews/cms-dev/cms/935#-LADnKwH76otDZp4Q4LJ:-LAFPRnVBzj7GsrRgCiS:bj695u5) ([raw file](https://github.com/cms-dev/cms/blob/dd8e67f4baa9f84a35dd22d02178a8e04c6d3920/cmscontrib/loaders/tps.py#L32)):*

Done.
</blockquote></details>

I only meant the one between absolute and relative, but never mind

---

*[cmscontrib/loaders/tps.py, line 82 at r1](https://beta.reviewable.io/reviews/cms-dev/cms/935#-LADlDUv-HeRnRqiMpKE:-LAFRUP80b06gZGKHLlA:bq3d13u) ([raw file](https://github.com/cms-dev/cms/blob/dd8e67f4baa9f84a35dd22d02178a8e04c6d3920/cmscontrib/loaders/tps.py#L82)):*

Uhm, how would it work then? Alchemy is nice and accepts both?
</blockquote></details>

Not sure. Either alchemy (or even postgres) automatically decode when you give a string for a JSON field or (IMO more likely) they store the string as the JSON structure (the import would succeed, but using the task would fail).

---

*[cmscontrib/loaders/tps.py, line 115 at r1](https://beta.reviewable.io/reviews/cms-dev/cms/935#-LADhCV82WoILCBl84RT:-LAFSJzr8yZQZ0Ej8uIX:b-z0yf1y) ([raw file](https://github.com/cms-dev/cms/blob/dd8e67f4baa9f84a35dd22d02178a8e04c6d3920/cmscontrib/loaders/tps.py#L115)):*

I think they are the way to understand if a reimport is required.
</blockquote></details>

They are in italy_yaml, where the loader reads back and checks these values. Here they're only just written to, hence definitely useless. (hint: search in the whole codebase for the strings "itime" and "import_error": you'll find them only in italy_yaml plus some copy-and-paste relicts like these in polygon).

---

*[cmscontrib/loaders/tps.py, line 28 at r2](https://beta.reviewable.io/reviews/cms-dev/cms/935#-LADnD1F6cGqebaTxbTF:-LAFPKqu521NmiR_JwKo:b-vwsc43) ([raw file](https://github.com/cms-dev/cms/blob/e2aff7387773f3acc6636e09c2f6e7553ce7bc31/cmscontrib/loaders/tps.py#L28)):*

Aren't they?
</blockquote></details>

I'm stupid

---

*[cmscontrib/loaders/tps.py, line 339 at r8](https://beta.reviewable.io/reviews/cms-dev/cms/935#-LAFQBZy-NOPGe3MT9ys:-LAFQBZy-NOPGe3MT9yt:b-cdoz9f) ([raw file](https://github.com/cms-dev/cms/blob/895708e5bf56594a41d83e4539e40dbc27849fd2/cmscontrib/loaders/tps.py#L339)):*
> ```Python
>             number_tests = max(len(testcase_codenames), 1)
>             args["score_type"] = "Sum"
>             args["score_type_parameters"] = str(100 / number_tests)
> ```

Can (should) be kept a float, no cast to str

---

*[cmscontrib/loaders/tps.py, line 264 at r10](https://beta.reviewable.io/reviews/cms-dev/cms/935#-LAFR7mL9rFA6jMoTq0W:-LAFR7mL9rFA6jMoTq0X:b-xulwrs) ([raw file](https://github.com/cms-dev/cms/blob/7d09f4c739db21e75806eddec48ef83d899e791c/cmscontrib/loaders/tps.py#L264)):*
> ```Python
>         args["task_type"] = data['task_type']
>         if data['task_type'] != 'OutputOnly' \
>                 and data['task_type'] != 'Notice':
> ```

There are at least 2 other mentions of Notice: should they be removed too?

---

*[cmscontrib/loaders/tps.py, line 119 at r12](https://beta.reviewable.io/reviews/cms-dev/cms/935#-LAFRR4d48JTJztTOiXP:-LAFRR4d48JTJztTOiXQ:b3ib2sm) ([raw file](https://github.com/cms-dev/cms/blob/33da5975470f5fcf9c576f75a896ba0a61e885f8/cmscontrib/loaders/tps.py#L119)):*
> ```Python
>             return [evaluation_param]
> 
>         return ""
> ```

Should probably be empty list?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cms-dev/cms/935)*

stefano-maggiolo:
Review status: all files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[cmscontrib/loaders/tps.py, line 32 at r1](https://beta.reviewable.io/reviews/cms-dev/cms/935#-LADnKwH76otDZp4Q4LJ:-LAHagWA1KIMv59C0LfP:b-qgqm0p) ([raw file](https://github.com/cms-dev/cms/blob/dd8e67f4baa9f84a35dd22d02178a8e04c6d3920/cmscontrib/loaders/tps.py#L32)):*

I only meant the one between absolute and relative, but never mind
</blockquote></details>

Ack. I'll keep that in mind :D

---

*[cmscontrib/loaders/tps.py, line 82 at r1](https://beta.reviewable.io/reviews/cms-dev/cms/935#-LADlDUv-HeRnRqiMpKE:-LAHamsP7j1P19WWN8Lz:b-jnzb32) ([raw file](https://github.com/cms-dev/cms/blob/dd8e67f4baa9f84a35dd22d02178a8e04c6d3920/cmscontrib/loaders/tps.py#L82)):*

Not sure. Either alchemy (or even postgres) automatically decode when you give a string for a JSON field or (IMO more likely) they store the string as the JSON structure (the import would succeed, but using the task would fail).
</blockquote></details>

Oh right, at the time I only checked visually and it looked fine, but probably I was tricked myself.

---

*[cmscontrib/loaders/tps.py, line 115 at r1](https://beta.reviewable.io/reviews/cms-dev/cms/935#-LADhCV82WoILCBl84RT:-LAHb3eh2U7CqfkTnaDj:b757wsf) ([raw file](https://github.com/cms-dev/cms/blob/dd8e67f4baa9f84a35dd22d02178a8e04c6d3920/cmscontrib/loaders/tps.py#L115)):*

They are in italy_yaml, where the loader reads back and checks these values. Here they're only just written to, hence definitely useless. (hint: search in the whole codebase for the strings "itime" and "import_error": you'll find them only in italy_yaml plus some copy-and-paste relicts like these in polygon).
</blockquote></details>

Uhm, ok. Should remove from here then, I guess.

---

*[cmscontrib/loaders/tps.py, line 251 at r1](https://beta.reviewable.io/reviews/cms-dev/cms/935#-LADmyro3BNltIWF554e:-LAHe8Ri0qNZhQA_EzUV:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/dd8e67f4baa9f84a35dd22d02178a8e04c6d3920/cmscontrib/loaders/tps.py#L251)):*

.
</blockquote></details>

Done.

---

*[cmscontrib/loaders/tps.py, line 339 at r8](https://beta.reviewable.io/reviews/cms-dev/cms/935#-LAFQBZy-NOPGe3MT9ys:-LAHdOM21DTnsROMgwHd:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/895708e5bf56594a41d83e4539e40dbc27849fd2/cmscontrib/loaders/tps.py#L339)):*

Can (should) be kept a float, no cast to str
</blockquote></details>

Done.

---

*[cmscontrib/loaders/tps.py, line 264 at r10](https://beta.reviewable.io/reviews/cms-dev/cms/935#-LAFR7mL9rFA6jMoTq0W:-LAHdqfDDT-mi6vWzV2O:b6k6wrj) ([raw file](https://github.com/cms-dev/cms/blob/7d09f4c739db21e75806eddec48ef83d899e791c/cmscontrib/loaders/tps.py#L264)):*

There are at least 2 other mentions of Notice: should they be removed too?
</blockquote></details>

No, here I removed it because it wasn't needed anymore after the removal of 2017 (which we might revisit), but if the format produces a Notice, I don't see why not giving it the correct parameters.

---

*[cmscontrib/loaders/tps.py, line 119 at r12](https://beta.reviewable.io/reviews/cms-dev/cms/935#-LAFRR4d48JTJztTOiXP:-LAHazre8kvCR0fEkgyY:b-p33006) ([raw file](https://github.com/cms-dev/cms/blob/33da5975470f5fcf9c576f75a896ba0a61e885f8/cmscontrib/loaders/tps.py#L119)):*

Should probably be empty list?
</blockquote></details>

Well, an empty list is as arbitrary as an empty string, but sure

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cms-dev/cms/935)*

lerks:
:lgtm:

---

Reviewed 1 of 1 files at r13, 1 of 1 files at r14.

---

*[cmscontrib/loaders/tps.py, line 264 at r10](https://beta.reviewable.io/reviews/cms-dev/cms/935#-LAFR7mL9rFA6jMoTq0W:-LAIBLm6-8AgNj91nHKb:bkwuo8g) ([raw file](https://github.com/cms-dev/cms/blob/7d09f4c739db21e75806eddec48ef83d899e791c/cmscontrib/loaders/tps.py#L264)):*

No, here I removed it because it wasn't needed anymore after the removal of 2017 (which we might revisit), but if the format produces a Notice, I don't see why not giving it the correct parameters.
</blockquote></details>

Oh ok...

---

*[cmscontrib/loaders/tps.py, line 119 at r12](https://beta.reviewable.io/reviews/cms-dev/cms/935#-LAFRR4d48JTJztTOiXP:-LAIBErH16t1dYTcAnA5:bgqax00) ([raw file](https://github.com/cms-dev/cms/blob/33da5975470f5fcf9c576f75a896ba0a61e885f8/cmscontrib/loaders/tps.py#L119)):*

Well, an empty list is as arbitrary as an empty string, but sure
</blockquote></details>

Right: why not raise something then? :P

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cms-dev/cms/935#-:-LAIBZpx8N_R0YKxyJeF:bnfp4nl)*

stefano-maggiolo:

---

*[cmscontrib/loaders/tps.py, line 119 at r12](https://beta.reviewable.io/reviews/cms-dev/cms/935#-LAFRR4d48JTJztTOiXP:-LAIv5Eo6qm8XSAdthgF:b9fbsgv) ([raw file](https://github.com/cms-dev/cms/blob/33da5975470f5fcf9c576f75a896ba0a61e885f8/cmscontrib/loaders/tps.py#L119)):*

Right: why not raise something then? :P
</blockquote></details>

Oh, I guess this is for Notice :P 

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cms-dev/cms/935)*

lerks:

---

*[cmscontrib/loaders/tps.py, line 119 at r12](https://beta.reviewable.io/reviews/cms-dev/cms/935#-LAFRR4d48JTJztTOiXP:-LAKrvC87VPfsG9uVXoD:bulcdeh) ([raw file](https://github.com/cms-dev/cms/blob/33da5975470f5fcf9c576f75a896ba0a61e885f8/cmscontrib/loaders/tps.py#L119)):*

Oh, I guess this is for Notice :P 
</blockquote></details>

Oooh! Can we make it more explicit? At least with a comment or, better, with another `if` and then raising an exception in the (implicit) else.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cms-dev/cms/935)*

akmohtashami:
The changes in 2017-suffixed task types are mostly minor details (for example Communication2017 uses grader instead of stub to avoid confusion for contestants). So I don't think there would be a problem to use normal task types instead. 

It is possible that I am forgetting something since some time has passed but hopefully I am not :-)

stefano-maggiolo:
PTAL. At least for regular submissions, Batch should be ok, Communication still has a problem (but I tried to make it as working as possible with the existing task type), see the last commit message.

Turns out, if we used dicts for TT parameters, it would have been a better idea than lists.

lerks:
Reviewed 1 of 1 files at r15, 1 of 1 files at r16.

---



*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/935)*

PR #938, opened 2018-04-15
Aggregate compilation stats instead of taking the last

### LINKED ISSUE #927: Compilation time reported only for the last compilation command
When a language definition has multiple compilation commands, compilation time seems to be reported only for the last of them.

Steps to reproduce:
- Modify language compilation commands, for example, in `java_jdk.py`, to contain
  ```
  ["/bin/sh", "-c", "X=$(seq 1 10000000)"], ["/bin/true"]
  ```
  as the last two items. The penultimate command takes about 1 second and 158 MiB to run.
- Submit a solution using the modified language.
- CWS submission result reports compilation time about 0.005 s. Used memory seems to be correct.

Probably it is also good to check that the compilation time limit is applied to all compilation commands together and not to each of them separately.

stefano-maggiolo:
Fixes #927

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/938)

### DIFF
--- a/cms/grading/__init__.py
+++ b/cms/grading/__init__.py
@@ -282,6 +282,7 @@ def compilation_step(sandbox, commands):
     logger.debug("Starting compilation step.")
     stdouts = []
     stderrs = []
+    stats = None
     for step, command in enumerate(commands):
         # Keep stdout and stderr of each compilation step
         sandbox.stdout_file = "compiler_stdout_%d.txt" % step
@@ -292,6 +293,7 @@ def compilation_step(sandbox, commands):
             logger.error("Compilation aborted because of "
                          "sandbox error in `%s'.", sandbox.path)
             return False, None, None, None
+
         stdout = sandbox.get_file_to_string(sandbox.stdout_file)\
             .decode("utf-8", errors="replace").strip()
         if len(stdout) > 0:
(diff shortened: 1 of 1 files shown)


lerks:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---

*[cms/grading/__init__.py, line 308 at r1](https://beta.reviewable.io/reviews/cms-dev/cms/938#-LA8VXrs8Lnumk4gvpkL:-LA8VXrs8Lnumk4gvpkM:b-vjy72p) ([raw file](https://github.com/cms-dev/cms/blob/c0669ac60b7ce857526cbcbf0982db585169c313/cms/grading/__init__.py#L308)):*
> ```Python
>         this_stats = execution_stats(sandbox)
>         stats = this_stats if stats is None else \
>             merge_execution_stats(stats, this_stats, concurrent=False)
> ```

Consider using a "normal" `if` rather than an inline ternary operator.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cms-dev/cms/938#-:-LA8Vgcb0H-CwfP0tDm9:bnfp4nl)*

stefano-maggiolo:

---

*[cms/grading/__init__.py, line 308 at r1](https://beta.reviewable.io/reviews/cms-dev/cms/938#-LA8VXrs8Lnumk4gvpkL:-LA8blsb7lXxhbBfRz8t:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/c0669ac60b7ce857526cbcbf0982db585169c313/cms/grading/__init__.py#L308)):*

Consider using a "normal" `if` rather than an inline ternary operator.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cms-dev/cms/938)*

stefano-maggiolo:
Merged.

PR #941, opened 2018-04-18
Some more work on steps

stefano-maggiolo:
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/941)


stefano-maggiolo:
Actually, not ready yet, sorry.

PR #942, opened 2018-04-18
Make names in tests more deterministic

stefano-maggiolo:
A common annoyance while developing and using functional tests to get
some sample data is that the names keep changing, and a browser cannot
remember the login, among other pains. Also it is impossible to
understand which one was the contest for the latest run.

With this change, the first run will have no suffix, the second will
have suffix "0", and so on.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/942)

### DIFF
--- a/cmstestsuite/functionaltestframework.py
+++ b/cmstestsuite/functionaltestframework.py
@@ -121,18 +121,37 @@ class FunctionalTestFramework(object):
             self._cws_browser.login(lr)
         return self._cws_browser
 
-    def initialize_aws(self, rand):
-        """Create an admin and logs in
+    def initialize_aws(self):
+        """Create an admin.
 
-        rand (int): some random bit to add to the admin username.
+        The username will be admin, unless already used. In that case it will
+        be admin<suffix>, where <suffix> will be the first integer (from 0)
+        for which an admin with that name doesn't yet exists.
+
+        return (str): the suffix.
 
         """
         logger.info("Creating admin...")
-        self.admin_info["username"] = "admin%s" % rand
         self.admin_info["password"] = "adminpwd"
-        sh([sys.executable, "cmscontrib/AddAdmin.py",
-            "%(username)s" % self.admin_info,
-            "-p", "%(password)s" % self.admin_info])
+
+        suffix = ""
+        while True:
+            self.admin_info["username"] = "admin%s" % suffix
+            logger.info("Trying %(username)s" % self.admin_info)
+            try:
+                sh([sys.executable, "cmscontrib/AddAdmin.py",
+                    "%(username)s" % self.admin_info,
+                    "-p", "%(password)s" % self.admin_info],
+                   ignore_failure=False)
+            except TestException:
+                if suffix == "":
+                    suffix = "0"
+                else:
+                    suffix = str(int(suffix) + 1)
+            else:
+                break
+
+        return suffix
 
     def get_cms_config(self):
         if self._cms_config is None:
--- a/cmstestsuite/functionaltestframework.py
+++ b/cmstestsuite/functionaltestframework.py
@@ -126,7 +126,7 @@ class FunctionalTestFramework(object):
 
         The username will be admin, unless already used. In that case it will
         be admin<suffix>, where <suffix> will be the first integer (from 0)
-        for which an admin with that name doesn't yet exists.
+        for which an admin with that name doesn't yet exist.
 
         return (str): the suffix.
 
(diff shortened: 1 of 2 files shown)


lerks:
:lgtm:

Just puzzled by the fact that the order is `""` -> `"0"` -> `"1"` -> ...
I would have chosen `"1"` -> `"2"` -> ...

PS just came to my mind: why not use the time of the test (formatted as a y-m-d-h-m-s or as a timestamp) instead of a random number or a counter?

---

Reviewed 2 of 2 files at r1.

---

*[cmstestsuite/functionaltestframework.py, line 129 at r1](https://beta.reviewable.io/reviews/cms-dev/cms/942#-LAQAl2D4x1EFa8HjX8K:-LAQAl2D4x1EFa8HjX8L:b-1h84ra) ([raw file](https://github.com/cms-dev/cms/blob/14acebc801b0bdc6ab796fa94e3a74a68932670e/cmstestsuite/functionaltestframework.py#L129)):*
> ```Python
>         The username will be admin, unless already used. In that case it will
>         be admin<suffix>, where <suffix> will be the first integer (from 0)
>         for which an admin with that name doesn't yet exists.
> ```

s/exists/exist/

---

*[cmstestsuite/testrunner.py, line 60 at r1](https://beta.reviewable.io/reviews/cms-dev/cms/942#-LAQBOYM-j3yYHzD6jH2:-LAQBOYM-j3yYHzD6jH3:b-vwsvrb) ([raw file](https://github.com/cms-dev/cms/blob/14acebc801b0bdc6ab796fa94e3a74a68932670e/cmstestsuite/testrunner.py#L60)):*
> ```Python
>         self.task_id_map = {}
> 
>         # String to append to object's names to avoid collisions. Will be the
> ```

s/object's/objects'/

---

*[cmstestsuite/testrunner.py, line 138 at r1](https://beta.reviewable.io/reviews/cms-dev/cms/942#-LAQBaJ8CcZDGE5MMLDj:-LAQBaJ8CcZDGE5MMLDk:ba5shht) ([raw file](https://github.com/cms-dev/cms/blob/14acebc801b0bdc6ab796fa94e3a74a68932670e/cmstestsuite/testrunner.py#L138)):*
> ```Python
>         self.contest_id = self.framework.add_contest(
>             name="testcontest" + self.suffix,
>             description="A test contest #%s." % self.suffix,
> ```

This and (below) the username look weird if suffix is empty, don't they?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cms-dev/cms/942#-:-LAQBsPPCRrhtqH5Y_YH:b-6pr1la)*

stefano-maggiolo:
Ok for "" "1" and so on, no preference. For times, I see the point, but then it defuse the advantage of having always the same name when you start clean (or you'd start with "", and you lose the advantage of having the time), and they are immediately long and unintelligible. 

---

Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---

*[cmstestsuite/functionaltestframework.py, line 129 at r1](https://beta.reviewable.io/reviews/cms-dev/cms/942#-LAQAl2D4x1EFa8HjX8K:-LASSs1EAvRorEAK_j8a:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/14acebc801b0bdc6ab796fa94e3a74a68932670e/cmstestsuite/functionaltestframework.py#L129)):*

s/exists/exist/
</blockquote></details>

Done.

---

*[cmstestsuite/testrunner.py, line 60 at r1](https://beta.reviewable.io/reviews/cms-dev/cms/942#-LAQBOYM-j3yYHzD6jH2:-LASSzcq4xkfAxht5Pqa:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/14acebc801b0bdc6ab796fa94e3a74a68932670e/cmstestsuite/testrunner.py#L60)):*

s/object's/objects'/
</blockquote></details>

Done.

---

*[cmstestsuite/testrunner.py, line 138 at r1](https://beta.reviewable.io/reviews/cms-dev/cms/942#-LAQBaJ8CcZDGE5MMLDj:-LASTGxmCGA2mbkni20z:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/14acebc801b0bdc6ab796fa94e3a74a68932670e/cmstestsuite/testrunner.py#L138)):*

This and (below) the username look weird if suffix is empty, don't they?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cms-dev/cms/942#-:-LASRfG30Rjb1RAemKN3:b-njcrfj)*

stefano-maggiolo:
Merged.

PR #949, opened 2018-04-25
Make TwoSteps better (1)

stefano-maggiolo:
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/949)


lerks:
:lgtm:

---

Reviewed 1 of 1 files at r1, 24 of 24 files at r2.

---

*[cmstestsuite/tasks/twosteps/code/manager.c, line 28 at r2](https://reviewable.io/reviews/cms-dev/cms/949#-LAxJPxP5mw61H-rW1BP:-LAxJPxP5mw61H-rW1BQ:b-h4dj5a) ([raw file](https://github.com/cms-dev/cms/blob/141b34ca2fc170fc1283860c376f607f783e5ece/cmstestsuite/tasks/twosteps/code/manager.c#L28)):*
> ```C
>     int output = userfunc2(message);
>     // Write output.
>     printf("correct %d\n", output);
> ```

What's the reason we need the `correct`?

---

*[cmstestsuite/tasks/twosteps_comparator/code/checker, line 16 at r2](https://reviewable.io/reviews/cms-dev/cms/949#-LAxKBoD48fXLYHr_IaN:-LAxKBoD48fXLYHr_IaO:b-hqvi13) ([raw file](https://github.com/cms-dev/cms/blob/141b34ca2fc170fc1283860c376f607f783e5ece/cmstestsuite/tasks/twosteps_comparator/code/checker#L16)):*
> ```
> if output_number == solution_number:
>     print("1.0")
>     print("Correcto", file=sys.stderr)
> ```

Is this a typo or just being funny?

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/949#-:-LAxKp4A8aS4vQQknOsF:bnfp4nl)*

stefano-maggiolo:

---

*[cmstestsuite/tasks/twosteps/code/manager.c, line 28 at r2](https://reviewable.io/reviews/cms-dev/cms/949#-LAxJPxP5mw61H-rW1BP:-LAxOrsE463Ny_bM596Z:b-vaqni8) ([raw file](https://github.com/cms-dev/cms/blob/141b34ca2fc170fc1283860c376f607f783e5ece/cmstestsuite/tasks/twosteps/code/manager.c#L28)):*

What's the reason we need the `correct`?
</blockquote></details>

It's in the output that is compared with the plain diff... it's kind of a strange way of doing a task, but all of these tasks are not good enough to be taken as samples anyway... If you want to make a pass to make them more representative, feel free.

---

*[cmstestsuite/tasks/twosteps_comparator/code/checker, line 16 at r2](https://reviewable.io/reviews/cms-dev/cms/949#-LAxKBoD48fXLYHr_IaN:-LAxP7SX5d7fKqH0mQ1n:bs5skgm) ([raw file](https://github.com/cms-dev/cms/blob/141b34ca2fc170fc1283860c376f607f783e5ece/cmstestsuite/tasks/twosteps_comparator/code/checker#L16)):*

Is this a typo or just being funny?
</blockquote></details>

The latter

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/949)*

stefano-maggiolo:
Looks like the tests in travis are failing due to some permission error when running the checker as a second evaluation_step in the same sandbox. Since I cannot repro locally, and I tried but failed to understand what's going on in travis, and I wanted to make a new sandbox anyway, I'll add the commit to do so here.

lerks:
Reviewed 1 of 25 files at r3, 24 of 24 files at r4, 1 of 1 files at r5, 1 of 1 files at r6, 1 of 1 files at r7, 2 of 2 files at r8, 1 of 1 files at r9.

---

*[cms/grading/tasktypes/TwoSteps.py, line 402 at r8](https://reviewable.io/reviews/cms-dev/cms/949#-LAz7Ejw7-zgiU7jhrTZ:-LAz7Ejw7-zgiU7jhrT_:b-a7re5d) ([raw file](https://github.com/cms-dev/cms/blob/47ea70fedd45ef0a234ddb060d29fafe0199b84f/cms/grading/tasktypes/TwoSteps.py#L402)):*
> ```Python
>         try:
>             if os.path.islink(output_src):
>                 raise FileNotFoundError
> ```

Isn't there a clearer way to do this, like `if os.path.exists(...) and not os.path.islink(...)`?

Also, shouldn't you raise an instance of the class rather than the class itself? (this is something I've never fully understood as both seem to work)

---

*[cmstestsuite/tasks/twosteps/code/manager.c, line 28 at r2](https://reviewable.io/reviews/cms-dev/cms/949#-LAxJPxP5mw61H-rW1BP:-LAz3DujCvpuEhta65aJ:b-r6cgpv) ([raw file](https://github.com/cms-dev/cms/blob/141b34ca2fc170fc1283860c376f607f783e5ece/cmstestsuite/tasks/twosteps/code/manager.c#L28)):*

It's in the output that is compared with the plain diff... it's kind of a strange way of doing a task, but all of these tasks are not good enough to be taken as samples anyway... If you want to make a pass to make them more representative, feel free.
</blockquote></details>

But you could just compare the number, without the `correct` and it would work just the same right? (provided you remove it in the reference outputs as well, of course)

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/949)*

lerks:
Reviewed 2 of 2 files at r10.

---

*[cms/grading/tasktypes/TwoSteps.py, line 364 at r10](https://reviewable.io/reviews/cms-dev/cms/949#-LB0q-Vo0RQj8oQGE2zY:-LB0q-Vo0RQj8oQGE2zZ:b-38zhkp) ([raw file](https://github.com/cms-dev/cms/blob/74c173e8ca8c7fe1bf7366d6fb8cc6e23a9f7335/cms/grading/tasktypes/TwoSteps.py#L364)):*
> ```Python
>                     # Create a brand-new sandbox just for checking. Only admin
>                     # code runs in it, so we allow multithreading.
>                     checkbox = create_sandbox(file_cacher, name="check")
> ```

Wouldn't it make more sense to only create the sandbox if we're actually running something in it, i.e., if we're using a checker?

---

*[cmstestsuite/tasks/twosteps/code/manager.c, line 28 at r2](https://reviewable.io/reviews/cms-dev/cms/949#-LAxJPxP5mw61H-rW1BP:-LB0qDJUBPVlLZhMi8z9:bwiysck) ([raw file](https://github.com/cms-dev/cms/blob/141b34ca2fc170fc1283860c376f607f783e5ece/cmstestsuite/tasks/twosteps/code/manager.c#L28)):*

Yes, but it's already there. I think the idea is that we had "correct" where the output was written by a manager, whereas we have only numbers when it needs to be checked via checker/whitediff
</blockquote></details>

This output here is both being printed by a manager and being checked via whitediff...

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/949)*

stefano-maggiolo:

---

*[cms/grading/tasktypes/TwoSteps.py, line 364 at r10](https://reviewable.io/reviews/cms-dev/cms/949#-LB0q-Vo0RQj8oQGE2zY:-LB0wxm12EZXn-4zyyiB:blfgbs6) ([raw file](https://github.com/cms-dev/cms/blob/74c173e8ca8c7fe1bf7366d6fb8cc6e23a9f7335/cms/grading/tasktypes/TwoSteps.py#L364)):*

Wouldn't it make more sense to only create the sandbox if we're actually running something in it, i.e., if we're using a checker?
</blockquote></details>

Maybe. I guess there is slightly more stress in doing it in the exact place where the user solution ran? I think the overhead of creating the sandbox is quite low anyway, no? 

---

*[cmstestsuite/tasks/twosteps/code/manager.c, line 28 at r2](https://reviewable.io/reviews/cms-dev/cms/949#-LAxJPxP5mw61H-rW1BP:-LB0wXe4EZ4EfBUb76Yr:bgngx8l) ([raw file](https://github.com/cms-dev/cms/blob/141b34ca2fc170fc1283860c376f607f783e5ece/cmstestsuite/tasks/twosteps/code/manager.c#L28)):*

This output here is both being printed by a manager and being checked via whitediff...
</blockquote></details>

I'll accept a PR to make it as you prefer 

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/949)*

stefano-maggiolo:

---

*[cms/grading/tasktypes/TwoSteps.py, line 364 at r10](https://reviewable.io/reviews/cms-dev/cms/949#-LB0q-Vo0RQj8oQGE2zY:-LB0xHFo6ZLdZCiRqSkq:b-oiwuj4) ([raw file](https://github.com/cms-dev/cms/blob/74c173e8ca8c7fe1bf7366d6fb8cc6e23a9f7335/cms/grading/tasktypes/TwoSteps.py#L364)):*

Maybe. I guess there is slightly more stress in doing it in the exact place where the user solution ran? I think the overhead of creating the sandbox is quite low anyway, no? 
</blockquote></details>

In case, this is for a different commit that would change also Batch and OO.

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/949)*

stefano-maggiolo:
Merged.

PR #950, opened 2018-04-26
Deduplicate checker code

stefano-maggiolo:
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/950)

### DIFF
--- a/cms/grading/steps/__init__.py
+++ b/cms/grading/steps/__init__.py
@@ -32,7 +32,7 @@ from .evaluation import EVALUATION_MESSAGES, evaluation_step, \
     human_evaluation_message, is_evaluation_passed
 from .messages import HumanMessage, MessageCollection
 from .stats import execution_stats, merge_execution_stats
-from .trusted import extract_outcome_and_text, trusted_step
+from .trusted import checker_step, extract_outcome_and_text, trusted_step
 from .whitediff import _WHITES, _white_diff, white_diff_step
 
 
@@ -48,7 +48,7 @@ __all__ = [
     # stats_test.py
     "execution_stats", "merge_execution_stats",
     # trusted.py
-    "extract_outcome_and_text", "trusted_step",
+    "checker_step", "extract_outcome_and_text", "trusted_step",
     # whitediff.py
     "_WHITES", "_white_diff", "white_diff_step",
 ]
--- a/cms/grading/steps/trusted.py
+++ b/cms/grading/steps/trusted.py
@@ -40,12 +40,10 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
-
-import errno
-
 from future.builtins.disabled import *  # noqa
 from future.builtins import *  # noqa
 
+import errno
 import logging
 
 from cms import config
--- a/cms/grading/steps/trusted.py
+++ b/cms/grading/steps/trusted.py
@@ -54,6 +54,10 @@ from .utils import generic_step
 logger = logging.getLogger(__name__)
 
 
+# Filename of the manager used to compare output (must be an executable).
+CHECKER_FILENAME = "checker"
+
+
 def _filter_ansi_escape(string):
     """Filter out ANSI commands from the given string.
 
@@ -83,6 +87,8 @@ def extract_outcome_and_text(sandbox):
     return (float, [str]): outcome and text.
 
     raise (ValueError): if cannot decode the data.
+    raise (FileNotFoundError): if any of the sandbox stdout or stderr file
+        is missing.
 
     """
     stdout_file = sandbox.get_file_text(sandbox.stdout_file)
@@ -180,3 +186,47 @@ def trusted_step(sandbox, commands):
         logger.error("Unrecognized sandbox exit status '%s' in trusted step.",
                      exit_status)
         return False, None, None
+
+
+def checker_step(sandbox, managers, input_, correct_output, output_):
+    """Run the explicit checker given by the admins
+
+    sandbox (Sandbox): the sandbox to run the checker in; should already
+        contain input, correct output, and user output; the checker is instead
+        copied from the managers.
+    managers ({str: Manager}): map filenames to the dataset's managers.
+    input_ (str): inner filename of the input (already in the sandbox).
+    correct_output (str): inner filename of the correct output (already in
+        the sandbox).
+    output_ (str): inner filename of the user output (already in the sandbox).
+
+    return (bool, float|None, [str]): success (true if the checker was able
+        to check the solution successfully), outcome and text.
+
+    """
+    # Copy the checker in the sandbox, after making sure it was provided.
+    if CHECKER_FILENAME not in managers:
+        logger.error("Configuration error: missing or invalid checker "
+                     "(it must be named '%s')", CHECKER_FILENAME)
+        return False, None, []
+    sandbox.create_file_from_storage(
+        CHECKER_FILENAME, managers[CHECKER_FILENAME].digest, executable=True)
+
+    command = ["./%s" % CHECKER_FILENAME, input_, correct_output, output_]
+    box_success, success, unused_stats = trusted_step(sandbox, [command])
+    if not box_success or not success:
+        logger.error("Sandbox failed during checker step. "
+                     "See previous logs for the reason.")
+        return False, None, []
+
+    try:
+        outcome, text = extract_outcome_and_text(sandbox)
+    except ValueError as e:
+        logger.error("Invalid output from checker: %s", e)
+        return False, None, []
+    except OSError as e:  # Change to FileNotFoundError when dropping Python 2.
+        # This should not happen, as the redirect is handled by the sandbox.
+        logger.error("Missing stdout or stderr file from checker: %s", e)
+        return False, None, []
+
+    return True, outcome, text
--- a/cms/grading/steps/trusted.py
+++ b/cms/grading/steps/trusted.py
@@ -192,14 +192,16 @@ def trusted_step(sandbox, commands):
         return False, None, None
 
 
-def checker_step(sandbox, managers, input_digest, correct_output_digest,
-                 output_filename, checker_codename=CHECKER_FILENAME):
+def checker_step(sandbox, checker_digest, input_digest, correct_output_digest,
+                 output_filename):
     """Run the explicit checker given by the admins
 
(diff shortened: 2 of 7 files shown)


stefano-maggiolo:
Yes, it's in the following PR.

---

---

*[cms/grading/steps/trusted.py, line 58 at r2](https://reviewable.io/reviews/cms-dev/cms/950#-LB5MWsGAKLsrlFdXi7z:-LB6yzGjEspQjg7Sxbcj:b2x952e) ([raw file](https://github.com/cms-dev/cms/blob/193131f44ea191c3c057ea0a1fc6077a8c038bb5/cms/grading/steps/trusted.py#L58)):*

I think I would prefer if this was passed in by the TaskType when it calls `checker_step`: TaskTypes should be able to provide custom values.

If you decide to go this way, the signature of `checker_step` could be simplified so that it takes the manager's digest and the name it should appear with in the sandbox

Also: the filename in the sandbox and the codename in the job's manager dict don't necessarily have to be same, do they?
</blockquote></details>

For the name: I can make it configurable, but I'd like to keep at least this as default to avoid proliferating of constants with the same value in all our TaskTypes.

For the signature: I was indeed in doubt whether to pass the job, or the manager map, or just the checker's digest. The first I didn't like because steps doesn't feel like something that has to do with Jobs. The third I didn't like because it forced me to handle "checker not found" outside.

As for your third question, no it doesn't need to be the same, but I don't understand what you want me to do about it.

---

*[cms/grading/steps/trusted.py, line 200 at r2](https://reviewable.io/reviews/cms-dev/cms/950#-LB5NC7R05Yab0uMvjDc:-LB72FLeBW3J9fUDW4hE:b-gw9w3v) ([raw file](https://github.com/cms-dev/cms/blob/193131f44ea191c3c057ea0a1fc6077a8c038bb5/cms/grading/steps/trusted.py#L200)):*

Input and correct output could just be provided as digests, so they can be fetched "fresh" (I wanna say "virgin": is that PC?) from the FileCacher.

PS: you may have already addressed this issue in later PRs, but this is the first one I'm reviewing so I don't know yet...
</blockquote></details>

No I didn't. Yeah, I think that's a good idea since they should always be digests (note that they were fresh anyway, just they were fetched from the digest outside).

---

*[cms/grading/steps/trusted.py, line 227 at r2](https://reviewable.io/reviews/cms-dev/cms/950#-LB5Nx_YC4KZGaot9i6Y:-LB7fkKx1uZOq7UmdIQC:bv3oju0) ([raw file](https://github.com/cms-dev/cms/blob/193131f44ea191c3c057ea0a1fc6077a8c038bb5/cms/grading/steps/trusted.py#L227)):*

If you only want to catch the case of a file not found (which is what the comment suggests) you should check the errno and reraise if different than enoent. But maybe it *does* make sense to catch all os errors (in that case, you may want to remove the comment, and maybe use EnvironmentError instead so that you also catch IOErrors).
</blockquote></details>

The problem is, in PY3 FNFE is an OSError, but doesn't have an errno, so it becomes tricky to handle both correctly. I don't particularly want to catch more... 

---

*[cmstestsuite/unit_tests/grading/steps/trusted_test.py, line 37 at r2](https://reviewable.io/reviews/cms-dev/cms/950#-LB5Pjz-6WxMhqSdxLTI:-LB7g1NPD5muJN1B00bh:b-yoh25d) ([raw file](https://github.com/cms-dev/cms/blob/193131f44ea191c3c057ea0a1fc6077a8c038bb5/cmstestsuite/unit_tests/grading/steps/trusted_test.py#L37)):*

Is there a better way to import `trusted`? Mixing symbols and submodules seems "gross", but I don't know what I would have done instead...
</blockquote></details>

¯\_(ツ)_/¯

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/950#-:-LB6yx86EQ-ZXvPnB6Dw:b7i4cee)*

lerks:
Reviewed 6 of 6 files at r4.

---

*[cms/grading/steps/trusted.py, line 58 at r2](https://reviewable.io/reviews/cms-dev/cms/950#-LB5MWsGAKLsrlFdXi7z:-LB8J_gm9B2thmIg43Yq:bmusym2) ([raw file](https://github.com/cms-dev/cms/blob/193131f44ea191c3c057ea0a1fc6077a8c038bb5/cms/grading/steps/trusted.py#L58)):*

For the name: I can make it configurable, but I'd like to keep at least this as default to avoid proliferating of constants with the same value in all our TaskTypes.

For the signature: I was indeed in doubt whether to pass the job, or the manager map, or just the checker's digest. The first I didn't like because steps doesn't feel like something that has to do with Jobs. The third I didn't like because it forced me to handle "checker not found" outside.

As for your third question, no it doesn't need to be the same, but I don't understand what you want me to do about it.
</blockquote></details>

What I was proposing was: this module defines a CHECKER_FILENAME constant, which is the name the checker will be stored as in the sandbox; the function's signature takes just the checker's digest (no dict of the all job's managers, no codename, no filename (the above constant is *always* used)); each TaskType has a CHECKER_CODENAME constant that is used to look up the checker's digest in the job's managers dict.

I understand your objections (mainly the fact that the checker-non-existing check needs to be done outside by each tasktype; I don't buy too much the proliferation of constants) so if you want to keep it this way it's fine.

---

*[cms/grading/steps/trusted.py, line 227 at r2](https://reviewable.io/reviews/cms-dev/cms/950#-LB5Nx_YC4KZGaot9i6Y:-LB8J7oE76BYCdCRcdWZ:bzgbimw) ([raw file](https://github.com/cms-dev/cms/blob/193131f44ea191c3c057ea0a1fc6077a8c038bb5/cms/grading/steps/trusted.py#L227)):*

The problem is, in PY3 FNFE is an OSError, but doesn't have an errno, so it becomes tricky to handle both correctly. I don't particularly want to catch more... 
</blockquote></details>

```
In [1]: try:
   ...:     open("nonexisting", "rb").read()
   ...: except OSError as e:
   ...:     print(type(e))
   ...:     print(getattr(e, "errno"))
   ...:
<class 'FileNotFoundError'>
2
```
Maybe it's just the FNFEs that you raise in the mock sandbox that don't have it set?

---

*[cmstestsuite/unit_tests/grading/steps/trusted_test.py, line 250 at r4](https://reviewable.io/reviews/cms-dev/cms/950#-LB8IGDvBioJ76xnIdkB:-LB8IGDwDGAqsz5aFXaz:b-cpmiec) ([raw file](https://github.com/cms-dev/cms/blob/d1d2466a26d4ac2b0eafb98cd7126b9a24c899f8/cmstestsuite/unit_tests/grading/steps/trusted_test.py#L250)):*
> ```Python
>     def setUp(self):
>         super(TestCheckerStep, self).setUp()
>         # By default, this
> ```

Incomplete?

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/950)*

stefano-maggiolo:

---

*[cms/grading/steps/trusted.py, line 58 at r2](https://reviewable.io/reviews/cms-dev/cms/950#-LB5MWsGAKLsrlFdXi7z:-LB8Xtq12qApCIj3Fw_x:b76mb86) ([raw file](https://github.com/cms-dev/cms/blob/193131f44ea191c3c057ea0a1fc6077a8c038bb5/cms/grading/steps/trusted.py#L58)):*

What I was proposing was: this module defines a CHECKER_FILENAME constant, which is the name the checker will be stored as in the sandbox; the function's signature takes just the checker's digest (no dict of the all job's managers, no codename, no filename (the above constant is *always* used)); each TaskType has a CHECKER_CODENAME constant that is used to look up the checker's digest in the job's managers dict.

I understand your objections (mainly the fact that the checker-non-existing check needs to be done outside by each tasktype; I don't buy too much the proliferation of constants) so if you want to keep it this way it's fine.
</blockquote></details>

I forgot to mention: in the case of passing a digest for the checker, I was thinking of allowing none to mean "checker wasn't provided". Would that be ok to you?

---

*[cms/grading/steps/trusted.py, line 227 at r2](https://reviewable.io/reviews/cms-dev/cms/950#-LB5Nx_YC4KZGaot9i6Y:-LB8YTvlA7496RD0_Kou:bis2v2a) ([raw file](https://github.com/cms-dev/cms/blob/193131f44ea191c3c057ea0a1fc6077a8c038bb5/cms/grading/steps/trusted.py#L227)):*

```
In [1]: try:
   ...:     open("nonexisting", "rb").read()
   ...: except OSError as e:
   ...:     print(type(e))
   ...:     print(getattr(e, "errno"))
   ...:
<class 'FileNotFoundError'>
2
```
Maybe it's just the FNFEs that you raise in the mock sandbox that don't have it set?
</blockquote></details>

Uhm, I'll try

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/950)*

lerks:
:lgtm:

---

Reviewed 5 of 5 files at r5, 2 of 2 files at r6.

---

*[cms/grading/steps/trusted.py, line 45 at r6](https://reviewable.io/reviews/cms-dev/cms/950#-LBA_BXg5USvOtHhpIxK:-LBA_BXg5USvOtHhpIxL:b-bgkifs) ([raw file](https://github.com/cms-dev/cms/blob/38d04b522e4285ea7d66ac4cc41b7c9200f28e9f/cms/grading/steps/trusted.py#L45)):*
> ```Python
> 
> import errno
> ```

Can we try to keep the `__future__` and the `future` together (even though technically they are different) and put the stdlib imports later?

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/950#-:-LBA_MMiEA3JVE-2A4LB:bnfp4nl)*

stefano-maggiolo:
Review status: 6 of 7 files reviewed at latest revision, 1 unresolved discussion.

---

*[cms/grading/steps/trusted.py, line 45 at r6](https://reviewable.io/reviews/cms-dev/cms/950#-LBA_BXg5USvOtHhpIxK:-LBAcNof9OwIukkK7BhI:bgcfcdu) ([raw file](https://github.com/cms-dev/cms/blob/38d04b522e4285ea7d66ac4cc41b7c9200f28e9f/cms/grading/steps/trusted.py#L45)):*

Can we try to keep the `__future__` and the `future` together (even though technically they are different) and put the stdlib imports later?
</blockquote></details>

Yup, sorry, it's pycharm that keeps adding them in the middle...

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/950)*

stefano-maggiolo:
Merged.

PR #951, opened 2018-04-26
Make sure we do not leak file descriptors

stefano-maggiolo:
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/951)

### DIFF
--- a/cms/db/filecacher.py
+++ b/cms/db/filecacher.py
@@ -878,8 +878,8 @@ class FileCacher(object):
         """
         clean = True
         for digest, _ in self.list():
+            d = Digester()
             with self.backend.get_file(digest) as fobj:
-                d = Digester()
                 buf = fobj.read(self.CHUNK_SIZE)
                 while len(buf) > 0:
                     d.update(buf)
--- a/cms/db/filecacher.py
+++ b/cms/db/filecacher.py
@@ -557,16 +557,9 @@ class FileCacher(object):
 
         ftmp_handle, temp_file_path = tempfile.mkstemp(dir=self.temp_dir,
                                                        text=False)
-        ftmp = io.open(ftmp_handle, 'wb')
-
-        fobj = self.backend.get_file(digest)
-
-        # Copy the file to a temporary position
-        try:
+        with io.open(ftmp_handle, 'wb') as ftmp, \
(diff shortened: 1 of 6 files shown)


lerks:
:lgtm:

---

Reviewed 6 of 6 files at r1.

---

*[cms/db/filecacher.py, line 882 at r1](https://reviewable.io/reviews/cms-dev/cms/951#-LB62Jz2-HDXhgitI-GT:-LB62Jz2-HDXhgitI-GU:b5rgdgl) ([raw file](https://github.com/cms-dev/cms/blob/5aa1f1d068291bcff83d206d034e66b3558c09be/cms/db/filecacher.py#L882)):*
> ```Python
>         for digest, _ in self.list():
>             with self.backend.get_file(digest) as fobj:
>                 d = Digester()
> ```

This could be before the `with`.

---

*[cms/server/util.py, line 94 at r1](https://reviewable.io/reviews/cms-dev/cms/951#-LB63O-x5JMAngExiOLT:-LB63O-x5JMAngExiOLU:b-g0ybkn) ([raw file](https://github.com/cms-dev/cms/blob/5aa1f1d068291bcff83d206d034e66b3558c09be/cms/server/util.py#L94)):*
> ```Python
>             try:
>                 with self.service.file_cacher.get_file(digest) \
>                         as self.temp_file:
> ```

I'm sort-of surprised that you can use an attribute (instead of a variable): are you sure this works?

---

*[cms/server/util.py, line 101 at r1](https://reviewable.io/reviews/cms-dev/cms/951#-LB632uLEckNdR-aGbw3:-LB632uLEckNdR-aGbw4:bwrgqcb) ([raw file](https://github.com/cms-dev/cms/blob/5aa1f1d068291bcff83d206d034e66b3558c09be/cms/server/util.py#L101)):*
> ```Python
>                 self.finish()
> 
>         def fetch_from_filesystem(self, filepath, content_type, filename):
> ```

I was sure I had removed this method! Maybe I just removed its last user but forgot about it? (relates to #882). If there's no one using it could you just remove it?

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/951#-:-LB64b1f0KX1Wv0oLxPg:bnfp4nl)*

stefano-maggiolo:
Merged.

PR #953, opened 2018-04-28
Extract eval_output from tasktypes, and run whitediff without sandbox

stefano-maggiolo:
- Move eval_output (the function running either checker or white diff
  on the user output) to TaskType.py; it cannot go in steps because it
  involves jobs, and creating and deleting sandboxes; already
  checker_step was a stretch, here we are also formally limited by
  circular imports.

- Add ability to run white diff on file objects instead than on paths.

- Use it to avoid having a sandbox for white diff, as suggested by
  @lerks: files are read directly from their current place (either a
  sandbox, or file cacher).

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/953)

### DIFF
--- a/cms/grading/Sandbox.py
+++ b/cms/grading/Sandbox.py
@@ -368,9 +368,8 @@ class SandboxBase(with_metaclass(ABCMeta, object)):
         executable (bool): to set permissions.
 
         """
-        file_ = self.create_file(path, executable)
-        self.file_cacher.get_file_to_fobj(digest, file_)
-        file_.close()
+        with self.create_file(path, executable) as dest_fobj:
+            self.file_cacher.get_file_to_fobj(digest, dest_fobj)
 
     def create_file_from_string(self, path, content, executable=False):
         """Write some data to a file in the sandbox.
@@ -380,9 +379,8 @@ class SandboxBase(with_metaclass(ABCMeta, object)):
         executable (bool): to set permissions.
 
         """
-        file_ = self.create_file(path, executable)
-        file_.write(content)
-        file_.close()
+        with self.create_file(path, executable) as dest_fobj:
+            dest_fobj.write(content)
 
     def get_file(self, path, trunc_len=None):
         """Open a file in the sandbox given its relative path.
--- a/cms/grading/TaskType.py
+++ b/cms/grading/TaskType.py
@@ -38,17 +38,25 @@ from future.builtins.disabled import *  # noqa
 from future.builtins import *  # noqa
 from six import with_metaclass
 
+import io
 import logging
+import os
 import re
+import shutil
 from abc import ABCMeta, abstractmethod
 
 from cms import config
 from cms.grading import JobException
 from cms.grading.Sandbox import Sandbox
 from cms.grading.Job import CompilationJob, EvaluationJob
+from cms.grading.steps import EVALUATION_MESSAGES, checker_step, \
+    white_diff_fobj_step
 
 
 logger = logging.getLogger(__name__)
+
+
+EVAL_USER_OUTPUT_FILENAME = "user_output.txt"
 
 
 def create_sandbox(file_cacher, multithreaded=False, name=None):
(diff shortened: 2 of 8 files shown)


lerks:
What exactly is the circular import?

---

Reviewed 6 of 6 files at r1.

---

*[cms/grading/TaskType.py, line 160 at r1](https://reviewable.io/reviews/cms-dev/cms/953#-LBBaQ9Z9ZYx494LhzSE:-LBBaQ9Z9ZYx494LhzSF:b-7fp3u) ([raw file](https://github.com/cms-dev/cms/blob/38669316e786532e6208ceff5a353f442b3f9bec/cms/grading/TaskType.py#L160)):*
> ```Python
>         if user_output_path is not None:
>             shutil.copyfile(
>                 user_output_path, sandbox.relative_path("output.txt"))
> ```

Wanna extract this to a constant? (same just below)

---

*[cms/grading/TaskType.py, line 167 at r1](https://reviewable.io/reviews/cms-dev/cms/953#-LBBbEdY4bz0FoIjb4jI:-LBBbEdY4bz0FoIjb4jJ:b-hblvm4) ([raw file](https://github.com/cms-dev/cms/blob/38669316e786532e6208ceff5a353f442b3f9bec/cms/grading/TaskType.py#L167)):*
> ```Python
>             if checker_codename in job.managers else None
>         success, outcome, text = checker_step(
>             sandbox, checker_digest, job.input, job.output, "output.txt")
> ```

It would be great if `checker_step` accepted a fileobj for the user's output: this would allow to use the same code as below for the white diff. In fact, except for that and for the fact that the sandbox path needs to be added back to the job, we could almost have the sandbox be entirely created inside `checker_step`. I believe both these can be addressed, what do you think?

PS: maybe this function itself could accept directly a fileobj for the user's output?

---

*[cms/grading/steps/whitediff.py, line 129 at r1](https://reviewable.io/reviews/cms-dev/cms/953#-LBBc1F44h5sWMvt71j2:-LBBc1F44h5sWMvt71j3:br0uqwg) ([raw file](https://github.com/cms-dev/cms/blob/38669316e786532e6208ceff5a353f442b3f9bec/cms/grading/steps/whitediff.py#L129)):*
> ```Python
> 
>     output_fobj (File): file for the user output, opened in binary mode.
>     correct_output_fobj (File): file for the correct output, opened in binary
> ```

Is `File` and actual builtin type? I think elsewhere we called it fobj or fileobj. If you want to use the language from Python 3 type annotations it should be something like IOBase[bytes] (or even more specific if you want/can). More info here: https://docs.python.org/3/library/io.html#class-hierarchy

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/953#-:-LBBdUCs3QoOdD_KOndH:b-v37mfh)*

stefano-maggiolo:
IIRC, task type calls eval_output, eval_output calls create_sandbox (in TaskType). Of course, you can move everything in steps, but then it's like having moved nothing :)

---

---

*[cms/grading/TaskType.py, line 160 at r1](https://reviewable.io/reviews/cms-dev/cms/953#-LBBaQ9Z9ZYx494LhzSE:-LBCpVfJ8uEORkSM0xmi:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/38669316e786532e6208ceff5a353f442b3f9bec/cms/grading/TaskType.py#L160)):*

Wanna extract this to a constant? (same just below)
</blockquote></details>

Done.

---

*[cms/grading/TaskType.py, line 167 at r1](https://reviewable.io/reviews/cms-dev/cms/953#-LBBbEdY4bz0FoIjb4jI:-LBCFZHmC5t6y_h1JuVC:b-hf3hnu) ([raw file](https://github.com/cms-dev/cms/blob/38669316e786532e6208ceff5a353f442b3f9bec/cms/grading/TaskType.py#L167)):*

It would be great if `checker_step` accepted a fileobj for the user's output: this would allow to use the same code as below for the white diff. In fact, except for that and for the fact that the sandbox path needs to be added back to the job, we could almost have the sandbox be entirely created inside `checker_step`. I believe both these can be addressed, what do you think?

PS: maybe this function itself could accept directly a fileobj for the user's output?
</blockquote></details>

I don't like the suggestion of moving the sandbox creation in checker_step: there is a thin division between stuff in TaskType and in steps; the (somehow arbitrary) division is that steps don't deal with jobs, and don't create/delete sandboxes; the guiding reason being that these two concepts allow them to be quite general.

I like the idea of passing a fileobj to checker_step, but honestly the code become a more messy. The convenience of a file in the sandbox is that you don't need to remember to close it, whereas a fileobj this problem.

---

*[cms/grading/steps/whitediff.py, line 129 at r1](https://reviewable.io/reviews/cms-dev/cms/953#-LBBc1F44h5sWMvt71j2:-LBCcpIe4nwF-g36kIuC:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/38669316e786532e6208ceff5a353f442b3f9bec/cms/grading/steps/whitediff.py#L129)):*

Is `File` and actual builtin type? I think elsewhere we called it fobj or fileobj. If you want to use the language from Python 3 type annotations it should be something like IOBase[bytes] (or even more specific if you want/can). More info here: https://docs.python.org/3/library/io.html#class-hierarchy
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/953#-:-LBCoJdZ2MedAZjoz2kf:b-bci18f)*

lerks:
:lgtm: 

Well, I don't see no specific reason for `create_sandbox` and `delete_sandbox` to be in TaskType. Wouldn't it make more sense to have them both in Sandbox? I'm not pushing for this, I'm just saying that maybe it's possible to move things to better places and kill the cyclic dependency.

---

Reviewed 1 of 1 files at r2, 3 of 3 files at r3.

---

*[cms/grading/TaskType.py, line 167 at r1](https://reviewable.io/reviews/cms-dev/cms/953#-LBBbEdY4bz0FoIjb4jI:-LBEX-t_6VBe2CQVTK3y:b-98xxcq) ([raw file](https://github.com/cms-dev/cms/blob/38669316e786532e6208ceff5a353f442b3f9bec/cms/grading/TaskType.py#L167)):*

I don't like the suggestion of moving the sandbox creation in checker_step: there is a thin division between stuff in TaskType and in steps; the (somehow arbitrary) division is that steps don't deal with jobs, and don't create/delete sandboxes; the guiding reason being that these two concepts allow them to be quite general.

I like the idea of passing a fileobj to checker_step, but honestly the code become a more messy. The convenience of a file in the sandbox is that you don't need to remember to close it, whereas a fileobj this problem.

</blockquote></details>

I wasn't aware of the division that sandboxes should only be created by tasktype methods or function, not by steps (maybe this comes from the technical reason that create_sandbox was in TaskType.py?). Anyway, you know the code better than me: if there's a convention, let's keep it.

As for the fileobj, not properly closing them is "not good" but it's not immediately bad: when the refcount of the fileobj gets to zero it is closed automatically. Hence leaks are very short-lived. And since you're now moving everything to use `with`, shouldn't that address this issue automatically? In short: I think fileobjs would make the interface more elegant, I think the leaks are a half-issue, but I don't really know how much messier the code that calls these functions would become. I'll leave the decision to you, I'm fine either way.

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/953#-:-LBEWUBr87KlcbQqHTV7:b-li0g7q)*

stefano-maggiolo:
See other comment. Will submit this now.

---

---

*[cms/grading/TaskType.py, line 167 at r1](https://reviewable.io/reviews/cms-dev/cms/953#-LBBbEdY4bz0FoIjb4jI:-LBHvqGLC8SBBoG83dO-:bf2z7us) ([raw file](https://github.com/cms-dev/cms/blob/38669316e786532e6208ceff5a353f442b3f9bec/cms/grading/TaskType.py#L167)):*

I wasn't aware of the division that sandboxes should only be created by tasktype methods or function, not by steps (maybe this comes from the technical reason that create_sandbox was in TaskType.py?). Anyway, you know the code better than me: if there's a convention, let's keep it.

As for the fileobj, not properly closing them is "not good" but it's not immediately bad: when the refcount of the fileobj gets to zero it is closed automatically. Hence leaks are very short-lived. And since you're now moving everything to use `with`, shouldn't that address this issue automatically? In short: I think fileobjs would make the interface more elegant, I think the leaks are a half-issue, but I don't really know how much messier the code that calls these functions would become. I'll leave the decision to you, I'm fine either way.
</blockquote></details>

Let's say I think things are in a better place than when I started and even if they can still be improved (and I fully agree, esp around create and delete sandbox) there are now many other places that would benefit more from our time :)

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/953#-:-LBHw5TQBOE9cWRPdhDP:b-z2fgnj)*

PR #954, opened 2018-04-28
Minor stuff on worker and ES

stefano-maggiolo:
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/954)

### DIFF
--- a/cms/grading/Sandbox.py
+++ b/cms/grading/Sandbox.py
@@ -1397,7 +1397,10 @@ class IsolateSandbox(SandboxBase):
             [self.box_exec]
             + (["--cg"] if self.cgroup else [])
             + ["--box-id=%d" % self.box_id]
-            + ["--cleanup"])
+            + ["--cleanup"],
+            # Use subprocess.DEVNULL when dropping Python 2.
+            stdout=io.open(os.devnull, "r+b"),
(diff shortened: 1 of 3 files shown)


lerks:
:lgtm: 

It is out of scope but maybe you can have a look at it: `except:` and `except Exception:` are different as the first captures all subclasses of `BaseException`, which is a superclass of `Exception`. Generally, the `BaseException`s that are not `Exception`s are stuff that is so bad and so low-level that it shouldn't be caught. In short, we should replace all `except:` with `except Exception:`. (See items 6 to 11 here: https://www.python.org/dev/peps/pep-0008/#programming-recommendations)

---

Reviewed 2 of 2 files at r1, 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/954#-:-LBBetXt5st0kbL6XTHH:bg3f0ho)*

stefano-maggiolo:
Yep, I agree in general except Exception is preferable. If I'm changing something around one of those, I'll change it.

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/954#-:-LBCEqtW-GnS3_Q0PPEF:blx3zms)*

stefano-maggiolo:
Merged.

PR #955, opened 2018-05-01
Some minor changes in grading, mostly around trying to make the steps package more uniform in what it returns.

stefano-maggiolo:
I think the idea of API compatibility for 1.4 has been thrown away a long time ago...

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/955)

### DIFF
--- a/cms/grading/TaskType.py
+++ b/cms/grading/TaskType.py
@@ -135,8 +135,9 @@ def eval_output(file_cacher, job, checker_codename,
     user_output_filename (str): the filename the user was expected to write to,
         or empty if stdout (used to return an error to the user).
 
-    return (bool, float|None, [str]): success (true if the checker was able
-        to check the solution successfully), outcome and text.
+    return (bool, float|None, [str]|None): success (true if the checker was
+        able to check the solution successfully), outcome and text (both None
+        if success is False).
 
     """
     if (user_output_path is None) == (user_output_digest is None):
--- a/cms/grading/steps/__init__.py
+++ b/cms/grading/steps/__init__.py
@@ -29,7 +29,7 @@ from future.builtins import *  # noqa
 from .compilation import COMPILATION_MESSAGES, compilation_step
 from .evaluation import EVALUATION_MESSAGES, evaluation_step, \
     evaluation_step_after_run, evaluation_step_before_run, \
-    human_evaluation_message, is_evaluation_passed
+    human_evaluation_message
 from .messages import HumanMessage, MessageCollection
 from .stats import execution_stats, merge_execution_stats
 from .trusted import checker_step, extract_outcome_and_text, trusted_step
@@ -43,7 +43,7 @@ __all__ = [
     # evaluation.py
     "EVALUATION_MESSAGES", "evaluation_step",
     "evaluation_step_after_run", "evaluation_step_before_run",
-    "human_evaluation_message", "is_evaluation_passed",
+    "human_evaluation_message",
     # messages.py
     "HumanMessage", "MessageCollection",
     # stats_test.py
--- a/cms/grading/steps/evaluation.py
+++ b/cms/grading/steps/evaluation.py
@@ -125,8 +125,11 @@ def evaluation_step(sandbox, commands,
         used.
     multiprocess (bool): whether to allow multiple thread/processes or not.
 
-    return ((bool, dict|None)): a tuple with two items:
+    return ((bool, bool|None, dict|None)): a tuple with three items:
         * success: True if the sandbox did not fail, in any command;
+        * evaluation_success: True if the solution ran correctly and the output
+            can be evaluated, False if it terminated with an error or was
+            terminated due to resource limitation; None if success is False;
         * stats: a dictionary with statistics about the evaluation, or None
             if success is False.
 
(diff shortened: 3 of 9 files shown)


lerks:
Down with API compatibility! :P

lerks:
:lgtm: 

Do you want to add a CHECK constraint on the DB that ensures that the status, when not NULL, isn't an empty list?

---

Reviewed 5 of 5 files at r1, 5 of 5 files at r2, 1 of 1 files at r3, 1 of 1 files at r4.

---

*[cms/grading/tasktypes/Communication.py, line 274 at r2](https://reviewable.io/reviews/cms-dev/cms/955#-LBSTM6H5LYg5egS3sAG:-LBSTM6H5LYg5egS3sAH:b-y2jmru) ([raw file](https://github.com/cms-dev/cms/blob/c85d7cfb5909c51ab1e74f69fb8d01addfab2c22/cms/grading/tasktypes/Communication.py#L274)):*
> ```Python
>         evaluation_success_user = all(r[1] for r in user_results)
>         plus_user = reduce(merge_execution_stats,
>                            [r[2] for r in user_results])
> ```

Honestly, using these "magical" indices is a bit ugly, and error-prone. This is maybe unorthodox, but we could have the method return a namedtuple, so that these fields can be accessed by name. What do you think?

---

*[cms/grading/tasktypes/Communication.py, line 288 at r2](https://reviewable.io/reviews/cms-dev/cms/955#-LBSTg8s5qrWd28g9ONM:-LBSTg8s5qrWd28g9ONN:b-5ok93q) ([raw file](https://github.com/cms-dev/cms/blob/c85d7cfb5909c51ab1e74f69fb8d01addfab2c22/cms/grading/tasktypes/Communication.py#L288)):*
> ```Python
>         # terminate correctly, we report an error.
>         if not success_user or not success_mgr or not evaluation_success_mgr:
>             success, outcome, text = False, None, None
> ```

I think the change on this line should go in the previous commit?

---

*[cms/service/EvaluationService.py, line 457 at r4](https://reviewable.io/reviews/cms-dev/cms/955#-LBSUAVREV5DM-7CDC4O:-LBSUAVREV5DM-7CDC4P:b-7065s9) ([raw file](https://github.com/cms-dev/cms/blob/9c1ab903de2c922ea8008927d9e1bcfc65cb306c/cms/service/EvaluationService.py#L457)):*
> ```Python
>                     logger.error("`%s' failed, see worker logs and (possibly) "
>                                  "sandboxes at '%s'.",
>                                  operation, " ".join(job.sandboxes))
> ```

Or just use `%r`?

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/955#-:-LBSS6xc1Qn98fzpUYVe:b389ij)*

stefano-maggiolo:
[clarified offline this was about *_text]

Not really, I think we want to accept the theoretical possibility of a non-null text, but where the tasktype has nothing to say to the user. The change here was for texts when CMS had some failure, in which case the text (and all other data for that matters) is not stored in the DB. (Btw, the *_text are non-nullable, rightly so)

---

---

*[cms/grading/tasktypes/Communication.py, line 274 at r2](https://reviewable.io/reviews/cms-dev/cms/955#-LBSTM6H5LYg5egS3sAG:-LBSVDVM5onr3-Uv2a8b:b-41e1it) ([raw file](https://github.com/cms-dev/cms/blob/c85d7cfb5909c51ab1e74f69fb8d01addfab2c22/cms/grading/tasktypes/Communication.py#L274)):*

Honestly, using these "magical" indices is a bit ugly, and error-prone. This is maybe unorthodox, but we could have the method return a namedtuple, so that these fields can be accessed by name. What do you think?
</blockquote></details>

If we'd have to fix this, I'd go in the direction of having a custom object returned, the stats + methods we talked about.

---

*[cms/grading/tasktypes/Communication.py, line 288 at r2](https://reviewable.io/reviews/cms-dev/cms/955#-LBSTg8s5qrWd28g9ONM:-LBSWicxC2cLpUGbwjpD:bxqes93) ([raw file](https://github.com/cms-dev/cms/blob/c85d7cfb5909c51ab1e74f69fb8d01addfab2c22/cms/grading/tasktypes/Communication.py#L288)):*

I think the change on this line should go in the previous commit?
</blockquote></details>

Ok, will change before pshing.

---

*[cms/service/EvaluationService.py, line 457 at r4](https://reviewable.io/reviews/cms-dev/cms/955#-LBSUAVREV5DM-7CDC4O:-LBSUxa13ker1cQsnBsN:bl5kbul) ([raw file](https://github.com/cms-dev/cms/blob/9c1ab903de2c922ea8008927d9e1bcfc65cb306c/cms/service/EvaluationService.py#L457)):*

Or just use `%r`?
</blockquote></details>

But it's ugly!

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/955#-:-LBSVocWEQY0lhBTrJoY:b2b5mjo)*

lerks:
Oh, I see, I misunderstood then, it's fine like this.

---

Review status: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/955#-:-LBUEWFH3M7CqwF8rRw_:bquabuw)*

PR #957, opened 2018-05-02
Move CWS workflows to accept submissions/user tests to own module

lerks:
And add tests.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/957)

### DIFF
--- a/cms/server/contest/handlers/tasksubmission.py
+++ b/cms/server/contest/handlers/tasksubmission.py
@@ -36,7 +36,6 @@ from __future__ import print_function
 from __future__ import unicode_literals
 from future.builtins.disabled import *  # noqa
 from future.builtins import *  # noqa
-from six import iterkeys, itervalues, iteritems
 
 import logging
 import re
@@ -46,18 +45,14 @@ import tornado.web
 from sqlalchemy.orm import joinedload
 
 from cms import config
-from cms.db import File, Submission, SubmissionResult
+from cms.db import Submission, SubmissionResult
 from cms.grading.languagemanager import get_language
 from cms.server import multi_contest
 from cms.server.contest.submission import get_submission_count, \
-    check_max_number, check_min_interval, InvalidArchive, \
-    extract_files_from_tornado, InvalidFilesOrLanguage, \
-    match_files_and_languages, fetch_file_digests_from_previous_submission, \
-    store_local_copy, StorageFailed
+    UnacceptableSubmission, accept_submission
 from cms.server.contest.tokening import UnacceptableToken, TokenAlreadyPlayed, \
     accept_token, tokens_available
 from cmscommon.crypto import encrypt_number
-from cmscommon.datetime import make_timestamp
 from cmscommon.mimetypes import get_type_for_file_name
 
 from ..phase_management import actual_phase_required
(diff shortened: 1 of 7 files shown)


stefano-maggiolo:
Reviewed 3 of 3 files at r1, 5 of 5 files at r2, 1 of 1 files at r3.

---

*[cms/server/contest/handlers/tasksubmission.py, line 97 at r2](https://reviewable.io/reviews/cms-dev/cms/957#-LBXYUaB6sRezI1BOqF4:-LBXYUaB6sRezI1BOqF5:b-wn70zd) ([raw file](https://github.com/cms-dev/cms/blob/c7ec1e44a4868b740cecaf2a2096bfa9301227ce/cms/server/contest/handlers/tasksubmission.py#L97)):*
> ```Python
>             self.sql_session.commit()
>         except UnacceptableSubmission as e:
>             logger.warning("Sent error: `%s' - `%s'", e.subject, e.text)
> ```

So we're logging with warning here, but should be an info, if it was due to the users doing something incorrect by mistake (e.g., submit too frequently). Maybe we should have two different exception types, one for these kinds of errors, and one for failures of CMS? [same for user tests]

or maybe just assume that if a real error happens it's logged upstream, and here just use info.

---

*[cms/server/contest/submission/workflow.py, line 62 at r2](https://reviewable.io/reviews/cms-dev/cms/957#-LBXUBXa31rosnwzE_kT:-LBXUBXa31rosnwzE_kU:b-9k51u1) ([raw file](https://github.com/cms-dev/cms/blob/c7ec1e44a4868b740cecaf2a2096bfa9301227ce/cms/server/contest/submission/workflow.py#L62)):*
> ```Python
> class UnacceptableSubmission(Exception):
>     def __init__(self, subject, text):
>         super(UnacceptableSubmission, self).__init__(subject, text)
> ```

This will render as `('<subject>', '<text>')` if printed, is this what you want? [in user tests as well]

---

*[cms/server/contest/submission/workflow.py, line 82 at r2](https://reviewable.io/reviews/cms-dev/cms/957#-LBXUO8VCde-Fbo4_u71:-LBXUO8VCde-Fbo4_u72:bxyynez) ([raw file](https://github.com/cms-dev/cms/blob/c7ec1e44a4868b740cecaf2a2096bfa9301227ce/cms/server/contest/submission/workflow.py#L82)):*
> ```Python
>     return (Submission):
> 
>     """
> ```

Intentionally types only?! [in user tests as well]

---

*[cmstestsuite/unit_tests/server/contest/submission/workflow_test.py, line 76 at r2](https://reviewable.io/reviews/cms-dev/cms/957#-LBXXhdo3S7070dJCz7f:-LBXXhdo3S7070dJCz7g:bq1q38y) ([raw file](https://github.com/cms-dev/cms/blob/c7ec1e44a4868b740cecaf2a2096bfa9301227ce/cmstestsuite/unit_tests/server/contest/submission/workflow_test.py#L76)):*
> ```Python
>         # Set up patches and mocks for a successful run. These are all
>         # controlled by the following values, which can be changed to
>         # make some steps fail.
> ```

Maybe describe at high level the starting point (task with 3 files, one is provided, one comes from previous submission, ...) [for user tests too]

---

*[cmstestsuite/unit_tests/server/contest/submission/workflow_test.py, line 221 at r2](https://reviewable.io/reviews/cms-dev/cms/957#-LBXWvldDz579IEOSgV9:-LBXWvldDz579IEOSgVA:bavj0oa) ([raw file](https://github.com/cms-dev/cms/blob/c7ec1e44a4868b740cecaf2a2096bfa9301227ce/cmstestsuite/unit_tests/server/contest/submission/workflow_test.py#L221)):*
> ```Python
>     def test_failure_due_to_max_number_on_contest(self):
>         max_number = unique_long_id()
>         self.contest.max_submission_number = max_number
> ```

Nit: overkill to have unique_long_id (here and everywhere else you're using it)

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/957)*

lerks:
I also compared again the two functions and ironed out some small discrepancies (long live symmetry!).

I also added a couple of comments in the tests and changed a minor thing or two to make them more general.

---

Review status: 3 of 7 files reviewed at latest revision, 5 unresolved discussions.

---

*[cms/server/contest/handlers/tasksubmission.py, line 97 at r2](https://reviewable.io/reviews/cms-dev/cms/957#-LBXYUaB6sRezI1BOqF4:-LBZ_trR0_KBErzuGEvr:bywom4i) ([raw file](https://github.com/cms-dev/cms/blob/c7ec1e44a4868b740cecaf2a2096bfa9301227ce/cms/server/contest/handlers/tasksubmission.py#L97)):*

So we're logging with warning here, but should be an info, if it was due to the users doing something incorrect by mistake (e.g., submit too frequently). Maybe we should have two different exception types, one for these kinds of errors, and one for failures of CMS? [same for user tests]

or maybe just assume that if a real error happens it's logged upstream, and here just use info.
</blockquote></details>

The two errors that are due to CMS and not to the user are the failure to store the files (either on the filecacher or the local copy on disk). Both those cases are (now) logged as errors in the workflows. So I think we can just use info here and call it done. Do you agree?

---

*[cms/server/contest/submission/workflow.py, line 62 at r2](https://reviewable.io/reviews/cms-dev/cms/957#-LBXUBXa31rosnwzE_kT:-LBZaGQG1rw9iWLhnBcL:b-cz9ygh) ([raw file](https://github.com/cms-dev/cms/blob/c7ec1e44a4868b740cecaf2a2096bfa9301227ce/cms/server/contest/submission/workflow.py#L62)):*

This will render as `('<subject>', '<text>')` if printed, is this what you want? [in user tests as well]
</blockquote></details>

I don't think I ever printed the exception itself (the handlers always extracted the two attributes) but I'm now calling the superclass constructor with a formatted message: was this what you had in mind?

---

*[cms/server/contest/submission/workflow.py, line 82 at r2](https://reviewable.io/reviews/cms-dev/cms/957#-LBXUO8VCde-Fbo4_u71:-LBZj1hL0atVcEsqpPNn:b-egtbf6) ([raw file](https://github.com/cms-dev/cms/blob/c7ec1e44a4868b740cecaf2a2096bfa9301227ce/cms/server/contest/submission/workflow.py#L82)):*

Intentionally types only?! [in user tests as well]
</blockquote></details>

Nope, just forgot about it, sorry!

---

*[cmstestsuite/unit_tests/server/contest/submission/workflow_test.py, line 76 at r2](https://reviewable.io/reviews/cms-dev/cms/957#-LBXXhdo3S7070dJCz7f:-LBZcohc6xonyjVsJMen:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/c7ec1e44a4868b740cecaf2a2096bfa9301227ce/cmstestsuite/unit_tests/server/contest/submission/workflow_test.py#L76)):*

Maybe describe at high level the starting point (task with 3 files, one is provided, one comes from previous submission, ...) [for user tests too]
</blockquote></details>

Done.

---

*[cmstestsuite/unit_tests/server/contest/submission/workflow_test.py, line 221 at r2](https://reviewable.io/reviews/cms-dev/cms/957#-LBXWvldDz579IEOSgV9:-LBZcrh2AJQBlpK4ckFK:b-n0bi5q) ([raw file](https://github.com/cms-dev/cms/blob/c7ec1e44a4868b740cecaf2a2096bfa9301227ce/cmstestsuite/unit_tests/server/contest/submission/workflow_test.py#L221)):*

Nit: overkill to have unique_long_id (here and everywhere else you're using it)
</blockquote></details>

I wanted to make sure that finding it in the exception message is not a false positive

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/957#-:-LBZj4f_EBRvqGHS3W-v:b-svztjp)*

PR #958, opened 2018-05-07
More grading stuff prompted by writing tests for Batch

stefano-maggiolo:
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/958)

### DIFF
--- a/cms/grading/TaskType.py
+++ b/cms/grading/TaskType.py
@@ -140,6 +140,72 @@ def set_configuration_error(job, msg, *args):
         raise ValueError("Unexpected type of job: %s.", job.__class__)
 
 
+def check_executables_number(job, n_executables):
+    """Check that the required number of executables were generated.
+
+    Since it depends only on the task type being correct, a mismatch here
+    should not happen. It might be caused (with a lot of effort) by compiling
+    under one task type and evaluating under another.
+
+    If there is a mismatch, log and store a configuration error in the job. In
+    this case, callers should terminate immediately the current operation.
+
+    job (Job): the job currently running.
+    n_executables (int): the required number of executables.
+
+    return (bool): whether there is the right number of executables in the job.
+
+    """
+    if len(job.executables) != n_executables:
+        msg = "submission contains %d executables, exactly %d are expected; " \
+              "consider invalidating compilations."
+        set_configuration_error(job, msg, len(job.executables), n_executables)
+        return False
+    return True
+
+
+def check_files_number(job, n_files):
+    """Check that the required number of files were provided by the user.
+
+    A mismatch here is likely caused by having had, at submission time, a wrong
+    submission format for the task type.
+
+    If there is a mismatch, log and store a configuration error in the job. In
+    this case, callers should terminate immediately the current operation.
+
+    job (Job): the job currently running.
+    n_files (int): the required number of files.
+
+    return (bool): whether there is the right number of files in the job.
+
+    """
+    if len(job.files) != n_files:
+        msg = "submission contains %d files, exactly %d are required; " \
+              "ensure the submission format is correct."
+        set_configuration_error(job, msg, len(job.files), n_files)
+        return False
+    return True
+
+
+def check_manager_present(job, codename):
+    """Check that the required manager was provided in the dataset.
+
+    If not provided, log and store a configuration error in the job. In this
+    case, callers should terminate immediately the current operation.
+
+    job (Job): the job currently running.
+    codename (str): the codename of the required manager.
+
+    return (bool): whether the required manager is in the job's managers.
+
+    """
+    if codename not in job.managers:
+        msg = "dataset is missing manager '%s'."
+        set_configuration_error(job, msg, codename)
+        return False
+    return True
+
+
 def eval_output(file_cacher, job, checker_codename,
                 user_output_path=None, user_output_digest=None,
                 user_output_filename=""):
@@ -175,9 +241,7 @@ def eval_output(file_cacher, job, checker_codename,
                                user_output_filename]
 
     if checker_codename is not None:
-        if checker_codename not in job.managers:
-            msg = "dataset is missing manager '%s'."
-            set_configuration_error(job, msg, checker_codename)
+        if not check_manager_present(job, checker_codename):
             return False, None, None
 
         # Create a brand-new sandbox just for checking.
--- a/cms/grading/TaskType.py
+++ b/cms/grading/TaskType.py
@@ -119,6 +119,27 @@ def is_manager_for_compilation(filename, language):
                for obj in language.object_extensions))
 
 
+def set_configuration_error(job, msg, *args):
+    """Log a configuration error and set the correct results in the job.
+
+    job (CompilationJob|EvaluationJob): the job currently executing
+    msg (str): the message to log.
+    args ([object]): formatting parameters for msg.
+
+    """
+    logger.error("Configuration error: " + msg, *args,
+                 extra={"operation": job.info})
+
+    job.success = False
+    job.text = None
+    if isinstance(job, CompilationJob):
+        job.compilation_success = None
+    elif isinstance(job, EvaluationJob):
+        job.outcome = None
+    else:
+        raise ValueError("Unexpected type of job: %s.", job.__class__)
+
+
 def eval_output(file_cacher, job, checker_codename,
                 user_output_path=None, user_output_digest=None,
                 user_output_filename=""):
@@ -154,6 +175,11 @@ def eval_output(file_cacher, job, checker_codename,
                                user_output_filename]
 
     if checker_codename is not None:
+        if checker_codename not in job.managers:
+            msg = "dataset is missing manager '%s'."
+            set_configuration_error(job, msg, checker_codename)
+            return False, None, None
+
         # Create a brand-new sandbox just for checking.
         sandbox = create_sandbox(file_cacher, name="check")
         job.sandboxes.append(sandbox.path)
--- a/cms/grading/tasktypes/Batch.py
+++ b/cms/grading/tasktypes/Batch.py
@@ -37,7 +37,8 @@ from cms.grading.languagemanager import LANGUAGES, get_language
 from cms.grading.ParameterTypes import ParameterTypeCollection, \
     ParameterTypeChoice, ParameterTypeString
 from cms.grading.TaskType import TaskType, \
-    create_sandbox, delete_sandbox, is_manager_for_compilation, eval_output
+    create_sandbox, delete_sandbox, eval_output, is_manager_for_compilation, \
+    set_configuration_error
 from cms.db import Executable
 
 
@@ -174,22 +175,17 @@ class Batch(TaskType):
 
     def compile(self, job, file_cacher):
         """See TaskType.compile."""
-        # Detect the submission's language. The checks about the
-        # formal correctedness of the submission are done in CWS,
-        # before accepting it.
         language = get_language(job.language)
         source_ext = language.source_extension
 
-        # TODO: here we are sure that submission.files are the same as
-        # task.submission_format. The following check shouldn't be
-        # here, but in the definition of the task, since this actually
-        # checks that task's task type and submission format agree.
+        # If the submission was accepted and has 0 or more than 1 files, it
+        # means that either the submission format is wrong for a Batch task, or
+        # it was at the time of submission. Either case, it's a configuration
+        # error that an admin must solve.
         if len(job.files) != 1:
-            job.success = True
-            job.compilation_success = False
-            job.text = [N_("Invalid files in submission")]
-            logger.error("Submission contains %d files, expecting 1",
-                         len(job.files), extra={"operation": job.info})
+            msg = "submission contains %d files, Batch requires exactly 1; " \
+                "ensure the submission format contains 1 element."
+            set_configuration_error(job, msg, len(job.files))
             return
 
         # Create the sandbox.
@@ -242,8 +238,10 @@ class Batch(TaskType):
     def evaluate(self, job, file_cacher):
         """See TaskType.evaluate."""
         if len(job.executables) != 1:
-            raise ValueError("Unexpected number of executables (%s)" %
-                             len(job.executables))
+            msg = "submission contains %d executables, Batch expects 1; " \
+                "consider invalidating compilations."
+            set_configuration_error(job, msg, len(job.executables))
+            return
 
         # Create the sandbox
         sandbox = create_sandbox(file_cacher, name="evaluate")
--- a/cms/grading/tasktypes/Batch.py
+++ b/cms/grading/tasktypes/Batch.py
@@ -188,13 +188,29 @@ class Batch(TaskType):
             set_configuration_error(job, msg, len(job.files))
             return
 
-        # Create the sandbox.
-        sandbox = create_sandbox(file_cacher, name="compile")
-        job.sandboxes.append(sandbox.path)
-
         user_file_format = next(iterkeys(job.files))
         user_source_filename = user_file_format.replace(".%l", source_ext)
         executable_filename = user_file_format.replace(".%l", "")
+
+        # Create the list of filenames to be passed to the compiler. If we use
+        # a grader, it needs to be in first position in the command line, and
+        # we check that it exists.
+        source_filenames = [user_source_filename]
+        if self._uses_grader():
+            grader_source_filename = Batch.GRADER_BASENAME + source_ext
+            if grader_source_filename not in job.managers:
+                msg = "dataset is missing manager '%s'."
+                set_configuration_error(job, msg, grader_source_filename)
+                return
+            source_filenames.insert(0, grader_source_filename)
+
+        # Prepare the compilation command.
+        commands = language.get_compilation_commands(
+            source_filenames, executable_filename)
+
+        # Create the sandbox.
+        sandbox = create_sandbox(file_cacher, name="compile")
+        job.sandboxes.append(sandbox.path)
 
         # Copy required files in the sandbox (includes the grader if present).
         sandbox.create_file_from_storage(
@@ -203,17 +219,6 @@ class Batch(TaskType):
             if is_manager_for_compilation(filename, language):
                 sandbox.create_file_from_storage(
                     filename, job.managers[filename].digest)
-
-        # Create the list of filenames to be passed to the compiler. If we use
-        # a grader, it needs to be in first position in the command line.
-        source_filenames = [user_source_filename]
-        if self._uses_grader():
-            grader_source_filename = Batch.GRADER_BASENAME + source_ext
-            source_filenames.insert(0, grader_source_filename)
-
-        # Prepare the compilation command.
-        commands = language.get_compilation_commands(
-            source_filenames, executable_filename)
 
         # Run the compilation
         operation_success, compilation_success, text, plus = \
@@ -243,10 +248,6 @@ class Batch(TaskType):
             set_configuration_error(job, msg, len(job.executables))
             return
 
-        # Create the sandbox
-        sandbox = create_sandbox(file_cacher, name="evaluate")
-        job.sandboxes.append(sandbox.path)
-
         # Prepare the execution
         executable_filename = next(iterkeys(job.executables))
         language = get_language(job.language)
@@ -274,6 +275,10 @@ class Batch(TaskType):
             stdout_redirect = self._actual_output
         else:
             files_allowing_write.append(self._actual_output)
+
+        # Create the sandbox
+        sandbox = create_sandbox(file_cacher, name="evaluate")
+        job.sandboxes.append(sandbox.path)
 
         # Put the required files into the sandbox
         for filename, digest in iteritems(executables_to_get):
--- a/cms/grading/tasktypes/Batch.py
+++ b/cms/grading/tasktypes/Batch.py
@@ -36,9 +36,9 @@ from cms.grading.steps import compilation_step, evaluation_step,\
 from cms.grading.languagemanager import LANGUAGES, get_language
 from cms.grading.ParameterTypes import ParameterTypeCollection, \
     ParameterTypeChoice, ParameterTypeString
-from cms.grading.TaskType import TaskType, \
-    create_sandbox, delete_sandbox, eval_output, is_manager_for_compilation, \
-    set_configuration_error
+from cms.grading.TaskType import TaskType, check_executables_number, \
+    check_files_number, check_manager_present, create_sandbox, \
+    delete_sandbox, eval_output, is_manager_for_compilation
 from cms.db import Executable
 
 
@@ -178,14 +178,7 @@ class Batch(TaskType):
         language = get_language(job.language)
         source_ext = language.source_extension
 
-        # If the submission was accepted and has 0 or more than 1 files, it
-        # means that either the submission format is wrong for a Batch task, or
-        # it was at the time of submission. Either case, it's a configuration
-        # error that an admin must solve.
-        if len(job.files) != 1:
-            msg = "submission contains %d files, Batch requires exactly 1; " \
-                "ensure the submission format contains 1 element."
-            set_configuration_error(job, msg, len(job.files))
+        if not check_files_number(job, 1):
             return
 
         user_file_format = next(iterkeys(job.files))
@@ -198,9 +191,7 @@ class Batch(TaskType):
         source_filenames = [user_source_filename]
         if self._uses_grader():
             grader_source_filename = Batch.GRADER_BASENAME + source_ext
-            if grader_source_filename not in job.managers:
-                msg = "dataset is missing manager '%s'."
-                set_configuration_error(job, msg, grader_source_filename)
+            if not check_manager_present(job, grader_source_filename):
                 return
             source_filenames.insert(0, grader_source_filename)
 
@@ -242,10 +233,7 @@ class Batch(TaskType):
 
     def evaluate(self, job, file_cacher):
         """See TaskType.evaluate."""
-        if len(job.executables) != 1:
-            msg = "submission contains %d executables, Batch expects 1; " \
-                "consider invalidating compilations."
-            set_configuration_error(job, msg, len(job.executables))
+        if not check_executables_number(job, 1):
             return
 
         # Prepare the execution
@@ -256,8 +244,7 @@ class Batch(TaskType):
         commands = language.get_evaluation_commands(
             executable_filename, main=main)
         executables_to_get = {
-            executable_filename:
-                job.executables[executable_filename].digest
+            executable_filename: job.executables[executable_filename].digest
         }
         files_to_get = {
             self._actual_input: job.input
--- a/cms/grading/tasktypes/Communication.py
+++ b/cms/grading/tasktypes/Communication.py
@@ -173,6 +173,13 @@ class Communication(TaskType):
 
     def evaluate(self, job, file_cacher):
         """See TaskType.evaluate."""
+        # Make sure we have the correct number of executables.
+        if len(job.executables) != 1:
+            msg = "submission contains %d executables, Communication " \
+                "expects 1; consider invalidating compilations."
+            set_configuration_error(job, msg, len(job.executables))
+            return
+
         # Make sure the required manager is among the job managers.
         manager_filename = "manager"
         if manager_filename not in job.managers:
@@ -227,7 +234,6 @@ class Communication(TaskType):
         # Second step: load the executables for the user processes
         # (done before launching the manager so that it does not
         # impact its wall clock time).
-        assert len(job.executables) == 1
         executable_filename = next(iterkeys(job.executables))
         executables_to_get = {
             executable_filename:
--- a/cms/grading/tasktypes/Communication.py
+++ b/cms/grading/tasktypes/Communication.py
@@ -42,7 +42,8 @@ from cms.grading.steps import compilation_step, evaluation_step, \
 from cms.grading.languagemanager import LANGUAGES, get_language
 from cms.grading.ParameterTypes import ParameterTypeInt
 from cms.grading.TaskType import TaskType, \
-    create_sandbox, delete_sandbox, is_manager_for_compilation
+    create_sandbox, delete_sandbox, is_manager_for_compilation, \
+    set_configuration_error
 from cms.db import Executable
 
 
@@ -113,15 +114,15 @@ class Communication(TaskType):
         language = get_language(job.language)
         source_ext = language.source_extension
 
-        # Create the sandbox
-        sandbox = create_sandbox(file_cacher, name="compile")
-        job.sandboxes.append(sandbox.path)
-
         # Prepare the source files in the sandbox
         files_to_get = {}
         source_filenames = []
         # Stub.
         stub_filename = "stub%s" % source_ext
+        if stub_filename not in job.managers:
+            msg = "dataset is missing manager '%s'."
+            set_configuration_error(job, msg, stub_filename)
+            return
         source_filenames.append(stub_filename)
         files_to_get[stub_filename] = job.managers[stub_filename].digest
         # User's submission.
@@ -136,15 +137,19 @@ class Communication(TaskType):
                 files_to_get[filename] = \
                     job.managers[filename].digest
 
-        for filename, digest in iteritems(files_to_get):
-            sandbox.create_file_from_storage(filename, digest)
-
         # Prepare the compilation command
         executable_filename = \
             "_".join(pattern.replace(".%l", "")
                      for pattern in iterkeys(job.files))
         commands = language.get_compilation_commands(
             source_filenames, executable_filename)
+
+        # Create the sandbox and put the required files in it.
+        sandbox = create_sandbox(file_cacher, name="compile")
+        job.sandboxes.append(sandbox.path)
+
+        for filename, digest in iteritems(files_to_get):
+            sandbox.create_file_from_storage(filename, digest)
 
         # Run the compilation
         operation_success, compilation_success, text, plus = \
@@ -168,6 +173,13 @@ class Communication(TaskType):
 
     def evaluate(self, job, file_cacher):
         """See TaskType.evaluate."""
+        # Make sure the required manager is among the job managers.
+        manager_filename = "manager"
+        if manager_filename not in job.managers:
+            msg = "dataset is missing manager '%s'."
+            set_configuration_error(job, msg, manager_filename)
+            return
+
         if len(self.parameters) <= 0:
             num_processes = 1
         else:
@@ -194,7 +206,6 @@ class Communication(TaskType):
             os.chmod(fifo_out[i], 0o666)
 
         # First step: prepare the manager.
-        manager_filename = "manager"
         manager_command = ["./%s" % manager_filename]
         for i in indices:
             manager_command.append(fifo_in[i])
--- a/cms/grading/tasktypes/Communication.py
+++ b/cms/grading/tasktypes/Communication.py
@@ -41,9 +41,9 @@ from cms.grading.steps import compilation_step, evaluation_step, \
     extract_outcome_and_text, human_evaluation_message,  merge_execution_stats
 from cms.grading.languagemanager import LANGUAGES, get_language
 from cms.grading.ParameterTypes import ParameterTypeInt
-from cms.grading.TaskType import TaskType, \
-    create_sandbox, delete_sandbox, is_manager_for_compilation, \
-    set_configuration_error
+from cms.grading.TaskType import TaskType, check_executables_number, \
+    check_manager_present, create_sandbox, delete_sandbox, \
+    is_manager_for_compilation
 from cms.db import Executable
 
 
@@ -119,9 +119,7 @@ class Communication(TaskType):
         source_filenames = []
         # Stub.
         stub_filename = "stub%s" % source_ext
-        if stub_filename not in job.managers:
-            msg = "dataset is missing manager '%s'."
-            set_configuration_error(job, msg, stub_filename)
+        if not check_manager_present(job, stub_filename):
             return
         source_filenames.append(stub_filename)
         files_to_get[stub_filename] = job.managers[stub_filename].digest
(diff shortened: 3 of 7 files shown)


lerks:
Reviewed 3 of 3 files at r1, 1 of 1 files at r2, 3 of 3 files at r3.

---

*[a discussion](https://reviewable.io/reviews/cms-dev/cms/958#-LBwH4KEE_1HG2TH0f0Y:-LBwH4KEE_1HG2TH0f0Z:b-dj62dl) (no related file):*
Could you add `staticmethod` to the ones that don't use `self`? (see codacy's comment)

---

*[a discussion](https://reviewable.io/reviews/cms-dev/cms/958#-LBwH6oPE9ewCOEMyO6w:-LBwH6oPE9ewCOEMyO6x:bsuqmst) (no related file):*
I was expecting that we should check both files and managers, for both operations, on all task types (hence 2 x 2 x 4 = 16 checks in total). In these commits I see only 3/4. Could you add the remaining ones?

---

*[a discussion](https://reviewable.io/reviews/cms-dev/cms/958#-LBwHWyUCDcXDs0gNkih:-LBwHWyUCDcXDs0gNkii:bsyc15o) (no related file):*
Setting a job's success to false means that ES will try to rerun it, right? Is this what we want in case of a misconfiguration?

---

*[cms/grading/tasktypes/Batch.py, line 243 at r1](https://reviewable.io/reviews/cms-dev/cms/958#-LBvEbL62UrKwG73ofgV:-LBvEbL62UrKwG73ofgW:b-rd6o01) ([raw file](https://github.com/cms-dev/cms/blob/b0d6af2af0a1218c34bbc05b3d683d4c031b9a38/cms/grading/tasktypes/Batch.py#L243)):*
> ```Python
>         """See TaskType.evaluate."""
>         if len(job.executables) != 1:
>             msg = "submission contains %d executables, expected 1."
> ```

Could you, like above, mention what task type this is and suggest how to solve this (i.e., invalidate compilation)?

---

*[cms/grading/tasktypes/Batch.py, line 183 at r2](https://reviewable.io/reviews/cms-dev/cms/958#-LBvGJg92Duw_5BpIcbu:-LBvGJg92Duw_5BpIcbv:bykxlj7) ([raw file](https://github.com/cms-dev/cms/blob/b8dfb2ba8f1e0e85cd2d4b93a99e2c14a854feb9/cms/grading/tasktypes/Batch.py#L183)):*
> ```Python
>         # means that either the submission format is wrong for a Batch task, or
>         # it was at the time of submission. Either case, it's a configuration
>         # error that an admin must solve.
> ```

Maybe this change should go in the previous commit? (not so important)

---

*[cms/grading/tasktypes/Batch.py, line 245 at r2](https://reviewable.io/reviews/cms-dev/cms/958#-LBvHD_aEQoIq9C42qqc:-LBvHD_aEQoIq9C42qqd:b-jld52o) ([raw file](https://github.com/cms-dev/cms/blob/b8dfb2ba8f1e0e85cd2d4b93a99e2c14a854feb9/cms/grading/tasktypes/Batch.py#L245)):*
> ```Python
>     def evaluate(self, job, file_cacher):
>         """See TaskType.evaluate."""
>         # This really should not happen.
> ```

Could also go in previous commit. (not so important)

---

*[cmstestsuite/unit_tests/grading/tasktypes/BatchTest.py, line 283 at r3](https://reviewable.io/reviews/cms-dev/cms/958#-LBvePig6RQZPw4DNXdI:-LBvePig6RQZPw4DNXdJ:b-moayw7) ([raw file](https://github.com/cms-dev/cms/blob/966f410921e3e55c0e0a7d1de1f4cb0b1aa20366/cmstestsuite/unit_tests/grading/tasktypes/BatchTest.py#L283)):*
> ```Python
>             # evaluation_step not called or its sandbox gave an error.
>             success, outcome, text, stats = False, None, None, None
>         elif not self.evaluation_step.return_value[1]:  # submission success
> ```

I don't understand the inline comment on this line (is it actually needed given the explanation on the next line?)

---

*[cmstestsuite/unit_tests/grading/tasktypes/BatchTest.py, line 382 at r3](https://reviewable.io/reviews/cms-dev/cms/958#-LBvfKxZC2CYnhZXUrRN:-LBvfKxZC2CYnhZXUrRO:b-mpw411) ([raw file](https://github.com/cms-dev/cms/blob/966f410921e3e55c0e0a7d1de1f4cb0b1aa20366/cmstestsuite/unit_tests/grading/tasktypes/BatchTest.py#L382)):*
> ```Python
>     def test_stdio_diff_eval_output_failure_(self):
>         tt, job = self.prepare(["alone", ["", ""], "diff"], {"foo": EXE_FOO})
>         self.evaluation_step.return_value = (True, True, STATS_OK)
> ```

This seems to be useless, right?

---

*[cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py, line 57 at r3](https://reviewable.io/reviews/cms-dev/cms/958#-LBvHr5PDZd1yFA2_zeh:-LBvHr5PDZd1yFA2_zei:b37oczj) ([raw file](https://github.com/cms-dev/cms/blob/966f410921e3e55c0e0a7d1de1f4cb0b1aa20366/cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py#L57)):*
> ```Python
>     language.get_compilation_commands.side_effect = \
>         lambda srcs, exe: fake_compilation_commands(
>             compilation_command, srcs, exe)
> ```

If you think it's more readable, this is `functools.partial(fake_compilation_commands, compilation_command)` (same below).

---

*[cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py, line 71 at r3](https://reviewable.io/reviews/cms-dev/cms/958#-LBvZq3X13EMLiahqNht:-LBvZq3X13EMLiahqNhu:boh242v) ([raw file](https://github.com/cms-dev/cms/blob/966f410921e3e55c0e0a7d1de1f4cb0b1aa20366/cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py#L71)):*
> ```Python
> LANG_1 = make_language("L1", [".l1"], [".hl1"],
>                        COMPILATION_COMMAND_1, EVALUATION_COMMAND_1)
> LANG_2 = make_language("L2", [".l2"], [".hl2"],
> ```

Wanna try having multiple extensions or extensions that doesn't start with a period?

---

*[cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py, line 85 at r3](https://reviewable.io/reviews/cms-dev/cms/958#-LBvJOJUBjZvfAYX7P6J:-LBvJOJUBjZvfAYX7P6K:bqk0qa7) ([raw file](https://github.com/cms-dev/cms/blob/966f410921e3e55c0e0a7d1de1f4cb0b1aa20366/cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py#L85)):*
> ```Python
>     """A test mixin to make it easier to test task types."""
> 
>     def setUp(self, tasktype):
> ```

Uhm, having a setup with a parameter feels very dirty. My issues are that this only works if it's a subclass calling it, and the subclass needs to know it's calling exactly this method (in particular, it needs to know in which order the superclass methods are resolved, which is a bit tricky with mixins and possibly multiple inheritance). I would prefer if you moved this to an `install_task_type` method.

---

*[cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py, line 92 at r3](https://reviewable.io/reviews/cms-dev/cms/958#-LBvJcCQ4_0MrUpQO7Vg:-LBvJcCQ4_0MrUpQO7Vh:b-ssn9kq) ([raw file](https://github.com/cms-dev/cms/blob/966f410921e3e55c0e0a7d1de1f4cb0b1aa20366/cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py#L92)):*
> ```Python
> 
>         # Ensure we don't retain all sandboxes so we can verify delete().
>         patcher = patch("cms.grading.TaskType.config.keep_sandbox", False)
> ```

You could use `patch.object(config, "keep_sandbox", False)`, so that the global config object is changed in-place and the change will be visible to all users of the config.

---

*[cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py, line 99 at r3](https://reviewable.io/reviews/cms-dev/cms/958#-LBvK3wA5rBuonTqw8UD:-LBvK3wA5rBuonTqw8UE:blew9a6) ([raw file](https://github.com/cms-dev/cms/blob/966f410921e3e55c0e0a7d1de1f4cb0b1aa20366/cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py#L99)):*
> ```Python
>         # the test to change the set of languages CMS understands.
>         self.languages = set()
>         patcher = patch("%s.LANGUAGES" % tasktype_pkg, self.languages)
> ```

Similarly, you can use `patch.dict(LANGUAGES, self.languages)` to update the object in-place and be safer.

---

*[cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py, line 103 at r3](https://reviewable.io/reviews/cms-dev/cms/958#-LBvKLDID6FOzMfXhMw7:-LBvKLDID6FOzMfXhMw8:b-dbd31v) ([raw file](https://github.com/cms-dev/cms/blob/966f410921e3e55c0e0a7d1de1f4cb0b1aa20366/cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py#L103)):*
> ```Python
>         patcher.start()
>         patcher = patch("%s.get_language" % tasktype_pkg,
>                         self._mock_get_language)
> ```

I don't know if it might ever be useful, but if you do the following you can assert how the mocked `get_language` is called:
```py
patcher = patch("%s.get_language" % tasktype_pkg)
self.addCleanup(patcher.stop)
self._mocked_get_language = patcher.start()
self._mocked_get_language.side_effect = self._mock_get_language
```

---

*[cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py, line 112 at r3](https://reviewable.io/reviews/cms-dev/cms/958#-LBvSXEpAHZlFpxpGFb8:-LBvSXEpAHZlFpxpGFb9:b-2m1e5t) ([raw file](https://github.com/cms-dev/cms/blob/966f410921e3e55c0e0a7d1de1f4cb0b1aa20366/cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py#L112)):*
> ```Python
>         self.sandboxes = deque()
>         patcher = patch("cms.grading.TaskType.Sandbox",
>                         MagicMock(side_effect=self._mock_sandbox))
> ```

How is this different from just `patch("cms.grading.TaskType.Sandbox", self._mock_sandbox)`?

Also, why not mock `create_sandbox`?

---

*[cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py, line 133 at r3](https://reviewable.io/reviews/cms-dev/cms/958#-LBvWLNJCG4Pgozeodbv:-LBvWLNJCG4Pgozeodbw:b-477wn) ([raw file](https://github.com/cms-dev/cms/blob/966f410921e3e55c0e0a7d1de1f4cb0b1aa20366/cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py#L133)):*
> ```Python
>         super(TaskTypeTestMixin, self).tearDown()
>         # Make sure the test used all declared sandboxes.
>         assert len(self.sandboxes) == 0
> ```

Why not `self.assertEqual`? (it may be a bug in the code, not in the test, or not?)

---

*[cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py, line 139 at r3](https://reviewable.io/reviews/cms-dev/cms/958#-LBvWqZC9npNBOpsgv0O:-LBvWqZC9npNBOpsgv0P:b-4of32c) ([raw file](https://github.com/cms-dev/cms/blob/966f410921e3e55c0e0a7d1de1f4cb0b1aa20366/cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py#L139)):*
> ```Python
>         sandbox = MagicMock()
>         sandbox.relative_path.side_effect = \
>             lambda p: "/path/%d/%s" % (len(self.sandboxes), p)
> ```

```
In [1]: l = list()

In [2]: f = lambda: "%d" % len(l)

In [3]: f()
Out[3]: '0'

In [4]: l.append(42)

In [5]: f()
Out[5]: '1'
```
which is probably not what you want? (you want it to consistently return the length it had at creation time, right?)

---

*[cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py, line 151 at r3](https://reviewable.io/reviews/cms-dev/cms/958#-LBvXlxs32JDRFEYLf64:-LBvXlxs32JDRFEYLf65:bpepap9) ([raw file](https://github.com/cms-dev/cms/blob/966f410921e3e55c0e0a7d1de1f4cb0b1aa20366/cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py#L151)):*
> ```Python
>     def _mock_sandbox(self, *args, **kwargs):
>         try:
>             return self.sandboxes.popleft()
> ```

You could use a deque rather than a list :P

---

*[cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py, line 153 at r3](https://reviewable.io/reviews/cms-dev/cms/958#-LBvXtZ71rWQOdLnzhsg:-LBvXtZ71rWQOdLnzhsh:b3ffpaq) ([raw file](https://github.com/cms-dev/cms/blob/966f410921e3e55c0e0a7d1de1f4cb0b1aa20366/cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py#L153)):*
> ```Python
>             return self.sandboxes.popleft()
>         except IndexError:
>             raise AssertionError("More calls to Sandbox() than expected.")
> ```

Same here: why not `self.fail(...)`?

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/958)*

stefano-maggiolo:
PTAL

---

Review status: all files reviewed at latest revision, 23 unresolved discussions, some commit checks failed.

---

*[a discussion](https://reviewable.io/reviews/cms-dev/cms/958#-LBwH6oPE9ewCOEMyO6w:-LByryFc7B2GfCQA4eTt:b-n3b2ht) (no related file):*

I was expecting that we should check both files and managers, for both operations, on all task types (hence 2 x 2 x 4 = 16 checks in total). In these commits I see only 3/4. Could you add the remaining ones?
</blockquote></details>

Uhm, not sure what you want me to do. Do you want us to check that we have the correct files and managers in evaluate() too? Why? (I don't even think we have the files, in an EvaluationJob, and of course we don't have the executable in a CompilationJob.) Do you want a test when we have wrong numbers of both managers and files (seems too much)?

---

*[a discussion](https://reviewable.io/reviews/cms-dev/cms/958#-LBwHWyUCDcXDs0gNkih:-LC0SeTd3mpMcvrA7jvd:b2duf87) (no related file):*

Setting a job's success to false means that ES will try to rerun it, right? Is this what we want in case of a misconfiguration?
</blockquote></details>

Well, we don't have a way to say "ignore results but don't retry". AFAICT, raising an unchecked exception is the same as returning success = False on *all* jobs in the job group (which doesn't make a lot of sense in general, maybe it does a little for stuff like "cannot create sandbox"). Success = true implies that we can store the result, which of course we shouldn't if we believe it's not the user's fault.

---

*[a discussion](https://reviewable.io/reviews/cms-dev/cms/958#-LBwH4KEE_1HG2TH0f0Y:-LC0TcmVDhNsoRM1a6cp:b-896fix) (no related file):*

Could you add `staticmethod` to the ones that don't use `self`? (see codacy's comment)
</blockquote></details>

Done.

---

*[cms/grading/TaskType.py, line 385 at r3](https://reviewable.io/reviews/cms-dev/cms/958#-LBufgNGwrRRngqrhgZv-r3-385:-LBytyN58jPGUxychpSb:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/966f410921e3e55c0e0a7d1de1f4cb0b1aa20366/cms/grading/TaskType.py#L385)):*

Codacy Issue found: [Method could be a function](https://app.codacy.com/app/s-maggiolo/cms/pullRequest?prid=1620032)
</blockquote></details>

Done.

---

*[cms/grading/tasktypes/Batch.py, line 243 at r1](https://reviewable.io/reviews/cms-dev/cms/958#-LBvEbL62UrKwG73ofgV:-LC0U5ai0mwXcuHPOuNZ:b-ji908q) ([raw file](https://github.com/cms-dev/cms/blob/b0d6af2af0a1218c34bbc05b3d683d4c031b9a38/cms/grading/tasktypes/Batch.py#L243)):*

Could you, like above, mention what task type this is and suggest how to solve this (i.e., invalidate compilation)?
</blockquote></details>

Done for the suggestions, where did I log the task type? It is implicit in the job.info operation, which displays the submission/UT id. Is that enough?

---

*[cms/grading/tasktypes/Batch.py, line 183 at r2](https://reviewable.io/reviews/cms-dev/cms/958#-LBvGJg92Duw_5BpIcbu:-LC0TeD79te406HThs_z:b-qpqasy) ([raw file](https://github.com/cms-dev/cms/blob/b8dfb2ba8f1e0e85cd2d4b93a99e2c14a854feb9/cms/grading/tasktypes/Batch.py#L183)):*

Maybe this change should go in the previous commit? (not so important)
</blockquote></details>

Will check at the end.

---

*[cms/grading/tasktypes/Batch.py, line 245 at r2](https://reviewable.io/reviews/cms-dev/cms/958#-LBvHD_aEQoIq9C42qqc:-LC0TgrCAWr1lwr9a7fj:b-qpqasy) ([raw file](https://github.com/cms-dev/cms/blob/b8dfb2ba8f1e0e85cd2d4b93a99e2c14a854feb9/cms/grading/tasktypes/Batch.py#L245)):*

Could also go in previous commit. (not so important)
</blockquote></details>

Will check at the end.

---

*[cmstestsuite/unit_tests/grading/tasktypes/BatchTest.py, line 51 at r3](https://reviewable.io/reviews/cms-dev/cms/958#-LBufgNGwrRRngqrhgZy-r3-51:-LC0UGjW3IeAsf-gD1u1:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/966f410921e3e55c0e0a7d1de1f4cb0b1aa20366/cmstestsuite/unit_tests/grading/tasktypes/BatchTest.py#L51)):*

Codacy Issue found: [Parameters differ from overridden 'setUp' method](https://app.codacy.com/app/s-maggiolo/cms/pullRequest?prid=1620032)
</blockquote></details>

Done.

---

*[cmstestsuite/unit_tests/grading/tasktypes/BatchTest.py, line 92 at r3](https://reviewable.io/reviews/cms-dev/cms/958#-LBufgNGwrRRngqrhgZy-r3-92:-LByuelU6Z2qD-Iij22P:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/966f410921e3e55c0e0a7d1de1f4cb0b1aa20366/cmstestsuite/unit_tests/grading/tasktypes/BatchTest.py#L92)):*

Codacy Issue found: [Method could be a function](https://app.codacy.com/app/s-maggiolo/cms/pullRequest?prid=1620032)
</blockquote></details>

Done.

---

*[cmstestsuite/unit_tests/grading/tasktypes/BatchTest.py, line 258 at r3](https://reviewable.io/reviews/cms-dev/cms/958#-LBufgNGwrRRngqrhgZy-r3-258:-LByugAzEW1unF3sRUzj:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/966f410921e3e55c0e0a7d1de1f4cb0b1aa20366/cmstestsuite/unit_tests/grading/tasktypes/BatchTest.py#L258)):*

Codacy Issue found: [Method could be a function](https://app.codacy.com/app/s-maggiolo/cms/pullRequest?prid=1620032)
</blockquote></details>

Done.

---

*[cmstestsuite/unit_tests/grading/tasktypes/BatchTest.py, line 283 at r3](https://reviewable.io/reviews/cms-dev/cms/958#-LBvePig6RQZPw4DNXdI:-LByunapBG5EyirEuIR5:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/966f410921e3e55c0e0a7d1de1f4cb0b1aa20366/cmstestsuite/unit_tests/grading/tasktypes/BatchTest.py#L283)):*

I don't understand the inline comment on this line (is it actually needed given the explanation on the next line?)
</blockquote></details>

Done.

---

*[cmstestsuite/unit_tests/grading/tasktypes/BatchTest.py, line 382 at r3](https://reviewable.io/reviews/cms-dev/cms/958#-LBvfKxZC2CYnhZXUrRN:-LByutlp-6aF_Sqy_Tk9:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/966f410921e3e55c0e0a7d1de1f4cb0b1aa20366/cmstestsuite/unit_tests/grading/tasktypes/BatchTest.py#L382)):*

This seems to be useless, right?
</blockquote></details>

Done.

---

*[cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py, line 57 at r3](https://reviewable.io/reviews/cms-dev/cms/958#-LBvHr5PDZd1yFA2_zeh:-LByvHd5-mWkImw1Cb-E:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/966f410921e3e55c0e0a7d1de1f4cb0b1aa20366/cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py#L57)):*

If you think it's more readable, this is `functools.partial(fake_compilation_commands, compilation_command)` (same below).
</blockquote></details>

Done.

---

*[cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py, line 71 at r3](https://reviewable.io/reviews/cms-dev/cms/958#-LBvZq3X13EMLiahqNht:-LByvAMf-sFF-psizLYD:bo1jgk5) ([raw file](https://github.com/cms-dev/cms/blob/966f410921e3e55c0e0a7d1de1f4cb0b1aa20366/cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py#L71)):*

Wanna try having multiple extensions or extensions that doesn't start with a period?
</blockquote></details>

Not really, it doesn't look very important here.

---

*[cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py, line 85 at r3](https://reviewable.io/reviews/cms-dev/cms/958#-LBvJOJUBjZvfAYX7P6J:-LByvdRw0nkYtQwMAdSt:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/966f410921e3e55c0e0a7d1de1f4cb0b1aa20366/cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py#L85)):*

Uhm, having a setup with a parameter feels very dirty. My issues are that this only works if it's a subclass calling it, and the subclass needs to know it's calling exactly this method (in particular, it needs to know in which order the superclass methods are resolved, which is a bit tricky with mixins and possibly multiple inheritance). I would prefer if you moved this to an `install_task_type` method.
</blockquote></details>

Done.

---

*[cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py, line 92 at r3](https://reviewable.io/reviews/cms-dev/cms/958#-LBvJcCQ4_0MrUpQO7Vg:-LC0ULAO7EsmoBwGjVgw:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/966f410921e3e55c0e0a7d1de1f4cb0b1aa20366/cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py#L92)):*

You could use `patch.object(config, "keep_sandbox", False)`, so that the global config object is changed in-place and the change will be visible to all users of the config.
</blockquote></details>

Done.

---

*[cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py, line 99 at r3](https://reviewable.io/reviews/cms-dev/cms/958#-LBvK3wA5rBuonTqw8UD:-LBywirA7MRcWvKYPZBb:brg02ia) ([raw file](https://github.com/cms-dev/cms/blob/966f410921e3e55c0e0a7d1de1f4cb0b1aa20366/cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py#L99)):*

Similarly, you can use `patch.dict(LANGUAGES, self.languages)` to update the object in-place and be safer.
</blockquote></details>

Uhm, this requires to have self.languages already defined and not changing later, though, no?

---

*[cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py, line 103 at r3](https://reviewable.io/reviews/cms-dev/cms/958#-LBvKLDID6FOzMfXhMw7:-LByyGkS5NQLmsfn_wW9:bebmi8l) ([raw file](https://github.com/cms-dev/cms/blob/966f410921e3e55c0e0a7d1de1f4cb0b1aa20366/cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py#L103)):*

I don't know if it might ever be useful, but if you do the following you can assert how the mocked `get_language` is called:
```py
patcher = patch("%s.get_language" % tasktype_pkg)
self.addCleanup(patcher.stop)
self._mocked_get_language = patcher.start()
self._mocked_get_language.side_effect = self._mock_get_language
```
</blockquote></details>

Done in a slightly different way.

---

*[cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py, line 112 at r3](https://reviewable.io/reviews/cms-dev/cms/958#-LBvSXEpAHZlFpxpGFb8:-LByyLHT2_9oEIWK_blX:b-um2d56) ([raw file](https://github.com/cms-dev/cms/blob/966f410921e3e55c0e0a7d1de1f4cb0b1aa20366/cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py#L112)):*

How is this different from just `patch("cms.grading.TaskType.Sandbox", self._mock_sandbox)`?

Also, why not mock `create_sandbox`?
</blockquote></details>

It's to solve the problem you wanted me to solve with get_language: verify it has been called.

---

*[cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py, line 133 at r3](https://reviewable.io/reviews/cms-dev/cms/958#-LBvWLNJCG4Pgozeodbv:-LByyfj34aV5AWEqRGC4:b1ddde8) ([raw file](https://github.com/cms-dev/cms/blob/966f410921e3e55c0e0a7d1de1f4cb0b1aa20366/cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py#L133)):*

Why not `self.assertEqual`? (it may be a bug in the code, not in the test, or not?)
</blockquote></details>

because this is a mixin and we don't have the methods of TestCase :(

---

*[cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py, line 139 at r3](https://reviewable.io/reviews/cms-dev/cms/958#-LBvWqZC9npNBOpsgv0O:-LC0SReg123Qkw66ZGc4:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/966f410921e3e55c0e0a7d1de1f4cb0b1aa20366/cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py#L139)):*

```
In [1]: l = list()

In [2]: f = lambda: "%d" % len(l)

In [3]: f()
Out[3]: '0'

In [4]: l.append(42)

In [5]: f()
Out[5]: '1'
```
which is probably not what you want? (you want it to consistently return the length it had at creation time, right?)
</blockquote></details>

Done.

---

*[cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py, line 151 at r3](https://reviewable.io/reviews/cms-dev/cms/958#-LBvXlxs32JDRFEYLf64:-LByyyEf2xqnS5_Q0q_F:b-ele6cg) ([raw file](https://github.com/cms-dev/cms/blob/966f410921e3e55c0e0a7d1de1f4cb0b1aa20366/cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py#L151)):*

You could use a deque rather than a list :P
</blockquote></details>

Ehr... I am?

---

*[cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py, line 153 at r3](https://reviewable.io/reviews/cms-dev/cms/958#-LBvXtZ71rWQOdLnzhsg:-LByyzyP6s3xR6O0egXL:bk607dg) ([raw file](https://github.com/cms-dev/cms/blob/966f410921e3e55c0e0a7d1de1f4cb0b1aa20366/cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py#L153)):*

Same here: why not `self.fail(...)`?
</blockquote></details>

Not TestCase

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/958#-:-LC0UWSo84vrZ71fw8t8:bb71pro)*

lerks:
:lgtm:

---

Reviewed 1 of 1 files at r4, 1 of 1 files at r5, 1 of 1 files at r6, 2 of 2 files at r7, 1 of 1 files at r8.
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[a discussion](https://reviewable.io/reviews/cms-dev/cms/958#-LBwH6oPE9ewCOEMyO6w:-LC0h9X82FiUBZhhbHrl:b4hue6s) (no related file):*

Uhm, not sure what you want me to do. Do you want us to check that we have the correct files and managers in evaluate() too? Why? (I don't even think we have the files, in an EvaluationJob, and of course we don't have the executable in a CompilationJob.) Do you want a test when we have wrong numbers of both managers and files (seems too much)?
</blockquote></details>

TaskTypes often (always?) retrieve managers' digests by directly accessing `job.managers` with the filenames they want: if a manager they expect isn't there this fails with a KeyError.

In the second commit you fix this in Batch by checking beforehand and returning a configuration error in case. I would want that to occur for other TaskTypes too. Observe that a similar check would be needed in eval_output too when retrieving the checker.

---

*[a discussion](https://reviewable.io/reviews/cms-dev/cms/958#-LBwHWyUCDcXDs0gNkih:-LC0iiq-5YbKJxVfJ2Ri:b-a1dg3p) (no related file):*

Well, we don't have a way to say "ignore results but don't retry". AFAICT, raising an unchecked exception is the same as returning success = False on *all* jobs in the job group (which doesn't make a lot of sense in general, maybe it does a little for stuff like "cannot create sandbox"). Success = true implies that we can store the result, which of course we shouldn't if we believe it's not the user's fault.
</blockquote></details>

Ok, so this is the best we can do so far, but ideally we would want a "job failed, it's our fault, but don't retry until the admin takes action".

---

*[cms/grading/tasktypes/Batch.py, line 243 at r1](https://reviewable.io/reviews/cms-dev/cms/958#-LBvEbL62UrKwG73ofgV:-LC0duLm-KiDzO5XwIPW:b2v5vjp) ([raw file](https://github.com/cms-dev/cms/blob/b0d6af2af0a1218c34bbc05b3d683d4c031b9a38/cms/grading/tasktypes/Batch.py#L243)):*

Done for the suggestions, where did I log the task type? It is implicit in the job.info operation, which displays the submission/UT id. Is that enough?
</blockquote></details>

In the above you said "*Batch* requires exactly 1". I don't know if we need to provide that information or if admins have other ways of figuring it out. I mainly liked the consistency.

---

*[cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py, line 99 at r3](https://reviewable.io/reviews/cms-dev/cms/958#-LBvK3wA5rBuonTqw8UD:-LC0j3K15NOzFUoZVXLC:bz83aib) ([raw file](https://github.com/cms-dev/cms/blob/966f410921e3e55c0e0a7d1de1f4cb0b1aa20366/cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py#L99)):*

Uhm, this requires to have self.languages already defined and not changing later, though, no?
</blockquote></details>

It's possible but I actually don't know. Since it's not that evident let's just leave it like you originally did.

---

*[cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py, line 133 at r3](https://reviewable.io/reviews/cms-dev/cms/958#-LBvWLNJCG4Pgozeodbv:-LC0epjF6J4HzsDocMe2:b-3ukooy) ([raw file](https://github.com/cms-dev/cms/blob/966f410921e3e55c0e0a7d1de1f4cb0b1aa20366/cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py#L133)):*

because this is a mixin and we don't have the methods of TestCase :(
</blockquote></details>

Some static analyzers might get mad at you but I wouldn't mind. This mixin only makes sense when mixed into a TestCase. (And the only reason we're not extending TestCase here is that TestCase's setUp method doesn't properly call super().setUp... maybe that will change in py3 and then we will be able to do it properly?)

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/958#-:-LC0j_b96CwyuqHB5R-b:bnfp4nl)*

stefano-maggiolo:
Added more stuff, please reLGTM.

---

---

*[a discussion](https://reviewable.io/reviews/cms-dev/cms/958#-LBwH6oPE9ewCOEMyO6w:-LC2uSa9AzSJe2jvIx6Z:b-90sx9e) (no related file):*

TaskTypes often (always?) retrieve managers' digests by directly accessing `job.managers` with the filenames they want: if a manager they expect isn't there this fails with a KeyError.

In the second commit you fix this in Batch by checking beforehand and returning a configuration error in case. I would want that to occur for other TaskTypes too. Observe that a similar check would be needed in eval_output too when retrieving the checker.
</blockquote></details>

I see, I misunderstood. Since you wanted it I added several commits :P take a look if you like them. One thing with the standardized checks is that I removed the task type name from the log - I don't think it's too important, but if you do, there are two ways of doing it: 1. add a parameter to all checks (I don't like it) 2. move checks and eval_output inside TaskType. Let me know. 

---

*[cms/grading/tasktypes/Batch.py, line 243 at r1](https://reviewable.io/reviews/cms-dev/cms/958#-LBvEbL62UrKwG73ofgV:-LC3MwgkBKArWiQyAAwm:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/b0d6af2af0a1218c34bbc05b3d683d4c031b9a38/cms/grading/tasktypes/Batch.py#L243)):*

In the above you said "*Batch* requires exactly 1". I don't know if we need to provide that information or if admins have other ways of figuring it out. I mainly liked the consistency.
</blockquote></details>

Done.

---

*[cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py, line 99 at r3](https://reviewable.io/reviews/cms-dev/cms/958#-LBvK3wA5rBuonTqw8UD:-LC3Myx35lJlh3pUl21g:b-ojtbo6) ([raw file](https://github.com/cms-dev/cms/blob/966f410921e3e55c0e0a7d1de1f4cb0b1aa20366/cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py#L99)):*

It's possible but I actually don't know. Since it's not that evident let's just leave it like you originally did.
</blockquote></details>

Ok, ignoring.

---

*[cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py, line 133 at r3](https://reviewable.io/reviews/cms-dev/cms/958#-LBvWLNJCG4Pgozeodbv:-LC3N1TP4sZYkGmxSwF4:bw9tbu3) ([raw file](https://github.com/cms-dev/cms/blob/966f410921e3e55c0e0a7d1de1f4cb0b1aa20366/cmstestsuite/unit_tests/grading/tasktypes/tasktypetestutils.py#L133)):*

Some static analyzers might get mad at you but I wouldn't mind. This mixin only makes sense when mixed into a TestCase. (And the only reason we're not extending TestCase here is that TestCase's setUp method doesn't properly call super().setUp... maybe that will change in py3 and then we will be able to do it properly?)
</blockquote></details>

Maybe.

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/958#-:-LC3N58A0tr7H4yMslBC:b-wkgjxp)*

lerks:
:lgtm:

---

Reviewed 3 of 3 files at r9, 1 of 1 files at r10, 1 of 1 files at r11, 1 of 1 files at r12, 1 of 1 files at r13, 1 of 1 files at r14, 1 of 1 files at r15, 4 of 4 files at r16.

---

*[cms/grading/TaskType.py, line 150 at r16](https://reviewable.io/reviews/cms-dev/cms/958#-LC3o1t9COe1SjJReCn-:-LC3o1t9COe1SjJReCn0:bcj5c66) ([raw file](https://github.com/cms-dev/cms/blob/0ffcd49c0af7c14a101f579ca73bedbb5d051a58/cms/grading/TaskType.py#L150)):*
> ```Python
>     under one task type and evaluating under another.
> 
>     If there is a mismatch, logs and store a configuration error in the job. In
> ```

Either s/logs/log/ or s/store/stores/ (also 2x below)

---

*[cms/grading/TaskType.py, line 156 at r16](https://reviewable.io/reviews/cms-dev/cms/958#-LC3na1MEZewC8hXM6hY:-LC3na1MEZewC8hXM6hZ:bo3r3wf) ([raw file](https://github.com/cms-dev/cms/blob/0ffcd49c0af7c14a101f579ca73bedbb5d051a58/cms/grading/TaskType.py#L156)):*
> ```Python
>     n_executables (int): the required number of executables.
> 
>     return (bool): whether the required manager is in the job's managers.
> ```

Bad copy and paste

---

*[cms/grading/TaskType.py, line 179 at r16](https://reviewable.io/reviews/cms-dev/cms/958#-LC3nlni1ew9UK76vcli:-LC3nlnjB-4UQkwIKIwQ:bb73xfv) ([raw file](https://github.com/cms-dev/cms/blob/0ffcd49c0af7c14a101f579ca73bedbb5d051a58/cms/grading/TaskType.py#L179)):*
> ```Python
>     n_files (int): the required number of files.
> 
>     return (bool): whether the required manager is in the job's managers.
> ```

Same

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/958#-:-LC3oVJS0c6SapkShUbg:bnfp4nl)*

stefano-maggiolo:

---

*[cms/grading/TaskType.py, line 150 at r16](https://reviewable.io/reviews/cms-dev/cms/958#-LC3o1t9COe1SjJReCn-:-LC5RakMEFEORi6kfrNP:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/0ffcd49c0af7c14a101f579ca73bedbb5d051a58/cms/grading/TaskType.py#L150)):*

Either s/logs/log/ or s/store/stores/ (also 2x below)
</blockquote></details>

Done.

---

*[cms/grading/TaskType.py, line 156 at r16](https://reviewable.io/reviews/cms-dev/cms/958#-LC3na1MEZewC8hXM6hY:-LC5RtWcDzma4jTL7jI5:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/0ffcd49c0af7c14a101f579ca73bedbb5d051a58/cms/grading/TaskType.py#L156)):*

Bad copy and paste
</blockquote></details>

Done.

---

*[cms/grading/TaskType.py, line 179 at r16](https://reviewable.io/reviews/cms-dev/cms/958#-LC3nlni1ew9UK76vcli:-LC5Ru-l6EXS77J6eCup:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/0ffcd49c0af7c14a101f579ca73bedbb5d051a58/cms/grading/TaskType.py#L179)):*

Same
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/958)*

lerks:
Reviewed 1 of 1 files at r17, 1 of 1 files at r18.

---



*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/958)*

stefano-maggiolo:
Thanks!

PR #960, opened 2018-05-10
Support subtasks with a max score of zero

### LINKED ISSUE #931: Show success/failure on subtasks with max score of 0
Some contest organizer set up a subtask with a maximum score of 0 that contains the statement's examples in order to provide contestants with some basic feedback. Group ScoreTypes show such subtasks as always successful, even if all of their testcases failed. That is because `score >= max_score`. A simple change could provide support for this use case: we could base our rendering on `max_score` and, instead of the score in "absolute" terms, the fraction of that `max_score` that the user achieved (e.g., a score of 20 out of a maximum of 25 would give a 0.8 ratio).

This change is easy to implement but one should note that an updater should be written to update all the stored score details.

lerks:
Fixes #931.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/960)


stefano-maggiolo:
Reviewed 1 of 1 files at r1, 3 of 3 files at r2.

---

*[cms/grading/ScoreType.py, line 223 at r2](https://reviewable.io/reviews/cms-dev/cms/960#-LCAdgWt5bnoRlD_ojaU:-LCAdgWt5bnoRlD_ojaV:b-xhdz6a) ([raw file](https://github.com/cms-dev/cms/blob/af285c18d2b959135b5b83fef705e69253917546/cms/grading/ScoreType.py#L223)):*
> ```Python
>         </span>
>     {% if "score_fraction" in st and "max_score" in st %}
>         {% set score = st["score_fraction"] * st["max_score"] %}
> ```

Can we move this earlier and use the same logic as before? It makes more obvious the class and the score will always agree. Maybe we could use score|round(2) directly...

---

*[cmscontrib/updaters/update_33.py, line 44 at r2](https://reviewable.io/reviews/cms-dev/cms/960#-LCAbWLHB9YK2bwP6Brq:-LCAb6aC4FI7UDAmEoYD:bv96c50) ([raw file](https://github.com/cms-dev/cms/blob/af285c18d2b959135b5b83fef705e69253917546/cmscontrib/updaters/update_33.py#L44)):*
> ```Python
> 
> PUBLIC_SUBTASK_KEYS = {"idx", "testcases"}
> PRIVATE_SUBTASK_KEYS = PUBLIC_SUBTASK_KEYS | {"score", "max_score"}
> ```

The names are misleading. The first is "private subtasks in the public details", the second is "private subtask in private details, or public subtask". It's too verbose to be precise, but in case I'd have picked the opposite!

How about "FULL_DETAILS_*" and "PARTIAL_DETAILS_*" and a comment to explain? (More comments in the rest of the code, too, please)

---

*[cmscontrib/updaters/update_33.py, line 52 at r2](https://reviewable.io/reviews/cms-dev/cms/960#-LCAcQWC64hwoW71u__W:-LCAcQWC64hwoW71u__X:b-idtxr9) ([raw file](https://github.com/cms-dev/cms/blob/af285c18d2b959135b5b83fef705e69253917546/cmscontrib/updaters/update_33.py#L52)):*
> ```Python
> OUTCOMES = {"Correct", "Not correct", "Partially correct"}
> 
> NUMBER = (int, float)
> ```

At this point define a is_number() function.

---

*[cmscontrib/updaters/update_33.py, line 103 at r2](https://reviewable.io/reviews/cms-dev/cms/960#-LCAd20j9D36UQ2t_TSR:-LCAd20j9D36UQ2t_TSS:b-2ysxtq) ([raw file](https://github.com/cms-dev/cms/blob/af285c18d2b959135b5b83fef705e69253917546/cmscontrib/updaters/update_33.py#L103)):*
> ```Python
> 
> 
> def convert_score_details(details):
> ```

Some pydoc please

---

*[cmscontrib/updaters/update_33.py, line 121 at r2](https://reviewable.io/reviews/cms-dev/cms/960#-LCAd3nM3z3rNAkO0ckv:-LCAd3nM3z3rNAkO0ckw:b50cj91) ([raw file](https://github.com/cms-dev/cms/blob/af285c18d2b959135b5b83fef705e69253917546/cmscontrib/updaters/update_33.py#L121)):*
> ```Python
>                 # Any value strictly between 0 and 1 would work here.
>                 subtask["score_fraction"] = 0.5
>             inaccurate = True
> ```

Is this because we don't use the actual score type (so we don't know if it is groupmin or something else)?

---

*[cmscontrib/updaters/update_33.py, line 124 at r2](https://reviewable.io/reviews/cms-dev/cms/960#-LCAdBfRAUGc66VCp0W6:-LCAdBfRAUGc66VCp0W7:b4dw6no) ([raw file](https://github.com/cms-dev/cms/blob/af285c18d2b959135b5b83fef705e69253917546/cmscontrib/updaters/update_33.py#L124)):*
> ```Python
> 
>         else:
>             subtask["score_fraction"] = subtask["score"] / subtask["max_score"]
> ```

This doesn't make the rounding different, does it?

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/960)*

lerks:
Review status: 2 of 3 files reviewed at latest revision, 7 unresolved discussions.

---

*[cms/grading/ScoreType.py, line 223 at r2](https://reviewable.io/reviews/cms-dev/cms/960#-LCAdgWt5bnoRlD_ojaU:-LCB8YXQAQ5m3PWYcPzB:brsxxih) ([raw file](https://github.com/cms-dev/cms/blob/af285c18d2b959135b5b83fef705e69253917546/cms/grading/ScoreType.py#L223)):*

Can we move this earlier and use the same logic as before? It makes more obvious the class and the score will always agree. Maybe we could use score|round(2) directly...
</blockquote></details>

Discussed in private and concluded that the score's rounding is really just aesthetic, hence the "real" (highly precise) value should be used to determine the class.

---

*[cmscontrib/updaters/update_33.py, line 44 at r2](https://reviewable.io/reviews/cms-dev/cms/960#-LCAbWLHB9YK2bwP6Brq:-LCB6OzV8c4MSS08IOxW:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/af285c18d2b959135b5b83fef705e69253917546/cmscontrib/updaters/update_33.py#L44)):*

The names are misleading. The first is "private subtasks in the public details", the second is "private subtask in private details, or public subtask". It's too verbose to be precise, but in case I'd have picked the opposite!

How about "FULL_DETAILS_*" and "PARTIAL_DETAILS_*" and a comment to explain? (More comments in the rest of the code, too, please)
</blockquote></details>

Done.

---

*[cmscontrib/updaters/update_33.py, line 52 at r2](https://reviewable.io/reviews/cms-dev/cms/960#-LCAcQWC64hwoW71u__W:-LCB6O366c-PCX-SAtvr:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/af285c18d2b959135b5b83fef705e69253917546/cmscontrib/updaters/update_33.py#L52)):*

At this point define a is_number() function.
</blockquote></details>

Done.

---

*[cmscontrib/updaters/update_33.py, line 103 at r2](https://reviewable.io/reviews/cms-dev/cms/960#-LCAd20j9D36UQ2t_TSR:-LCB6NOe3nFAd0gKB_6z:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/af285c18d2b959135b5b83fef705e69253917546/cmscontrib/updaters/update_33.py#L103)):*

Some pydoc please
</blockquote></details>

Done.

---

*[cmscontrib/updaters/update_33.py, line 121 at r2](https://reviewable.io/reviews/cms-dev/cms/960#-LCAd3nM3z3rNAkO0ckv:-LCAkU_69k2To72KOnbn:bynbuo0) ([raw file](https://github.com/cms-dev/cms/blob/af285c18d2b959135b5b83fef705e69253917546/cmscontrib/updaters/update_33.py#L121)):*

Is this because we don't use the actual score type (so we don't know if it is groupmin or something else)?
</blockquote></details>

The "guessing" algorithm here should work correctly for all of GroupMin, GroupMul and GroupThreshold, because they all happen to have the nice property that any outcome being 0 causes the subtask to be incorrect, and that a subtask is correct iff all outcomes are 1. Theoretically some users could have defined new Group subclasses where this doesn't hold, e.g., GroupMax.

Also, we arbitrarily set the fraction to 0.5 when the real result should have been something else.

Also, we don't support outcomes that are negative or greater than one.

---

*[cmscontrib/updaters/update_33.py, line 124 at r2](https://reviewable.io/reviews/cms-dev/cms/960#-LCAdBfRAUGc66VCp0W6:-LCAi_k1-B39olxfaym9:b63eryd) ([raw file](https://github.com/cms-dev/cms/blob/af285c18d2b959135b5b83fef705e69253917546/cmscontrib/updaters/update_33.py#L124)):*

This doesn't make the rounding different, does it?
</blockquote></details>

Rounding only occurs at rendering time, so the score that we store has a very high precision.

In this case, the score that will eventually be displayed will have been obtained from the original score fraction by multiplying by max score (by the pre-patch score type), then dividing by it (here, in the updater), then multiplying again (in the template of the new score type).

That may sound a lot, but let's say we start with a precision of 9 digits, the max score is almost always 100, hence every time we multiply or divide we lose (at most!) 2 digits: that still leaves enough precision to correctly display the two fractional digits that we round the result to.

(Multiplying/dividing by 100 changes mainly the exponent but should have a small impact on the mantissa, hence I don't really expect two full digits to be affected)

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/960)*

stefano-maggiolo:
:lgtm:

---

Reviewed 1 of 1 files at r3.

---

*[cmscontrib/updaters/update_33.py, line 124 at r2](https://reviewable.io/reviews/cms-dev/cms/960#-LCAdBfRAUGc66VCp0W6:-LCD8SZk6a023YQGQ-XI:b-jgvrkt) ([raw file](https://github.com/cms-dev/cms/blob/af285c18d2b959135b5b83fef705e69253917546/cmscontrib/updaters/update_33.py#L124)):*

Rounding only occurs at rendering time, so the score that we store has a very high precision.

In this case, the score that will eventually be displayed will have been obtained from the original score fraction by multiplying by max score (by the pre-patch score type), then dividing by it (here, in the updater), then multiplying again (in the template of the new score type).

That may sound a lot, but let's say we start with a precision of 9 digits, the max score is almost always 100, hence every time we multiply or divide we lose (at most!) 2 digits: that still leaves enough precision to correctly display the two fractional digits that we round the result to.

(Multiplying/dividing by 100 changes mainly the exponent but should have a small impact on the mantissa, hence I don't really expect two full digits to be affected)
</blockquote></details>

Well, if all subtask's scores were powers of 2 you'd be right :) 

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/960#-:-LCD8i6h0XZqp-D0VsL4:bnfp4nl)*

lerks:
Merged.

PR #965, opened 2018-05-16
Better limits for Communication manager

### LINKED ISSUE #626: Set time limit for comparators
Generally, a comparator (in Batch task) must be written carefully so that it can judge arbitrary user's output correctly. However, sometimes there is a bug in a comparator which causes infinite loop for unexpected user's output. It will prevent all workers from processing another job, and break the entire contest.

I want to add a (optional) time limit and a memory limit for comparators (in Batch task) and managers (in Communication task) to avoid the situation (at least to make the system continue to judge submissions for other tasks). There are two ways to implement this feature, I think.
- Add `comparator_time_limit` field in `Dataset` class, or
- Provide a time limit via `task_type_parameters`

I'm wondering which one is better. What do you think?

stefano-maggiolo:
Re-fixes #626

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/965)

### DIFF
--- a/cms/grading/tasktypes/Communication.py
+++ b/cms/grading/tasktypes/Communication.py
@@ -248,11 +248,27 @@ class Communication(TaskType):
         manager_command = ["./%s" % Communication.MANAGER_FILENAME]
         for i in indices:
             manager_command += [fifo_in[i], fifo_out[i]]
+        # We could use trusted_step for the manager, since it's fully
+        # admin-controlled. But trusted_step is only synchronous at the moment.
+        # Thus we use evaluation_step, and we set a time limit generous enough
+        # to prevent user programs from sending the manager in timeout.
+        # This means that:
+        # - the manager wall clock timeout must be greater than the sum of all
+        #     wall clock timeouts of the user programs;
+        # - with the assumption that the work the manager performs is not
+        #     greater than the work performed by the user programs, the manager
+        #     user timeout must be greater than the maximum allowed total time
+        #     of the user programs; in theory, this is the task's time limit,
+        #     but in practice is num_processes times that because the
+        #     constraint on the total time can only be enforced after all user
+        #     programs terminated.
+        manager_time_limit = max(self.num_processes * (job.time_limit + 1.0),
+                                 config.trusted_sandbox_max_time_s)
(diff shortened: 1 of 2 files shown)


lerks:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---

*[cms/grading/tasktypes/Communication.py, line 257 at r1](https://reviewable.io/reviews/cms-dev/cms/965#-LCeqfGMCeWf-XzaQA_D:-LCeqfGMCeWf-XzaQA_E:bchoqal) ([raw file](https://github.com/cms-dev/cms/blob/7655b9b227862902dbe5608ef08012c66e896ea0/cms/grading/tasktypes/Communication.py#L257)):*
> ```Python
>         # limit (so that the user cannot make the manager timeout).
>         manager_time_limit = self.num_processes \
>             * max(job.time_limit + 1.0, config.trusted_sandbox_max_time_s)
> ```

I'm not sure, but `num_processes * trusted_sandbox_max_time` could easily become a lot, couldn't it? I may prefer `max(num_processes * time_limit + 1, trusted_sandbox_max_time)`, what do you think?

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/965#-:-LCeqdLz--d2tTlgDQn4:bnfp4nl)*

stefano-maggiolo:

---

*[cms/grading/tasktypes/Communication.py, line 257 at r1](https://reviewable.io/reviews/cms-dev/cms/965#-LCeqfGMCeWf-XzaQA_D:-LChLUTcA9fBKgtelY03:bbxbqqn) ([raw file](https://github.com/cms-dev/cms/blob/7655b9b227862902dbe5608ef08012c66e896ea0/cms/grading/tasktypes/Communication.py#L257)):*

I'm not sure, but `num_processes * trusted_sandbox_max_time` could easily become a lot, couldn't it? I may prefer `max(num_processes * time_limit + 1, trusted_sandbox_max_time)`, what do you think?
</blockquote></details>

I'd guess num_processes will never be more than 2, but looks fine to use your formula.

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/965)*

PR #966, opened 2018-05-18
I reviewed user tests for Batch and Communication.

stefano-maggiolo:
I also looked briefly at TwoSteps, and it looks already fine.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/966)

### DIFF
--- a/cms/grading/Job.py
+++ b/cms/grading/Job.py
@@ -5,7 +5,7 @@
 # Copyright © 2012 Giovanni Mascellani <mascellani@poisson.phc.unipi.it>
 # Copyright © 2013-2018 Luca Wehrstedt <luca.wehrstedt@gmail.com>
 # Copyright © 2013 Bernard Blackham <bernard@largestprime.net>
-# Copyright © 2013-2017 Stefano Maggiolo <s.maggiolo@gmail.com>
+# Copyright © 2013-2018 Stefano Maggiolo <s.maggiolo@gmail.com>
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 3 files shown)


lerks:
:lgtm:

---

Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 1 of 1 files at r3.

---

*[cms/grading/Job.py, line 368 at r1](https://reviewable.io/reviews/cms-dev/cms/966#-LCoiQnf0QtfG9dPqaZY:-LCoiQng4J22YmA-DXGR:b-lk0m17) ([raw file](https://github.com/cms-dev/cms/blob/2a192e35e8ef3e71ec55ed85367ba9d861bba52b/cms/grading/Job.py#L368)):*
> ```Python
>         # dict() is required to detach the dictionary that gets added
>         # to the Job from the control of SQLAlchemy
>         language = get_language(user_test.language)
> ```

Should we worry that this may raise KeyError?

---

*[cms/grading/Job.py, line 374 at r1](https://reviewable.io/reviews/cms-dev/cms/966#-LCognyO1PXjOeNLkbpn:-LCognyO1PXjOeNLkbpo:bsxjh0t) ([raw file](https://github.com/cms-dev/cms/blob/2a192e35e8ef3e71ec55ed85367ba9d861bba52b/cms/grading/Job.py#L374)):*
> ```Python
>         if auto_managers is not None:
>             for manager_filename in auto_managers:
>                 if ".%l" in manager_filename:
> ```

`if manager_filename.endswith(".%l"):` (same below)

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/966#-:-LCoiwUj1qa3XWzQLaMR:bnfp4nl)*

stefano-maggiolo:

---

*[cms/grading/Job.py, line 368 at r1](https://reviewable.io/reviews/cms-dev/cms/966#-LCoiQnf0QtfG9dPqaZY:-LComrK123eYew2Oc02r:blzbhr5) ([raw file](https://github.com/cms-dev/cms/blob/2a192e35e8ef3e71ec55ed85367ba9d861bba52b/cms/grading/Job.py#L368)):*

Should we worry that this may raise KeyError?
</blockquote></details>

Done. Good catch.

I decided not to handle the case "manager has %l and language is None" (that is, in that case I don't replace %l) since that will probably cause an error in the manager lookup anyway (and this function is anyway already quite poor in error management...)

---

*[cms/grading/Job.py, line 374 at r1](https://reviewable.io/reviews/cms-dev/cms/966#-LCognyO1PXjOeNLkbpn:-LComtgWEQuG7PRev2Cu:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/2a192e35e8ef3e71ec55ed85367ba9d861bba52b/cms/grading/Job.py#L374)):*

`if manager_filename.endswith(".%l"):` (same below)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/966)*

PR #972, opened 2018-05-29
Profiling tests

stefano-maggiolo:
Also adjusted a bit how we handled coverage since that is now a
parallel operation to profiling.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,7 +60,7 @@ before_script:
   - cmsInitDB
 
 script:
-  - cmsRunTests -v --codecov
+  - cmsRunTests -v --coverage --codecov
 
 notifications:
   email: no
--- a/cmstestsuite/RunFunctionalTests.py
+++ b/cmstestsuite/RunFunctionalTests.py
@@ -3,7 +3,7 @@
 
 # Contest Management System - http://cms-dev.github.io/
 # Copyright © 2012 Bernard Blackham <bernard@largestprime.net>
-# Copyright © 2013-2017 Stefano Maggiolo <s.maggiolo@gmail.com>
+# Copyright © 2013-2018 Stefano Maggiolo <s.maggiolo@gmail.com>
 # Copyright © 2014 Luca Versari <veluca93@gmail.com>
 # Copyright © 2016 Luca Wehrstedt <luca.wehrstedt@gmail.com>
 #
@@ -37,8 +37,8 @@ import sys
 from cms import utf8_decoder
 from cmstestsuite import CONFIG,\
     clear_coverage, combine_coverage, send_coverage_to_codecov
+from cmstestsuite.testrunner import TestRunner
 from cmstestsuite.Tests import ALL_TESTS
-from testrunner import TestRunner
 
 
 logger = logging.getLogger(__name__)
--- a/cmstestsuite/RunFunctionalTests.py
+++ b/cmstestsuite/RunFunctionalTests.py
@@ -35,8 +35,9 @@ import re
 import sys
 
 from cms import utf8_decoder
-from cmstestsuite import CONFIG,\
-    clear_coverage, combine_coverage, send_coverage_to_codecov
+from cmstestsuite import CONFIG
+from cmstestsuite.coverage import clear_coverage, combine_coverage, \
+    send_coverage_to_codecov
 from cmstestsuite.testrunner import TestRunner
 from cmstestsuite.Tests import ALL_TESTS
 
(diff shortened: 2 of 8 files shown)


lerks:
The reason codecov didn't comment is that maybe the third commit broke something. Can you look into it?

---

---

*[cmstestsuite/programstarter.py, line 253 at r2](https://reviewable.io/reviews/cms-dev/cms/972#-LDlnyxfCsU6dVKx_CrO:-LDlnyxfCsU6dVKx_CrP:bidnpw3) ([raw file](https://github.com/cms-dev/cms/blob/f7d00ee6433cd4d81c06b5c2481d7cd8c22e12da/cmstestsuite/programstarter.py#L253)):*
> ```Python
>             stderr = stdout
>         job = subprocess.Popen(cmdline, stdout=stdout, stderr=stderr)
>         atexit.register(lambda: log_cpu_time(job))
> ```

Doesn't this mean that now, if there are any uncaught exceptions that cause `runner.shutdown` not to be called, the spawned services will survive as orphans? That's undesirable as it may interfere with respawning those services and I think this handler was installed exactly for that. What's the problem with double-kill anyway?

---

*[cmstestsuite/RunFunctionalTests.py, line 183 at r3](https://reviewable.io/reviews/cms-dev/cms/972#-LDlpQlT-y-9EcFrxQ9m:-LDlpQlT-y-9EcFrxQ9n:br6jrkl) ([raw file](https://github.com/cms-dev/cms/blob/4dfdff26d04fd864271bca63864de678cd3853b2/cmstestsuite/RunFunctionalTests.py#L183)):*
> ```Python
>     parser.add_argument(
>         "--codecov", action="store_true",
>         help="send coverage results to Codecov (requires --coverage")
> ```

Missing `)` in help msg

---

*[cmstestsuite/RunFunctionalTests.py, line 189 at r4](https://reviewable.io/reviews/cms-dev/cms/972#-LDlrND94yEgvGtA2XHL:-LDlrND94yEgvGtA2XHM:b-v4dtz0) ([raw file](https://github.com/cms-dev/cms/blob/33d3116ae712ebdf8e50b83c09cff18fbdb64246/cmstestsuite/RunFunctionalTests.py#L189)):*
> ```Python
>     g.add_argument(
>         "--profiler", choices=[
>             PROFILER_NONE, PROFILER_YAPPI, PROFILER_KERNPROF],
> ```

Nit: does it make sense to allow to explicitly choose the "none" one?  (also below)

PS: what would the syntax be? `--profiler ''`, `--profiler=`, ...?

---

*[cmstestsuite/RunTimeTest.py, line 102 at r4](https://reviewable.io/reviews/cms-dev/cms/972#-LDlrkj00sIXgRJvjTPj:-LDlrkj00sIXgRJvjTPk:b-48240n) ([raw file](https://github.com/cms-dev/cms/blob/33d3116ae712ebdf8e50b83c09cff18fbdb64246/cmstestsuite/RunTimeTest.py#L102)):*
> ```Python
>     parser.add_argument(
>         "--codecov", action="store_true",
>         help="send coverage results to Codecov (requires --coverage")
> ```

Why do you need a no-op codecov argument on a command that doesn't measure coverage?

Also, `)` missing in help msg

---

*[cmstestsuite/RunUnitTests.py, line 155 at r3](https://reviewable.io/reviews/cms-dev/cms/972#-LDlpe9GEp7jHs0y-c2H:-LDlpe9GEp7jHs0y-c2I:bb73xfv) ([raw file](https://github.com/cms-dev/cms/blob/4dfdff26d04fd864271bca63864de678cd3853b2/cmstestsuite/RunUnitTests.py#L155)):*
> ```Python
>     parser.add_argument(
>         "--codecov", action="store_true",
>         help="send coverage results to Codecov (requires --coverage")
> ```

Same

---

*[cmstestsuite/RunUnitTests.py, line 169 at r4](https://reviewable.io/reviews/cms-dev/cms/972#-LDlsYRg2FEDYM2fUhmH:-LDlsYRg2FEDYM2fUhmI:b-3u1cl2) ([raw file](https://github.com/cms-dev/cms/blob/33d3116ae712ebdf8e50b83c09cff18fbdb64246/cmstestsuite/RunUnitTests.py#L169)):*
> ```Python
>     args = parser.parse_args()
>     if args.codecov and not args.coverage:
>         parser.error("--codecov requires --coverage")
> ```

Probably bad copy-paste: you are already parsing the args and doing this check below...

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/972#-:-LDlofxG1AdSQ1KIjk8M:b-6jn9at)*

stefano-maggiolo:
Thanks for the review. Let's merge #976, then I'll reopen on top of that with the comments addressed, sorry about that.

PR #974, opened 2018-05-31
Use BigInteger for storing execution_memory in the database

fagu:
When a submission or user test used more than 2GB of memory, the number
of used bytes doesn't fit in an Integer (which has to be < 2^31 in Postgresql). EvaluationService will throw an exception
``sqlalchemy.exc.DataError: (psycopg2.DataError) integer out of range``.

This commit changes the type of the execution_memory columns to BigInteger (< 2^63 in Postgresql).

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/974)

### DIFF
--- a/cms/db/__init__.py
+++ b/cms/db/__init__.py
@@ -88,7 +88,7 @@ __all__ = [
 
 # Instantiate or import these objects.
 
-version = 33
+version = 34
 
 engine = create_engine(config.database, echo=config.database_debug,
                        pool_timeout=60, pool_recycle=120)
--- a/cms/db/base.py
+++ b/cms/db/base.py
@@ -37,7 +37,8 @@ from sqlalchemy.orm.session import object_session
 from sqlalchemy.orm import \
     class_mapper, object_mapper, ColumnProperty, RelationshipProperty
 from sqlalchemy.types import \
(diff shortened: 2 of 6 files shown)


stefano-maggiolo:
Hi @fagu, nice to see you again! Thanks for the patch - it's something we had to do for a long time.

Just a couple of amends, please:

- for uniformity, there are a few other places to change: compilation_memory in submission.py and usertest.py, and memory_limit in dataset (task.py);
- even if the old schema works fine with BigInteger in the code, we should bump the db version in db/__init__.py and add a no-op updater in cmscontrib/updater_34.py (similar to _14).

fagu:
Thanks for the comments, Stefano!

I've changed everything except the memory_limit. I think the memory_limit is in MB, so the maximum is currently roughly 2^21 GB. It therefore doesn't seem like a pressing issue. :) If you like, I could of course make it a BigInteger anyways.

lerks:
Yes, please, make the memory limit a BigInteger too: I've been wanting to change it to bytes for a while and one day I'll do it, if it's already a BigInteger I won't risk forgetting to change it.

fagu:
Done!

stefano-maggiolo:
Merged, thank you!

PR #981, opened 2018-06-27
Option to limit feedback to avoid reverse engineering test data

stefano-maggiolo:
Before, we offered many bits of information for each testcase back to
the contestant: high precision time and memory, killing signal.
Combined with showing information for all testcases, this gave the
possibility to discover a lot of information about the test data.

This patch introduce an option for the task to limit this feedback. By
default, the option is active (limited feedback) but to preserve the
old behavior for existing tasks the updater sets it to inactive.

The idea comes from the IOI, and the initial code from Amir Keivan
Mohtashami (see in particular [1] and [2]).

[1] https://github.com/akmohtashami/cms/commit/cecd9863
[2] https://github.com/akmohtashami/cms/commit/06b217a7

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/981)

### DIFF
--- a/cms/__init__.py
+++ b/cms/__init__.py
@@ -40,6 +40,7 @@ __all__ = [
     "__version__",
     "SCORE_MODE_MAX", "SCORE_MODE_MAX_TOKENED_LAST", "TOKEN_MODE_DISABLED",
     "TOKEN_MODE_FINITE", "TOKEN_MODE_INFINITE", "TOKEN_MODE_MIXED",
+    "FEEDBACK_LEVEL_FULL", "FEEDBACK_LEVEL_RESTRICTED",
     # log
     # Nothing intended for external use, no need to advertise anything.
     # conf
@@ -73,6 +74,15 @@ TOKEN_MODE_INFINITE = "infinite"
 # Only used when aggregating modes (e.g., for all tasks of a contest).
 # Acts as a "none of the above".
 TOKEN_MODE_MIXED = "mixed"
+
+# Feedback level.
+
+# Full information (killing signals, time and memory, status for all
+# testcases).
+FEEDBACK_LEVEL_FULL = "full"
+# Restricted set of information (no killing signal, time or memory, testcases
+# can be omitted).
+FEEDBACK_LEVEL_RESTRICTED = "restricted"
 
 
 from .conf import Address, ServiceCoord, ConfigError, async_config, config
--- a/cms/db/__init__.py
+++ b/cms/db/__init__.py
@@ -89,7 +89,7 @@ __all__ = [
 
 # Instantiate or import these objects.
 
-version = 35
+version = 36
 
 engine = create_engine(config.database, echo=config.database_debug,
                        pool_timeout=60, pool_recycle=120)
--- a/cms/db/task.py
+++ b/cms/db/task.py
@@ -45,7 +45,8 @@ from sqlalchemy.orm.collections import attribute_mapped_collection
 from sqlalchemy.dialects.postgresql import ARRAY, JSONB
 
 from cms import SCORE_MODE_MAX, SCORE_MODE_MAX_TOKENED_LAST, \
-    TOKEN_MODE_DISABLED, TOKEN_MODE_FINITE, TOKEN_MODE_INFINITE
+    TOKEN_MODE_DISABLED, TOKEN_MODE_FINITE, TOKEN_MODE_INFINITE, \
+    FEEDBACK_LEVEL_FULL, FEEDBACK_LEVEL_RESTRICTED
 from cms.db.validation import FilenameListConstraint
 
 from . import Base, Contest, CodenameConstraint, FilenameConstraint, \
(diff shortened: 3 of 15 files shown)


stefano-maggiolo:
2018-06-27-221441_1001x553_scrot
2018-06-27-221501_998x569_scrot

fagu:
I think this would be a very useful feature! The danger of reverse-engineering test cases is perhaps the main reason we sometimes used partial feedback (feedback for only a subset of the test cases) in our selection camps.

However, I'm a bit worried providing full feedback for decision problems might sometimes be exploitable even in safe mode: It should be easy to "solve" any decision problem with T test cases using about T log_5(T) submissions. The trivial lower bound for the number of submissions needed is about T / (log_2(5*T)). Where between these two bounds is the optimal strategy? What are safe bounds on the number of test cases and number of allowed submissions?

But without safe mode, a contestant would need only log_5(T) submissions, so it's already a **huge** improvement!

And a small unrelated question: Why do we bother to hide the signal of a compile error in safe mode? Isn't compilation independent of test data? Also, we do show the compilation message in any case, which should provide way more information to the contestant.

stefano-maggiolo:
Good point re compilation, I added back the signal. I'm a bit confused by your calculations (where does the 5 come from?) but decision problems are always going to be easier to revengineering, better compound multiple logical testcases in a physical one, or ask for a justification...

fagu:
It seems like there are ~6 possible results for a test case: timeout, timeout (Wall), signal, exit code != 0, wrong answer, correct
So there are roughly T*5 possible outcomes for any submission. (The index of the first failing testcase and the corresponding outcome.) There are 2^T possible maps {testcases} -> {0,1}, so a contestant would need at the very least log_{5T}(2^T)=T/log_2(5T) submissions to reverse-engineer the entire problem.
In any case, it's a rough back-of-the-envelope calculation. I just wanted to point out that the safe feedback doesn't solve all problems. :)

The safest but perhaps least friendly way would probably be to only tell contestants their score but nothing about specific test cases. (Perhaps tell them whether there was any timeout; if not, tell them whether there was any timeout(wall); if not, tell them whether there was any signal; ...)

One more point: as far as I see, the table stops after the first test case with a score less than 100%. (Assuming this is equivalent to outcome != "Correct"!?) So if a contestant gets 50% on test case 1 and 25% on test case 2, only test case 1 will be shown, but the total score will be 25%. That seems potentially confusing.

stefano-maggiolo:
I think what you're proposing in the middle was another possibility considered for the IOI; basically like the safe option but without providing the index of the first failing testcase, but it makes indeed really difficult to understand if you're making progress.

I agree with you on the possibility of confusion. Maybe instead of writing a simple N/A we should explain what is happening. Again this has the usual drawback that is easier for English-speaking contestants (in settings where translations are not used), so it should be something simple in case...

stefano-maggiolo:
Review status: all files reviewed, 8 unresolved discussions (waiting on @stefano-maggiolo)

---

*[a discussion](https://reviewable.io/reviews/cms-dev/cms/981#-LGP2Cga3pWANGiy0KsF:-LGXBGwEA7WSi1xFveXJ:b4yzfnl) (no related file):*

I don't like that the feedback level is passed to the workers through the jobs. It is needed by the tasktype running on the worker to produce the human evaluation message. I consider the feedback level to be just a rendering setting, which should have no effect on what data is stored in the DB. Is there a simple way that can accomplish this for the evaluation message?
</blockquote></details>

Before going to the actual review, let's try to settle this.

To be fair, the "rendering" is already in the DB by means of evaluation.text. If we wanted to follow your opinion, I guess we would need to introduce special strings for evaluation.text, that can be decoded by human_evaluation_message(), and at that point we might use the feedback level in the call (basically "signal" would point to two different strings depending on whether the feedback is full or not).

IMHO, it's quite a big change that is not fully justified by what we're doing here, given that we could have already decided to do the "rendering" of the message at display time rather than at evaluation time.

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/981)*

stefano-maggiolo:
So I wrote the other change, and I like it even less. In addition to the flaws already noted (DB not self-contained, need to support status codes forever), the change requires touching a lot of files (including those regarding printjobs, whose status is handled in the same way), and we end up having status-specific classes to handle the rendering, which I don't like.

Let me explain a bit more the last point: for example, in the evaluation:signal status, the feedback level changes the number of placeholders used in the rendering (in the safe mode we have none, in full mode we have one). So the rendering code must take this into account and ignore the value in safe mode. This code doesn't necessarily generalize to other statuses (we might have a class for "render using all values in full mode and none in safe mode", but that's hardly general). 

So all in all I think the product flaws and the bad code is enough to counterbalance the advantages (possibility to switch feedback on the fly, standardization).

wafrelka:
Could you modify this PR so that the limited feedback mode is automatically off when a contest enters its analysis round?

stefano-maggiolo:
Merged.

stefano-maggiolo:
For posterity, we eventually decided to remove the specific signal from all feedback level. The code is still in the history in case we need/want to revert this change.

PR #982, opened 2018-07-02
task_score() tests

stefano-maggiolo:
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/982)

### DIFF
--- a/cms/grading/__init__.py
+++ b/cms/grading/__init__.py
@@ -171,8 +171,8 @@ def task_score(participation, task):
         compute the score.
     task (Task): the task for which to compute the score.
 
-    return ((float, bool)): the score of user on task, and True if the
-        score could change because of a submission yet to score.
+    return ((float, bool)): the score of user on task, and True if not
+        all submissions of the participation in the task have been scored.
 
     """
     # As this function is primarily used when generating a rankings table
@@ -183,8 +183,7 @@ def task_score(participation, task):
     # submission_results table.  Doing so means that this function should incur
     # no exta database queries.
 
-    # If the score could change due to submission still being compiled
-    # / evaluated / scored.
+    # If some submission is yet to be scored.
     partial = False
 
     submissions = [s for s in participation.submissions
@@ -228,16 +227,14 @@ def task_score(participation, task):
 
         if last_sr is not None and last_sr.scored():
             last_score = last_sr.score
-        else:
-            partial = True
 
         for s in submissions:
-            if s.tokened():
-                sr = s.get_result(task.active_dataset)
-                if sr is not None and sr.scored():
+            sr = s.get_result(task.active_dataset)
+            if sr is not None and sr.scored():
+                if s.tokened():
                     max_tokened_score = max(max_tokened_score, sr.score)
-                else:
-                    partial = True
+            else:
+                partial = True
 
         score = max(last_score, max_tokened_score)
 
--- a/cms/grading/__init__.py
+++ b/cms/grading/__init__.py
@@ -37,7 +37,7 @@ from collections import namedtuple
 
 from sqlalchemy.orm import joinedload
 
-from cms import SCORE_MODE_MAX
+from cms import SCORE_MODE_MAX, SCORE_MODE_MAX_TOKENED_LAST
 from cms.db import Submission
 from cms.locale import DEFAULT_TRANSLATION
 
@@ -191,13 +191,11 @@ def task_score(participation, task):
                    if s.task is task and s.official]
     submissions.sort(key=lambda s: s.timestamp)
 
-    if submissions == []:
+    if len(submissions) == 0:
         return 0.0, False
 
-    score = 0.0
-
     if task.score_mode == SCORE_MODE_MAX:
-        # Like in IOI 2013-: maximum score amongst all submissions.
+        # Like in IOI 2013-2016: maximum score amongst all submissions.
 
         # The maximum score amongst all submissions (not yet computed
         # scores count as 0.0).
@@ -211,7 +209,8 @@ def task_score(participation, task):
                 partial = True
 
         score = max_score
-    else:
+
+    elif task.score_mode == SCORE_MODE_MAX_TOKENED_LAST:
         # Like in IOI 2010-2012: maximum score among all tokened
         # submissions and the last submission.
 
@@ -233,13 +232,16 @@ def task_score(participation, task):
             partial = True
 
         for s in submissions:
-            sr = s.get_result(task.active_dataset)
             if s.tokened():
+                sr = s.get_result(task.active_dataset)
                 if sr is not None and sr.scored():
                     max_tokened_score = max(max_tokened_score, sr.score)
                 else:
                     partial = True
 
         score = max(last_score, max_tokened_score)
+
+    else:
+        raise ValueError("Unknown score mode '%s'" % task.score_mode)
 
     return score, partial
--- a/cmstestsuite/unit_tests/databasemixin.py
+++ b/cmstestsuite/unit_tests/databasemixin.py
@@ -71,13 +71,14 @@ class DatabaseObjectGeneratorMixin(object):
     This is to be preferred to DatabaseMixin when a session is not required, in
     order to save some initialization and cleanup time.
 
-    The methods in this mixin are (mostly) static; we use a mixin to keep them
-    together and avoid the need to import a lot of names.
+    The methods in this mixin are static (actually, class methods to allow
+    overriding); we use a mixin to keep them together and avoid the need to
+    import a lot of names.
 
     """
 
-    @staticmethod
-    def get_contest(**kwargs):
+    @classmethod
+    def get_contest(cls, **kwargs):
         """Create a contest"""
         args = {
             "name": unique_unicode_id(),
@@ -87,11 +88,10 @@ class DatabaseObjectGeneratorMixin(object):
         contest = Contest(**args)
         return contest
 
-    @staticmethod
-    def get_announcement(contest=None, **kwargs):
+    @classmethod
+    def get_announcement(cls, contest=None, **kwargs):
         """Create an announcement"""
-        contest = contest if contest is not None \
-            else DatabaseObjectGeneratorMixin.get_contest()
+        contest = contest if contest is not None else cls.get_contest()
         args = {
             "contest": contest,
             "subject": unique_unicode_id(),
@@ -102,8 +102,8 @@ class DatabaseObjectGeneratorMixin(object):
         announcement = Announcement(**args)
         return announcement
 
-    @staticmethod
-    def get_user(**kwargs):
+    @classmethod
+    def get_user(cls, **kwargs):
         """Create a user"""
         args = {
             "username": unique_unicode_id(),
@@ -115,13 +115,11 @@ class DatabaseObjectGeneratorMixin(object):
         user = User(**args)
         return user
 
-    @staticmethod
-    def get_participation(user=None, contest=None, **kwargs):
+    @classmethod
+    def get_participation(cls, user=None, contest=None, **kwargs):
         """Create a participation"""
-        user = user if user is not None \
-            else DatabaseObjectGeneratorMixin.get_user()
-        contest = contest if contest is not None \
-            else DatabaseObjectGeneratorMixin.get_contest()
+        user = user if user is not None else cls.get_user()
+        contest = contest if contest is not None else cls.get_contest()
         args = {
             "user": user,
             "contest": contest,
@@ -130,11 +128,11 @@ class DatabaseObjectGeneratorMixin(object):
         participation = Participation(**args)
         return participation
 
-    @staticmethod
-    def get_message(participation=None, **kwargs):
+    @classmethod
+    def get_message(cls, participation=None, **kwargs):
         """Create a message."""
         participation = participation if participation is not None \
-            else DatabaseObjectGeneratorMixin.get_participation()
+            else cls.get_participation()
         args = {
             "participation": participation,
             "subject": unique_unicode_id(),
@@ -146,11 +144,11 @@ class DatabaseObjectGeneratorMixin(object):
         message = Message(**args)
         return message
 
-    @staticmethod
-    def get_question(participation=None, **kwargs):
+    @classmethod
+    def get_question(cls, participation=None, **kwargs):
         """Create a question."""
         participation = participation if participation is not None \
-            else DatabaseObjectGeneratorMixin.get_participation()
+            else cls.get_participation()
         args = {
             "participation": participation,
             "subject": unique_unicode_id(),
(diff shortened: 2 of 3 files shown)


lerks:
Reviewed 1 of 1 files at r1, 2 of 2 files at r2.
Review status: all files reviewed, 5 unresolved discussions (waiting on @stefano-maggiolo)

---

*[cmstestsuite/unit_tests/databasemixin.py, line 94 at r1](https://reviewable.io/reviews/cms-dev/cms/982#-LGPKW7u1Zqzb1F2_8FZ:-LGPKW7u1Zqzb1F2_8F_:bmrknbu) ([raw file](https://github.com/cms-dev/cms/blob/70e5043dbd5609dd1aa36e24bdd2372e74804b5b/cmstestsuite/unit_tests/databasemixin.py#L94)):*
> ```Python
>         """Create an announcement"""
>         contest = contest if contest is not None \
>             else DatabaseObjectGeneratorMixin.get_contest()
> ```

If we made this a `classmethod` and used `cls` here instead of `DatabaseObjectGeneratorMixin` we would allow subclasses to override some of these methods. That's something we like, don't we?

(I know this was how we did it before, but since you touched the code you brought this to my attention again).

The same holds several times below.

---

*[cmstestsuite/unit_tests/databasemixin.py, line 432 at r1](https://reviewable.io/reviews/cms-dev/cms/982#-LGPLdf28ReA3b1Vl82t:-LGPLdf28ReA3b1Vl82u:b-rsy3vf) ([raw file](https://github.com/cms-dev/cms/blob/70e5043dbd5609dd1aa36e24bdd2372e74804b5b/cmstestsuite/unit_tests/databasemixin.py#L432)):*
> ```Python
>         submission_result = self.get_submission_result(
>             submission, dataset, **kwargs)
>         return submission_result
> ```

I think you accidentally removed the `self.session.add(submission_result)`.

---

*[cmstestsuite/unit_tests/grading/init_test.py, line 34 at r2](https://reviewable.io/reviews/cms-dev/cms/982#-LGPMMGIEiTfrF4UAZV0:-LGPMMGIEiTfrF4UAZV1:b-oyedbd) ([raw file](https://github.com/cms-dev/cms/blob/5e136104f47ab5f5259a087ac9cf2794ef23e010/cmstestsuite/unit_tests/grading/init_test.py#L34)):*
> ```Python
> from datetime import timedelta
> 
> from cmstestsuite.unit_tests.databasemixin import DatabaseMixin
> ```

Add comment explaining why it needs to be first?

---

*[cmstestsuite/unit_tests/grading/init_test.py, line 134 at r2](https://reviewable.io/reviews/cms-dev/cms/982#-LGPN_bo0TkJADxg2TEv:-LGPN_bo0TkJADxg2TEw:bx850ci) ([raw file](https://github.com/cms-dev/cms/blob/5e136104f47ab5f5259a087ac9cf2794ef23e010/cmstestsuite/unit_tests/grading/init_test.py#L134)):*
> ```Python
>         self.add_result(self.at(1), 44.4, tokened=True)
>         self.add_result(self.at(2), 66.6, tokened=True)
>         self.session.commit()
> ```

I am surprised that you need to commit here: `task_score` never creates a new Session thus it uses the one associated to the objects it receives as inputs; Alchemy flushes the data of a Session before each query thus the database is aware of it when it runs the query (even if the data still only exists inside an uncommitted transaction).

Have you tried without committing? Does it break?

(Holds for all test methods below too)

---

*[cmstestsuite/unit_tests/grading/init_test.py, line 238 at r2](https://reviewable.io/reviews/cms-dev/cms/982#-LGPOVT14eLKk1SyUVOJ:-LGPOVT14eLKk1SyUVOK:bqh7j01) ([raw file](https://github.com/cms-dev/cms/blob/5e136104f47ab5f5259a087ac9cf2794ef23e010/cmstestsuite/unit_tests/grading/init_test.py#L238)):*
> ```Python
>         self.add_result(self.at(2), None, tokened=False)
>         self.session.commit()
>         self.assertEqual(self.call(), (0.0, True))
> ```

"Morally", if a submission result already achieved the maximum score, then when using the `max` method `task_score` should not say that the score is partial even when there are unscored submission results. Is it something we take into account and should test for?

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/982)*

stefano-maggiolo:
Review status: all files reviewed, 5 unresolved discussions (waiting on @lerks and @stefano-maggiolo)

---

*[cmstestsuite/unit_tests/databasemixin.py, line 94 at r1](https://reviewable.io/reviews/cms-dev/cms/982#-LGPKW7u1Zqzb1F2_8FZ:-LGRhczb52YMD6yETAmf:bqv1y5b) ([raw file](https://github.com/cms-dev/cms/blob/70e5043dbd5609dd1aa36e24bdd2372e74804b5b/cmstestsuite/unit_tests/databasemixin.py#L94)):*

If we made this a `classmethod` and used `cls` here instead of `DatabaseObjectGeneratorMixin` we would allow subclasses to override some of these methods. That's something we like, don't we?

(I know this was how we did it before, but since you touched the code you brought this to my attention again).

The same holds several times below.
</blockquote></details>

Done in a commit I will keep separate.

---

*[cmstestsuite/unit_tests/databasemixin.py, line 432 at r1](https://reviewable.io/reviews/cms-dev/cms/982#-LGPLdf28ReA3b1Vl82t:-LGRix-63wL9ifl3qebW:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/70e5043dbd5609dd1aa36e24bdd2372e74804b5b/cmstestsuite/unit_tests/databasemixin.py#L432)):*

I think you accidentally removed the `self.session.add(submission_result)`.
</blockquote></details>

Done.

---

*[cmstestsuite/unit_tests/grading/init_test.py, line 34 at r2](https://reviewable.io/reviews/cms-dev/cms/982#-LGPMMGIEiTfrF4UAZV0:-LGRl5hhAYnM9ha_fw6G:b-896fix) ([raw file](https://github.com/cms-dev/cms/blob/5e136104f47ab5f5259a087ac9cf2794ef23e010/cmstestsuite/unit_tests/grading/init_test.py#L34)):*

Add comment explaining why it needs to be first?
</blockquote></details>

Done.

---

*[cmstestsuite/unit_tests/grading/init_test.py, line 134 at r2](https://reviewable.io/reviews/cms-dev/cms/982#-LGPN_bo0TkJADxg2TEv:-LGRlVU22sG5pTL3B-iv:b-t7om2w) ([raw file](https://github.com/cms-dev/cms/blob/5e136104f47ab5f5259a087ac9cf2794ef23e010/cmstestsuite/unit_tests/grading/init_test.py#L134)):*

I am surprised that you need to commit here: `task_score` never creates a new Session thus it uses the one associated to the objects it receives as inputs; Alchemy flushes the data of a Session before each query thus the database is aware of it when it runs the query (even if the data still only exists inside an uncommitted transaction).

Have you tried without committing? Does it break?

(Holds for all test methods below too)
</blockquote></details>

Upon lengthy investigation... the simplest that breaks (but there are many) is `TestTaskScoreMax.test_all_submissions_scored`. It adds two submission & results to the session, task_score sees the 2 submissions, but is only able to find the result of the last (that is, get_result() returns None for the first unless there is the commit).

The reason is that before the first query, submissions have no ids; the first query fails but also apparently flushes, so after that ids are present. Might be a bug?

Anyway, if you prefer I can change all the commit() to flush() and avoid the delete_data(). WDYT?

---

*[cmstestsuite/unit_tests/grading/init_test.py, line 238 at r2](https://reviewable.io/reviews/cms-dev/cms/982#-LGPOVT14eLKk1SyUVOJ:-LGRl7Mq97TGUSP5w0Rp:bmq3z3k) ([raw file](https://github.com/cms-dev/cms/blob/5e136104f47ab5f5259a087ac9cf2794ef23e010/cmstestsuite/unit_tests/grading/init_test.py#L238)):*

"Morally", if a submission result already achieved the maximum score, then when using the `max` method `task_score` should not say that the score is partial even when there are unscored submission results. Is it something we take into account and should test for?
</blockquote></details>

Yeah, I was indeed a bit surprised when writing the test about the unscored but irrelevant submission.

Given it's used only to show a note in the admin ranking, and looking at the max score seems a bit brittle, how about we change the semantic of the return value to say "True if there is at least one submission which hasn't been scored"?

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/982)*

lerks:
Reviewed 1 of 1 files at r3, 1 of 1 files at r4, 1 of 1 files at r5.
Review status: all files reviewed, 3 unresolved discussions (waiting on @lerks and @stefano-maggiolo)

---

*[cmstestsuite/unit_tests/databasemixin.py, line 81 at r3](https://reviewable.io/reviews/cms-dev/cms/982#-LGS-rOL1XxeGwR4dl6A:-LGS-rOL1XxeGwR4dl6B:b-o1w8t4) ([raw file](https://github.com/cms-dev/cms/blob/f75c593098cd791cf5edc3a5632fa9cdba3a7dc0/cmstestsuite/unit_tests/databasemixin.py#L81)):*
> ```Python
> 
>     @classmethod
>     def get_contest(cls, **kwargs):
> ```

This, `get_user` and `get_task` can actually be static. (I believe that one can override a static method with a class method, right?)

---

*[cmstestsuite/unit_tests/grading/init_test.py, line 134 at r2](https://reviewable.io/reviews/cms-dev/cms/982#-LGPN_bo0TkJADxg2TEv:-LGS1foC5ziHwosGoU_Y:b67rhmv) ([raw file](https://github.com/cms-dev/cms/blob/5e136104f47ab5f5259a087ac9cf2794ef23e010/cmstestsuite/unit_tests/grading/init_test.py#L134)):*

Upon lengthy investigation... the simplest that breaks (but there are many) is `TestTaskScoreMax.test_all_submissions_scored`. It adds two submission & results to the session, task_score sees the 2 submissions, but is only able to find the result of the last (that is, get_result() returns None for the first unless there is the commit).

The reason is that before the first query, submissions have no ids; the first query fails but also apparently flushes, so after that ids are present. Might be a bug?

Anyway, if you prefer I can change all the commit() to flush() and avoid the delete_data(). WDYT?
</blockquote></details>

Thanks for the investigation! I definitely prefer explicit flushing to explicit committing. However, given what you say, I believe the proper solution (that would also benefit the rest of the code) would be to flush in `Base.get_from_id` before accessing any `.id` field. What do you think?

---

*[cmstestsuite/unit_tests/grading/init_test.py, line 238 at r2](https://reviewable.io/reviews/cms-dev/cms/982#-LGPOVT14eLKk1SyUVOJ:-LGS26skC056F9L6vyn7:bhy1qd2) ([raw file](https://github.com/cms-dev/cms/blob/5e136104f47ab5f5259a087ac9cf2794ef23e010/cmstestsuite/unit_tests/grading/init_test.py#L238)):*

Yeah, I was indeed a bit surprised when writing the test about the unscored but irrelevant submission.

Given it's used only to show a note in the admin ranking, and looking at the max score seems a bit brittle, how about we change the semantic of the return value to say "True if there is at least one submission which hasn't been scored"?
</blockquote></details>

Yeah, this is really minor and the fixes aren't elegant (I had never fully realized that maximum scores are only defined per submission, and not per task). I'm ok with leaving it as it is.

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/982)*

lerks:
Review status: all files reviewed, 2 unresolved discussions (waiting on @lerks and @stefano-maggiolo)

---

*[cmstestsuite/unit_tests/grading/init_test.py, line 134 at r2](https://reviewable.io/reviews/cms-dev/cms/982#-LGPN_bo0TkJADxg2TEv:-LGUvQNz51IemGTKN_nc:b1beo4u) ([raw file](https://github.com/cms-dev/cms/blob/5e136104f47ab5f5259a087ac9cf2794ef23e010/cmstestsuite/unit_tests/grading/init_test.py#L134)):*

Thanks for the investigation! I definitely prefer explicit flushing to explicit committing. However, given what you say, I believe the proper solution (that would also benefit the rest of the code) would be to flush in `Base.get_from_id` before accessing any `.id` field. What do you think?
</blockquote></details>

Never mind, I thought about this again and I think I prefer that the flushing is done by the code that creates a new object just after it creates it rather than by the code that accesses it. So, in this case, you should just replace the commits by flushes.

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/982)*

stefano-maggiolo:
Review status: all files reviewed, 2 unresolved discussions (waiting on @stefano-maggiolo)

---

*[cmstestsuite/unit_tests/databasemixin.py, line 81 at r3](https://reviewable.io/reviews/cms-dev/cms/982#-LGS-rOL1XxeGwR4dl6A:-LGWyf9I6U0OoLsT781l:b-y8bjfz) ([raw file](https://github.com/cms-dev/cms/blob/f75c593098cd791cf5edc3a5632fa9cdba3a7dc0/cmstestsuite/unit_tests/databasemixin.py#L81)):*

This, `get_user` and `get_task` can actually be static. (I believe that one can override a static method with a class method, right?)
</blockquote></details>

Uhm, sure, whatever. (Yes, you can override.)

---

*[cmstestsuite/unit_tests/grading/init_test.py, line 134 at r2](https://reviewable.io/reviews/cms-dev/cms/982#-LGPN_bo0TkJADxg2TEv:-LGX6yQD2SvJe1Olwl-c:b-y9x8wp) ([raw file](https://github.com/cms-dev/cms/blob/5e136104f47ab5f5259a087ac9cf2794ef23e010/cmstestsuite/unit_tests/grading/init_test.py#L134)):*

Never mind, I thought about this again and I think I prefer that the flushing is done by the code that creates a new object just after it creates it rather than by the code that accesses it. So, in this case, you should just replace the commits by flushes.
</blockquote></details>

Upon deeper investigation: https://groups.google.com/forum/#!topic/sqlalchemy/ocXbxSHF__8 - get() doesn't trigger autoflush, apparently.

More precisely, looking at the code (thanks pycharm that makes it easy!): alchemy receives the get(), looks up in the identity map and doesn't find the object (because ids are not created yet). So it issues the query to the DB; that actually autoflushes, but  I suppose at that point the query issued to postgres only looks at persistent data (?! I didn't follow after that).

Obviously I don't care for this test - I wonder if we'd better switch to a filter() + one_or_none() (it becomes a bit more cumbersome because it needs to be specialized for classes such as SR with multiple fields for primary key). I think we're pretty safe because it seems strange to create an object and get_from_id' it in the same session... Also seems an error that would trigger, not something that stays hidden and happens sporadically.

For this test: I moved to flush() and avoided the delete_data().  

---

*[cmstestsuite/unit_tests/grading/init_test.py, line 238 at r2](https://reviewable.io/reviews/cms-dev/cms/982#-LGPOVT14eLKk1SyUVOJ:-LGXAjTQ4ab4YPDEBL8F:bojo2hs) ([raw file](https://github.com/cms-dev/cms/blob/5e136104f47ab5f5259a087ac9cf2794ef23e010/cmstestsuite/unit_tests/grading/init_test.py#L238)):*

Yeah, this is really minor and the fixes aren't elegant (I had never fully realized that maximum scores are only defined per submission, and not per task). I'm ok with leaving it as it is.
</blockquote></details>

Would you object changing the semantic like in the previous comment? In a different PR in case.

---


*Comments from [Reviewable](https://reviewable.io/reviews/cms-dev/cms/982)*

PR #986, opened 2018-08-04
Statistics feature for CWS

wil93:
This PR is a proposed patch to implement some basic stats visualization for CWS (that is, for contestants).

The feature can be enabled from AWS (admin) and, if enabled, it will show a graph in the contestant interface.

cws_stats

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/986)


lerks:
@stefano-maggiolo told me that you were working on something like this, but my understanding was that you were first going to submit a proposal to the ISC and the GA to see whether they wanted to use this at the IOI2018. I don't think we should merge this until we're sure someone wants it and is going to use it.

wil93:
I'd say this one is not necessary anymore. The feature was used in IOI2018 but dropped afterwards.

PR #990, opened 2018-08-16
Rename async to async_config

fagu:
Since python 3.7, `async` is a reserved keyword, so we get a parser error in `conf.py`.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/990)


lerks:
Thanks! I'll wait for the automatic tools' analysis and then I'll push.

PR #992, opened 2018-08-18
Some work on isolate

stefano-maggiolo:
Mostly, a quick fix for the build.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/992)


stefano-maggiolo:
I'm going to retire this PR, and post an updated one with a better fix soon. In short, the new PR ensures that all map have a constant mount point in the sandbox that we control and doesn't clash. We mount our inner_temp_dir for the sandbox to /tmp, so that compilers that need it have /tmp (this is the existing behaviour, is not great, but waiting for @gollux to see if we can unload the creation of /tmp to isolate).

PR #994, opened 2018-08-20
Really fix the build after isolate's update.

stefano-maggiolo:
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/994)


stefano-maggiolo:
I also would like to simplify names:
- self._inner_temp_dir --> disappears (no need to store it now that we have *_dest)
- self._inner_temp_dir_dest --> self._path_dest
- self._outer_temp_dir --> self._outer_dir

If you're ok I'm going to do it before pushing, issuing fixups for this is a bit complicated.

PR #998, opened 2018-08-22
Support having 2 tasks with the same name on DB

### LINKED ISSUE #765: Support having 2 tasks on DB with the same Task.name
I remember that this was discussed and @giomasce told me that all cms-dev agreed that it would be a good idea to allow duplicate Task.name

This is painful in situations where you have a "practice session" (e.g. IOI day0) where you may want to put old tasks, which are already on the DB.

The current workaround I'm using is to manually rename the old tasks to "old_<name>" and then importing the new ones.

wil93:
Currently we have three separate uniqueness constraints: (name), (name, contest_id), and (num, contest_id). The first one is not really useful and in some cases it's really annoying (current workaround is adding the `contestname_` as a prefix of taskname).

Fixes #765 

Before:

```
Indexes:
    "tasks_pkey" PRIMARY KEY, btree (id)
    "tasks_contest_id_name_key" UNIQUE CONSTRAINT, btree (contest_id, name)
    "tasks_contest_id_num_key" UNIQUE CONSTRAINT, btree (contest_id, num)
    "tasks_name_key" UNIQUE CONSTRAINT, btree (name)
    "ix_tasks_contest_id" btree (contest_id)
```

After:

```
Indexes:
    "tasks_pkey" PRIMARY KEY, btree (id)
    "tasks_contest_id_name_key" UNIQUE CONSTRAINT, btree (contest_id, name)
    "tasks_contest_id_num_key" UNIQUE CONSTRAINT, btree (contest_id, num)
    "ix_tasks_contest_id" btree (contest_id)
```

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/998)

### DIFF
--- a/cms/db/task.py
+++ b/cms/db/task.py
@@ -99,8 +99,7 @@ class Task(Base):
     # Short name and long human readable title of the task.
     name = Column(
         Codename,
-        nullable=False,
-        unique=True)
+        nullable=False)
     title = Column(
         Unicode,
         nullable=False)
--- a/cmscontrib/AddStatement.py
+++ b/cmscontrib/AddStatement.py
@@ -4,6 +4,7 @@
 # Contest Management System - http://cms-dev.github.io/
 # Copyright © 2016 Myungwoo Chun <mc.tamaki@gmail.com>
 # Copyright © 2016 Stefano Maggiolo <s.maggiolo@gmail.com>
+# Copyright © 2018 William Di Luigi <williamdiluigi@gmail.com>
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as
@@ -35,31 +36,29 @@ import sys
 import os
 
 from cms import utf8_decoder
-from cms.db import SessionGen, Statement, Task
+from cms.db import SessionGen, Statement
 from cms.db.filecacher import FileCacher
+from cmscontrib.importing import ImportDataError, task_from_db
 
 
 logger = logging.getLogger(__name__)
 
 
-def add_statement(task_name, language_code, statement_file, overwrite):
+def add_statement(task_name, task_id, language_code, statement_file,
+                  overwrite):
     logger.info("Adding the statement(language: %s) of task %s "
                 "in the database.", language_code, task_name)
 
     if not os.path.exists(statement_file):
-        logger.error("Statement file (path: %s) does not exist.",
-                     statement_file)
-        return False
+        raise ImportDataError("Statement file (path: %s) does not exist."
+                              % statement_file)
+
     if not statement_file.endswith(".pdf"):
-        logger.error("Statement file should be a pdf file.")
-        return False
+        raise ImportDataError("Statement file should be a pdf file.")
 
     with SessionGen() as session:
-        task = session.query(Task)\
-            .filter(Task.name == task_name).first()
-        if not task:
-            logger.error("No task named %s", task_name)
-            return False
+        task = task_from_db(session, task_name, task_id)
+
         try:
             file_cacher = FileCacher()
             digest = file_cacher.put_file_from_path(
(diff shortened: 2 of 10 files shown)


lerks:
Are you serious? Did you really not even try to fix all the things that would break because of this?

Here are just the first ones that came to my mind: AddTestcases, AddStatement, ImportTask, RemoveTask, RWS, ...

lerks:
I searched a bit more and also ImportContest and ImportDataset would break.

wil93:
I assume by "will break" you mean "will print a SQL integrity error instead of a user-friendly error message, in some cases" (especially given the fact that all workflows used until now, including the test suite, are working with no problems).

Anyway, OK, I will look into adding more user-friendly error messages. It must be said though that there are some situations where you can get this kind of SQL error printouts already (if you edit tasks in the wrong way, using AWS ...) so I would tend to assume that "contest admins" should mostly be fine with such errors.

lerks:
No, by "break" I mean it will become impossible to add a statement to a
task if there is another one with the same name in another contest, or that
the wrong task will be removed, or that a dataset will be added to the
wrong task, or that data will be overwritten. In short, tools will not do
what they're supposed to. Like, they will break.

wil93:
Oh, now I see, good points.

I guess I'll replace most of "task name" requests with "task id"...would that work? These are admin tools anyway so we can assume an admin knows how to find out the ID of a task. We could eventually add some scripts to ease data browsing from the command line (`cmsList (tasks|users|...)`) so that to retrieve the IDs one doesn't need to use AWS or psql.

stefano-maggiolo:
I would prefer a disambiguation (at least via command line argument, possibly interactive too) to avoid breaking existing scripts.

wil93:
OK, I will add an optional `-t <task_id>` argument that overrides the positional `task_name` when specified, and becomes mandatory when the task name is ambiguous.

stefano-maggiolo:
[Note this might not cover all issues, in some other places you might need to do something different...]

lerks:
I would prefer a `--contest-name`/`-c` option. We already have it, for example in AddTestcases, except that now it's only used to verify that the task with the given name belongs to that contest. So it would be nice to fix that behavior, and to extend that flag to all affected commands.

lerks:
As for RWS, it may be enough to just fix the PS side. PS should keep the `short_name` field of the task set to its name, but the identifier (the last component of the URL) should now become the combination of the contest name and the task name (with some precautions to avoid ambiguities of course).

lerks:
Also, in AWS there are some views where we list all tasks, for all contests, by name. We should now add their contests' names as well, to allow the admins to select without ambiguity.

wil93:
I think the id of the task must be specified in some cases. There can be 10 tasks with the same name and not tied to any contest

wil93:
OK, now almost everything works. Soon I will fix the rest.

```
(cms) wil93@xps13:~/git/cms$ cmsRemoveTask batch
2018-08-24 01:03:16,075 - INFO [<unknown>] Using configuration file /usr/local/etc/cms.conf.
2018-08-24 01:03:16,396 - ERROR [<unknown>] Task name is ambiguous, please use the -t argument to specify a valid ID:
 ID =   1 | batch (Batch) | con_test
 ID =  10 | batch (Batch) | <none>
 ID =  11 | batch (Batch) | <none>
2018-08-24 01:03:16,396 - INFO [<unknown>] Error while importing, no changes were made.
```

And this is when I specify an ID:

```
(cms) wil93@xps13:~/git/cms$ cmsRemoveTask batch -t 10
2018-08-24 01:04:25,519 - INFO [<unknown>] Using configuration file /usr/local/etc/cms.conf.
This will delete task `batch' and all related data, including submissions. Are you sure? [y/N] y
2018-08-24 01:04:28,023 - INFO [<unknown>] Task `batch' removed.
```

wil93:
Well it's not finished yet but the tools should work. AWS needs some aesthetic work but it's not broken. I didn't look into RWS yet.

PR #999, opened 2018-08-25
Make sure we can delete the sandbox once done

stefano-maggiolo:
In some systems, Java compilation creates directories and files; since
they are owned by the user within isolate, the user outside is unable
to delete them. This change essentially issues a chmod within isolate
when we want to later delete the whole sandbox, to make sure we will
be able to.

As a side effect, this merges cleanup() and delete(), that were anyway
very close already.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/999)

### DIFF
--- a/cms/grading/Sandbox.py
+++ b/cms/grading/Sandbox.py
@@ -518,18 +518,14 @@ class SandboxBase(with_metaclass(ABCMeta, object)):
         pass
 
     @abstractmethod
-    def cleanup(self):
+    def cleanup(self, delete=False):
         """Cleanup the sandbox.
 
         To be called at the end of the execution, regardless of
         whether the sandbox should be deleted or not.
 
-        """
-        pass
-
-    @abstractmethod
-    def delete(self):
-        """Delete the directory where the sandbox operated.
+        delete (bool): if True, also delete get_root_path() and everything it
+            contains.
 
         """
         pass
@@ -831,22 +827,12 @@ class StupidSandbox(SandboxBase):
         """
         return True
 
-    def cleanup(self):
-        """Cleanup the sandbox.
-
-        This sandbox needs no cleanup.
-
-        """
-        pass
-
-    def delete(self):
-        """Delete the directory where the sandbox operated.
-
-        """
-        logger.debug("Deleting sandbox in %s.", self._path)
-
-        # Delete the working directory.
-        rmtree(self._path)
+    def cleanup(self, delete=False):
+        """See Sandbox.cleanup()."""
+        # This sandbox doesn't have any cleanup, but we might want to delete.
+        if delete:
+            logger.debug("Deleting sandbox in %s.", self._path)
+            rmtree(self._path)
 
 
 class IsolateSandbox(SandboxBase):
@@ -950,9 +936,10 @@ class IsolateSandbox(SandboxBase):
         # symlink to one out of many alternatives.
         self.maybe_add_mapped_directory("/etc/alternatives")
 
-        # Tell isolate to get the sandbox ready. We do our best to
-        # cleanup after ourselves, but we might have missed something
-        # if the worker was interrupted in the middle of an execution.
+        # Tell isolate to get the sandbox ready. We do our best to cleanup
+        # after ourselves, but we might have missed something if a previous
+        # worker was interrupted in the middle of an execution, so we issue an
+        # idempotent cleanup.
         self.cleanup()
         self.initialize_isolate()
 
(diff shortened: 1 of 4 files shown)


stefano-maggiolo:
@gollux is there a more elegant way of achieving the same result?

stefano-maggiolo:
(sorry for the push -f, I hadn't included the latest (relevant) changed from master)

gollux:
Are you sure you are using "isolate --cleanup" properly? It should be able to delete everything created inside the sandbox. If you are writing to directories not belonging to the sandbox, please use the new temporary directory rules instead.

stefano-maggiolo:
The problem here are files created inside the sandbox in a directory mapped from elsewhere. We use cleanup to remove isolate's temp files, but the files in the mapped directory are not touched by cleanup, no? 

I'll take a look at the new rule.

gollux:
> The problem here are files created inside the sandbox in a directory mapped
> from elsewhere. We use cleanup to remove isolate's temp files, but the files
> in the mapped directory are not touched by cleanup, no? 

I still do not understand why you need creating the files elsewhere.
What prevents you from calling isolate --init, then populating the
sandbox directory with files you need, and after isolate --run copying
the results from the sandbox directory?

stefano-maggiolo:
Oh, I see /box is treated differently and all content is chowned at the beginning and back at the end - sorry for not having clear all of isolate inner working, but I'm learning :)

I think one reason not to use /box is that sometimes we want to keep the content around indefinitely, and in those cases we would have to copy it out before cleaning up. Nothing necessarily bad with that, just different from what we are doing now. 

Let me think if it makes sense for us to change behavior.

fagu:
When I run `cmsWorker` directly in a terminal, I now get the following type of error messages (ordinary task, C++ submission). The sandbox directory is still deleted, though...
```
... - INFO [Worker,0] [evaluate submission 93 on testcase 0056] Starting job.
/bin/chmod: changing permissions of '/tmp': Operation not permitted
/bin/chmod: changing permissions of '/tmp/output.txt': Operation not permitted
/bin/chmod: changing permissions of '/tmp/stderr.txt': Operation not permitted
/bin/chmod: changing permissions of '/tmp/input.txt': Operation not permitted
/bin/chmod: changing permissions of '/tmp/taskname': Operation not permitted
Exited with error status 1
... - INFO [Worker,0] [evaluate submission 93 on testcase 0056] Finished job.
```

stefano-maggiolo:
@fagu you're right, and those errors are not a problem (the files are already of the rightful owner). I will send a commit to suppress the output today.

stefano-maggiolo:
@gollux I think that changing CMS to use "the right way" is not going to change much in terms of complexity on our side (because of the "sometimes we want to persist the content of the sandbox" problem), so we'll stay as we are for the time being. 

If you release the /tmp change, we will be able to mount our filesystem elsewhere and thus very slightly decrease the possibility of rogue compilers doing strange things (e.g., while assuming that ./ != /tmp). In that case we should probably cleanup whatever /tmp is, which is easy.

gollux:
> @gollux I think that changing CMS to use "the right way" is not going
> to change much in terms of complexity on our side (because of the
> "sometimes we want to persist the content of the sandbox" problem), so
> we'll stay as we are for the time being. 

It does not change complexity, but it can easily change correctness :)

I really wouldn't assume that whatever strange things does the sandboxed
solution do in the filesyste, you can fix it by a simple recursive chmod.

I understand that you sometimes want to preserve some content between
runs in the sandbox, but I am pretty much certain that this should be
limited to well-specified contents instead of all files which were
created in the box directory :)

Both look like strong enough reasons to do a proper "isolate --cleanup"
between runs.

> If you release the /tmp change, we will be able to mount our
> filesystem elsewhere and thus very slightly decrease the possibility
> of rogue compilers doing strange things (e.g., while assuming that ./
> != /tmp). In that case we should probably cleanup whatever /tmp is,
> which is easy.  

Once you test that the /tmp change works for you, I will release it
officially.

Thanks!

stefano-maggiolo:
Actually the reason to save content is to debugging, so the more the content is the same as the one that failed, the easier to debug. Also note that we already do --cleanup across runs!

As for chmod vs using isolate's, I think that what isolate's does is a chown + rmtree, so not much different - the difference is that isolate's runs as root :D

stefano-maggiolo:
Forgot: feel free to release the /tmp change, it worked for me.

gollux:
> As for chmod vs using isolate's, I think that what isolate's does is a chown
> + rmtree, so not much different - the difference is that isolate's runs as
> root :D

That's one difference :)  Another is that all settings of permissions is done
under Isolate's control, so it knows what should be done to clean up the files.

gollux:
> Forgot: feel free to release the /tmp change, it worked for me.

Fine. I just released isolate-1.7.

PR #1004, opened 2018-08-26
Fix Polygon importer to work with both py2 and py3

alex65536:
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1004)


lerks:
Thanks!

I'm surprised that months after we started supporting py3 we're still finding issues like these... (luckily this time it was the other way around: it worked on py3 but was broken on py2).

PR #1013, opened 2018-09-06
RWS: return 0 after dropping stored data

tuankiet65:
In RankingWebServer's main(), after the data is dropped, the return code should return 0 (to signify success) instead of 1.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1013)


lerks:
Thanks!

PR #1014, opened 2018-09-07
Don't have RPC clients block on connect

lerks:
This is my version of the fix proposed in #814. I had shared a simpler version on Gitter but some more changes were required in order to make the tests pass.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1014)

### DIFF
--- a/cms/io/rpc.py
+++ b/cms/io/rpc.py
@@ -94,7 +94,7 @@ class RemoteServiceBase(object):
 
         """
         self.remote_address = remote_address
-        self.connected = False
+        self._connection_event = gevent.event.Event()
 
         self._on_connect_handlers = list()
         self._on_disconnect_handlers = list()
@@ -105,6 +105,15 @@ class RemoteServiceBase(object):
 
         self._read_lock = gevent.lock.RLock()
         self._write_lock = gevent.lock.RLock()
+
+    @property
+    def connected(self):
+        """Return whether we're connected to the other endpoint.
+
+        return (bool): the status of the connection.
+
+        """
+        return self._connection_event.is_set()
 
     def add_on_connect_handler(self, handler):
         """Register a callback for connection establishment.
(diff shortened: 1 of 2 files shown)


stefano-maggiolo:
LGTM, shall we ask @edomora97 and @tuankiet65 if this fixes their issues?

tuankiet65:
Thanks for having me, I'll try to test this out by next week (we're having a contest this Sunday so I don't feel like trying it out right now). Also can I just patch this directly on top of v1.3.2 or I have to use the master branch?

lerks:
@tuankiet65 Thanks for checking!

@stefano-maggiolo I realized that the unit test wasn't really testing the right thing, so I fixed it. I also "merged" `connected` and `connected_event`. PTAL

cmstestsuite/unit_tests/io/rpc_test.py
```diff
@@ -239,42 +239,56 @@ def test_method_return_list(self):
         self.assertTrue(result.successful())
         self.assertEqual(result.value, ["Hello", 42, "World"])
 
-    def test_background_connect(self):
-        self.kill_listener()
-        client = self.get_client(ServiceCoord("Foo", 0), block=False)
-        self.assertFalse(client.connected)
-        self.assertFalse(client.connected_event.is_set())
+    @patch("cms.io.rpc.gevent.socket.socket")
+    def test_background_connect(self, socket_mock):
```
stefano-maggiolo: Comment a bit what it's doing.

cms/io/rpc.py
```diff
@@ -94,8 +94,7 @@ def __init__(self, remote_address):
 
         """
         self.remote_address = remote_address
-        self.connected = False
-        self.connected_event = gevent.event.Event()
+        self.connected = gevent.event.Event()
```
stefano-maggiolo: This was a somehow public-ish api (hence all the places you had to change). Maybe a property connected that looks at _connected_event.is_set()?

stefano-maggiolo:
Nit: there are a few places in the test that you changed than reverted, and eventually only the formatting changed, if you want to revert them fully, feel free, otherwise I don't care that much.

PR #1016, opened 2018-09-09
Multi-source tasks and Python language

stefano-maggiolo:
I also manually tested locally with Python 3. For next version we
might want to move the tests to Python 3...

Initially the PR contained also the move of Pascal from "fpc
source_filenames[0]" to "fpc source_filenames [...]", but only worked
locally, travis' version doesn't like it.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1016)


stefano-maggiolo:
Thanks, done also the comment.

PR #1024, opened 2018-09-13
Subtask max score mode, as per IOI2017+

stefano-maggiolo:
UI suggestions are appreciated. Also thorough manual testing (I did it, but more eyes are better than fewer).

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1024)


lerks:
PS: since your PR has merge conflicts when rebased on master, Travis is refusing to test it, so please make sure the tests pass before pushing.

PR #1029, opened 2018-09-25
Do not use "average" of token modes when looking at a specific task

stefano-maggiolo:
Use instead the token mode of that task.

This might be a bit controversial, in the sense that it might have
been intended to show comments about the tokens for all tasks if at
least one allows tokens. But I think this is still clear (probably
more). An alternative could be to show in the task description and
before the submission rows "Tokens are not allowed on this tasks" if
tokens_contest != DISABLED && tokens_task == DISABLED && tokens_tasks
!= DISABLED.


stefano-maggiolo:
Discussing offline, we decided to revert the change in the description page because format_token_rules was already doing a decent job.

PR #1033, opened 2018-10-02
Improve handling of processes by RS

wil93:
ResourceService checks that each sub-process is already running and,
in that case, it does not start them. This is broken when trying to
run multiple CMS instances (with different virtual environments,
different databases, different cms.conf) because RS will think that
the process "cmsChecker" for example is already started.

To solve this, we now include the full path of each executable when
checking for already-started processes.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1033)

### DIFF
--- a/cms/service/ResourceService.py
+++ b/cms/service/ResourceService.py
@@ -139,7 +139,7 @@ class ProcessMatcher(object):
         if "python" not in cl_interpreter:
             return None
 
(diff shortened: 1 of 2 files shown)


wil93:
Any news on merging this? @stefano-maggiolo

stefano-maggiolo:
Apologies, this slipped in my inbox. Merging now.

PR #1034, opened 2018-10-03
Drop support for Python 2

lerks:
This PR removes any mention that we support Python 2 (from the setup.py, the doc, ...) and stops our CI tests on Python 2. It then proceeds to remove some of the libraries that we used to ensure intercompatibility (a notable exception is six, to be addressed at a later date).

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1034)


.github/CONTRIBUTING.md
```diff
@@ -19,7 +19,7 @@ We get around Python flexible type system in several ways:
 * we are fairly verbose with naming, trying to help the reader with following the types;
 * we follow our type annotation system for method and function docstrings (planning to switch to [PEP 484](https://www.python.org/dev/peps/pep-0484/) when we will remove support for Python 2); see later for the format.
 
-We support both Python 2 and 3, so we use the package six as required. In the medium future we will drop support to Python 2 and remove all temporary measures to support both at the same time. If you write some code that will need to be changed at that time, try to phrase it with six.
+We support Python 3 only.
```
stefano-maggiolo: Mention >= 3.6

cms/__init__.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
stefano-maggiolo: Does this commit makes the previous one ($(which python) in travis) useless?

lerks: The shebang only affects executions of the form `./foo.py`, whereas we used to explicitly invoke `python prerequisites.py`. I tried changing the Travis config to use `./prerequisites.py` and it does indeed work, however I am not sure it is doing what we expect: it is invoking a Python 3 interpreter, but probably not the one we think.

On Travis, `sudo` doesn't fully preserve the environment. Since we are in a venv, the `PATH` is modified so that `python` refers to the venv's Python interpreter. As proven by the bug I'm fixing here, within `sudo`, `python` resolves to Python 2: this means that inside `sudo` a different `PATH` is being used. (I haven't been unable to replicate this issue locally, I guess Travis is using a custom config for sudo and/or venvs). We've already faced this issue in f1d5955e8365827a15c57af89f06c5c5b1824de4.

If we now used `./prerequisites.py` with the `python3` shebang my guess is that we would be using the system Python 3 interpreter, not the venv one. I don't know what version the system interpreter is, but since it's Ubuntu 14.04 (trusty), it may very well be 3.4 which, although `prerequisites.py` doesn't do anything fancy, we probably don't want to use.

I may be totally wrong, but I think it is safer to stick with `$(which python)`.

lerks:
Merged.

PR #1039, opened 2018-10-10
Fix newlines in plain text ranking

andreyv:
Due to trim_blocks being enabled in Jinja environment, the plain text template didn't actually output newlines after each ranking entry. The simplest fix is to add the missing newline manually.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1039)


stefano-maggiolo:
I will merge this later today.

stefano-maggiolo:
Merged, thank you.

PR #1042, opened 2018-10-12
Last (?) set of broad changes following our becoming py3-only

lerks:
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1042)

### DIFF
--- a/cms/db/base.py
+++ b/cms/db/base.py
@@ -22,15 +22,15 @@
 import ipaddress
 from datetime import datetime, timedelta
 
+from sqlalchemy.dialects.postgresql import ARRAY, CIDR, JSONB, OID
 from sqlalchemy.ext.declarative import as_declarative
-from sqlalchemy.orm.exc import ObjectDeletedError
-from sqlalchemy.orm.session import object_session
 from sqlalchemy.orm import \
     class_mapper, object_mapper, ColumnProperty, RelationshipProperty
+from sqlalchemy.orm.exc import ObjectDeletedError
+from sqlalchemy.orm.session import object_session
 from sqlalchemy.types import \
     Boolean, Integer, Float, String, Unicode, Enum, DateTime, Interval, \
     BigInteger
-from sqlalchemy.dialects.postgresql import ARRAY, CIDR, JSONB, OID
 
 from . import engine, metadata, CastingArray, Codename, Filename, \
     FilenameSchema, FilenameSchemaArray, Digest
--- a/cms/db/base.py
+++ b/cms/db/base.py
@@ -60,7 +60,7 @@ _TYPE_MAP = {
 
 
 @as_declarative(bind=engine, metadata=metadata, constructor=None)
-class Base(object):
+class Base:
     """Base class for all classes managed by SQLAlchemy. Extending the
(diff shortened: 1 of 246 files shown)


cms/io/triggeredservice.py
```diff
@@ -35,7 +35,7 @@
 logger = logging.getLogger(__name__)
 
 
-class Executor(object, metaclass=ABCMeta):  # pylint: disable=R0921
+class Executor(metaclass=ABCMeta):  # pylint: disable=R0921
```
stefano-maggiolo: I think you can remove the R0921 thingie, now that you're using metaclass.

cms/grading/ParameterTypes.py
```diff
@@ -168,7 +168,7 @@ def __init__(self, name, short_name, description, values):
         values (dict): dictionary mapping each choice to a short description.
 
         """
-        super(ParameterTypeChoice, self).__init__(
+        super().__init__(
```
stefano-maggiolo: Unwrap.

cms/grading/ParameterTypes.py
```diff
@@ -209,7 +209,7 @@ def __init__(self, name, short_name, description, subparameters):
         subparameters ([ParameterType]): list of types of each sub-parameter.
 
         """
-        super(ParameterTypeCollection, self).__init__(
+        super().__init__(
```
stefano-maggiolo: Unwrap.

cms/db/base.py
```diff
@@ -22,15 +22,15 @@
 import ipaddress
 from datetime import datetime, timedelta
 
+from sqlalchemy.dialects.postgresql import ARRAY, CIDR, JSONB, OID
```
stefano-maggiolo: Nit: typo in commit title (incresse).

stefano-maggiolo: Also, isn't there some other place where ordering matters (I'm thinking cms/cms.log). Please double check.

stefano-maggiolo: It would be nice to have a command line tool to enforce/check this as we go on. Maybe pycharm uses something that we can use as well in a precommit hook?

cms/io/triggeredservice.py
lerks: Done. (The pylint warning was because we were defining an abstract class without immediately using it, however it seems that warning is gone, so the disabling can be removed)

cms/db/base.py
lerks: Will fix the commit when rebasing.

`cms/log.py` is correctly imported by `cms/__init__.py`. I did the following check: I looked for comments that were close to the lines affected by the commit and made sure that none of them said "needs to be first", "do not remove", etc. This showed that in fact PyCharm had removed an unused import, `cmsranking.Logger`, that we needed. So I readded it.

I don't know if we can use the PyCharm operation from the command line. And I don't know of other tools that do it.

lerks:
When looking at the messages from pylint I realized that another "fix" could be done, regarding checking for emptiness (in strings and containers). So I added it to this PR.

lerks:
Merged.

PR #1048, opened 2018-10-19
Fix bug with RWS and the new score mode

wil93:
Also slightly rewrite a max expression

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1048)


stefano-maggiolo:
Thanks, merged in master & 1.4 (after splitting the commits and rewording)

PR #1051, opened 2018-10-22
Update Latvian translation

andreyv:
Add new strings and improve existing translations.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1051)


lerks:
Merged, both in master and in v1.4. Thanks!

PR #1052, opened 2018-10-23
Update Vietnamese translations

kuroni:
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1052)

PR #1053, opened 2018-10-23
Update Vietnamese translations

kuroni:
Fix spacing issues on several multi-line translations.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1053)


lerks:
Thanks for your contribution! (And I appreciate the dedication you're
showing for achieving the best quality.) I've seen that you've pushed 5
different commits to this pull request, and did the same for #1052. While
we don't expect anyone to nail the perfect version right away and welcome
iterations, we don't think pull requests should be active development
grounds (apart from the notifications we receive for every change, it makes
it hard to figure out when the code is intended to be reviewed/merged,
which is what happened with #1052). If you need to make further changes to
this code could you do so offline and only update this PR when you're done
and ready for us to take a look? Thanks!

tuankiet65:
Since #1052 is merges this pull request might as well be closed?

lerks:
This PR was opened to fix up some issues in #1052. I'm waiting for @kuroni to confirm that this is done before I review it.

kuroni:
This is done actually, sorry for the late reply.

lerks:
Thanks, sorry for the misunderstanding, pushed to master and v1.4.

PR #1055, opened 2018-10-26
Fixed bug if the memory ends between 1000 and 1024 XiB

edomora97:
If the memory used (or other values used as parameter of format_size) ends in the range [1000, 1024) KiB/MiB/GiB... AWS and CWS will fail rendering that size and the submission details **wont be shown** for both the contestants and (partially) the admins! This because the number of digits of the number was supposed being at most 3, not true anymore since 1024 is used as the base.

Allowing an extra significant digit for those cases is the proposed solution.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1055)

### DIFF
--- a/cms/locale/locale.py
+++ b/cms/locale/locale.py
@@ -10,6 +10,7 @@
 # Copyright © 2014 Fabian Gundlach <320pointsguy@gmail.com>
 # Copyright © 2015 William Di Luigi <williamdiluigi@gmail.com>
 # Copyright © 2016 Myungwoo Chun <mc.tamaki@gmail.com>
+# Copyright © 2018 Edoardo Morassutto <edoardo.morassutto@gmail.com>
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as
--- a/cms/locale/locale.py
+++ b/cms/locale/locale.py
@@ -223,7 +223,8 @@ class Translation:
             if n < self.PREFIX_FACTOR:
                 f = copy.copy(self.locale.decimal_formats[None])
                 # We need int because floor returns a float in py2.
-                d = int(2 - math.floor(math.log10(n)))
+                # if 1000 <= n < 1024 d can be negative, cap to 0 decimals
+                d = max(0, int(2 - math.floor(math.log10(n))))
                 f.frac_prec = (d, d)
                 return (self.gettext(unit)
                         % babel.numbers.format_decimal(n, format=f,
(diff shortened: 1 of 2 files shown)


lerks:
I don't understand what you're fixing. We have unit tests for this and they pass:
```
self.assertEqual(ENGLISH.format_size(999), "999 bytes")
self.assertEqual(ENGLISH.format_size(1000), "1,000 bytes")
self.assertEqual(ENGLISH.format_size(1023), "1,023 bytes")
self.assertEqual(ENGLISH.format_size(1024), "1.00 KiB")
```
Could you provide a unit test that fails on master but passes with your fix?

wil93:
We had this traceback during a contest:

```
  File "/home/cmsuser/cms_venv/lib/python3.6/site-packages/cms-1.4rc0-py3.6.egg/cms/grading/scoretypes/abc.py", line 144, in get_html_details
    gettext=_, ngettext=n_)
  File "/home/cmsuser/cms_venv/lib/python3.6/site-packages/jinja2/asyncsupport.py", line 76, in render
    return original_render(self, *args, **kwargs)
  File "/home/cmsuser/cms_venv/lib/python3.6/site-packages/jinja2/environment.py", line 1008, in render
    return self.environment.handle_exception(exc_info, True)
  File "/home/cmsuser/cms_venv/lib/python3.6/site-packages/jinja2/environment.py", line 780, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/home/cmsuser/cms_venv/lib/python3.6/site-packages/jinja2/_compat.py", line 37, in reraise
    raise value.with_traceback(tb)
  File "<template>", line 79, in top-level template code
  File "/home/cmsuser/cms_venv/lib/python3.6/site-packages/cms-1.4rc0-py3.6.egg/cms/server/jinja2_toolbox.py", line 254, in format_size
    return translation.format_size(s)
  File "/home/cmsuser/cms_venv/lib/python3.6/site-packages/cms-1.4rc0-py3.6.egg/cms/locale/locale.py", line 245, in format_size
    locale=self.locale))
  File "/home/cmsuser/cms_venv/lib/python3.6/site-packages/babel/numbers.py", line 254, in format_decimal
    return pattern.apply(number, locale)
  File "/home/cmsuser/cms_venv/lib/python3.6/site-packages/babel/numbers.py", line 613, in apply
    self._format_frac(b or '0', locale, force_frac))
  File "/home/cmsuser/cms_venv/lib/python3.6/site-packages/babel/numbers.py", line 681, in _format_frac
    while len(value) > min and value[-1] == '0':
IndexError: string index out of range
```

The fix in this PR makes the error go away. I didn't know there were unit tests for this, maybe it has to do with the language used (the tests are in English, while we were most likely using Italian..)

lerks:
PS: This should also be backported to v1.4.

edomora97:
Thanks @wil93 for the datails I missed and sorry for being a bit cryptic!
I hope to be able to commit the unit test tomorrow morning :)

lerks:
Thanks, I added a few more tests (they are very cheap to run and may prevent new versions of this bug in the future) and pushed to master and v1.4.

PR #1057, opened 2018-11-01
RWS: add -y switch to allow unattended data drop

tuankiet65:
If the `-y` switch is specified then CMS would drop the data directory without asking for confirmation. This is desirable for automation scripts since without the switch, the script would have to pipe yes into the RWS process (`yes | cmsRankingWebServer -d`). `cmsDropDB` has this switch so I think RWS should also have one.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1057)

### DIFF
--- a/cmsranking/RankingWebServer.py
+++ b/cmsranking/RankingWebServer.py
@@ -528,7 +528,7 @@ def main():
     config.load(args.config)
 
     if args.drop:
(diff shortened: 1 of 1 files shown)


lerks:
Thanks, merged into master!

tuankiet65:
Wew, that was quick, thanks!

PR #1059, opened 2018-11-01
Ensure compiled translation .mo files are picked up during build phase

lerks:
There was a bug where simply running `./setup.py install` from a fresh checkout would compile the .po files into .mo files but wouldn't copy them to the `build/` subdirectory. Running `build` then `install` (or running `install` twice) would fix that issue as the files compiled by the first pass would get picked up at the second one.

This issue didn't affect the `develop` mode, since there the "installed" package is just a symlink to the source tree, where these files exist.

This needs to be backported to v1.4.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1059)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -107,6 +107,10 @@ def find_version():
 class build_py_and_l10n(build_py):
     def run(self):
         self.run_command("compile_catalog")
+        # The build command of distutils/setuptools searches the tree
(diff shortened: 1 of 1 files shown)


lerks:
LGTM given after private discussion.

lerks:
Merged into master and v1.4.

PR #1063, opened 2018-11-03
Fix whitespace in consecutive {% trans %} blocks

andreyv:
Due to active Jinja whitespace handling, whitespace was removed between consecutive `{% trans %}` blocks, resulting in translated sentences being glued together. Fix this by disabling the `lstrip_blocks` behaviour in relevant places.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1063)


lerks:
Very good catch, I had never noticed, thanks! Pushed to master and v1.4.

PR #1064, opened 2018-11-04
HTML fixes

andreyv:
Also add missing style for the empty print job table case. Perhaps the table may benefit from more extensive styling, like with the submission and the user test tables, but this change fixes the immediate visual problem.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1064)


lerks:
Thanks for doing these checks! Merged in master and v1.4.

PR #1066, opened 2018-11-04
prerequisites.py: fail properly when isolate doesn't build

andreyv:
Also fix other similar places and apply minor improvements to the make command.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1066)


stefano-maggiolo:
Merged.

PR #1070, opened 2018-11-06
RWS HTML fixes

andreyv:
This PR fixes some HTML validation warnings in RWS.

There are still three errors from the validator (check at https://validator.w3.org/):

1. Line 60, `<colgroup id="Scoreboard_cols"></colgroup>`: _Table column 1 established by element colgroup has no cells beginning in it._
2. Line 78, `Face`: _Bad value "" for attribute src on element img: Must be non-empty._
3. Line 82, `Flag`: _Bad value "" for attribute src on element img: Must be non-empty._

I assume that these will be filled by JavaScript later.

Perhaps there is a way to fix these without going out of one's way? For example, adding dummy placeholders for `src` or generating whole tags in JavaScript would seem useless and detrimental. Otherwise leaving the code as is should be good enough.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1070)

### DIFF
--- a/cmsranking/static/Ranking.html
+++ b/cmsranking/static/Ranking.html
@@ -1,32 +1,32 @@
 <!DOCTYPE html>
-<html>
+<html lang="en">
     <head>
         <meta charset='utf-8'/>
         <link rel="shortcut icon" href="img/favicon.ico" />
         <link rel='stylesheet' href='Ranking.css'>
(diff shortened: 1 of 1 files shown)


stefano-maggiolo:
For src="", does it improve by omitting src entirely?

Merging this one for the moment.

andreyv:
It doesn't — the `src` attribute is required on `img`.

Issue #1071, opened 2018-11-09
How to pass ERROR of ContestWebServer 


vuhatuananh:
How to pass ERROR of ContestWebServer (CMS 1.5.dev0 on Ubuntu 18.10)

2018-11-08 23:43:13,782 - INFO [Contest,0] ContestWebServer 0 up and running!
2018-11-08 23:43:20,661 - INFO [Contest,0] Unsuccessful cookie authentication: no cookie provided
2018-11-08 23:43:20,662 - ERROR [Contest,0] Uncaught exception GET / (::ffff:192.168.11.205)
HTTPServerRequest(protocol='http', host='192.168.11.239:8888', method='GET', uri='/', version='HTTP/1.1', remote_ip='::ffff:192.168.11.205', headers={'Host': '192.168.11.239:8888', 'User-Agent': 'Mozilla/5.0 (Windows NT 6.1; Win64; x64; rv:63.0) Gecko/20100101 Firefox/63.0', 'Accept': 'text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8', 'Accept-Language': 'en-US,en;q=0.5', 'Accept-Encoding': 'gzip, deflate', 'Connection': 'keep-alive', 'Cookie': '_xsrf=2|c25aa454|46837b25984a05f09d7cd915cc68d566|1541748421; awslogin="egupOjqnf4eg/r+X7M0YiutUdCQ=?id=MQ==&ip=Ijo6ZmZmZjoxOTIuMTY4LjExLjIwNSI=&timestamp=MTU0MTc0OTM4NS44MjY1OTQ2"', 'Upgrade-Insecure-Requests': '1', 'Cache-Control': 'max-age=0'})
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/dist-packages/tornado/web.py", line 1489, in _execute
    result = self.prepare()
  File "/usr/local/lib/python3.6/dist-packages/cms-1.5.dev0-py3.6.egg/cms/server/contest/handlers/contest.py", line 86, in prepare
    self.r_params = self.render_params()
  File "/usr/local/lib/python3.6/dist-packages/cms-1.5.dev0-py3.6.egg/cms/server/contest/handlers/contest.py", line 176, in render_params
    if self.current_user is not None:
  File "/usr/local/lib/python3.6/dist-packages/tornado/web.py", line 1158, in current_user
    self._current_user = self.get_current_user()
[...]

PR #1073, opened 2018-11-11
Fix custom PostgreSQL domains not being used in table definitions

lerks:
Spotted thanks to #1071.

Needs to be backported to v1.4.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1073)


lerks:
The reason for backporting is that we just got a bug report due to this.

I added a no-op updater that just prints a message and fails if any data is invalid. Is this what you had in mind?

PR #1075, opened 2018-11-13
Fix encoding issues in cmscommon/terminal.py

lucach:
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1075)


stefano-maggiolo:
Thanks, merged and backported to 1.4

Issue #1077, opened 2018-11-16
Various styling errors with CWS


kuroni:
**Description:**

1. When a submission receives 100/100 for a task with non-100 total but with decimal precision (eg. the total score is not 100 but 99.9999), the general scoring uses `score_0_100` format instead of `score_100`.

2. When a submission receives "Compilation failed" verdict, the loading GIF in the general scoring remains.

[Demo screenshots](https://imgur.com/a/jOGB2bq)

**CMS version:** v1.4.rc1
**Was CMS installed:** yes
**Using a virtualenv:** yes

stefano-maggiolo:
Also backported to v1.4. Thanks!

PR #1078, opened 2018-11-16
Fix inconsistent color coding of task’s score

kuroni:
The problem arises due to floating point imprecision, making float comparison faulty. Rounding such floats to be precise as much as 6 decimal digits seems to fix the problem.

Related to the first problem in #1077.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1078)


kuroni:
It did work, thank you very much for looking into this!

stefano-maggiolo:
Done and merged see #1077 for the list of commits.

PR #1079, opened 2018-11-23
AWS: Added score as seen by the user field to submission details

PPakalns:
In AWS for submission, there is a view of evaluation as seen by the user, but it doesn't provide a good summary of this view (sum of points).
Therefore I added a field that shows the score that is visible for the user for this submission.

This would allow to easier follow up contestants progress in contest and reason about their taken decisions based on the score visible to them.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1079)

### DIFF
--- a/cms/server/admin/templates/submission.html
+++ b/cms/server/admin/templates/submission.html
@@ -122,6 +122,18 @@
           {% endif %}
         </td>
       </tr>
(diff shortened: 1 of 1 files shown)


stefano-maggiolo:
Merged, thanks!

PR #1083, opened 2018-11-26
testlib improvements

andreyv:
This is a set of testlib CMS patch improvements.

- Main feature: print translatable messages by default (`translate:success` etc.) [1]
- Add the `CMS_VERBOSE_FEEDBACK` macro for printing detailed checker messages
- In verbose mode, print also the user message (for example, `OK answer is 3`)
- Improve partial points output
- For failure codes, `halt(1)` results in an evaluation failure, and the contestant never sees the checker message, so we don't need to output result or print `CONTACT JURY`

Translatable messages ensure a better user interface and hide implementation details — the user sees the same messages as they would see with the default white-diff comparator, meaning that tasks with checkers no longer look different from others.

I took the liberty of rearranging comments at the top to reduce their length. The new comment remains compatible with the testlib license.

CC @artikz @alex65536

[1] https://cms.readthedocs.io/en/latest/Task%20types.html#standard-manager-output

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1083)

### DIFF
--- a/cmscontrib/loaders/polygon/testlib-cms.patch
+++ b/cmscontrib/loaders/polygon/testlib-cms.patch
@@ -1,124 +1,108 @@
-*** testlib.h.orig	2018-08-22 12:26:04.104893746 +0300
---- testlib.h	2018-08-22 12:27:59.172759328 +0300
-***************
-*** 36,51 ****
---- 36,67 ----
-   * Permission to use or copy this software for any purpose is hereby granted 
-   * without fee, provided the above notices are retained on all copies.
-   * Permission to modify the code and to distribute modified code is granted,
-   * provided the above notices are retained, and a notice that the code was
-   * modified is included with the above copyright notice.
-   *
-   */
-  
-+ /*
-+  * Artem Iglikov
-+  *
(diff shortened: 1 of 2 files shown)


alex65536:
Nice improvements :)
By the way, did you mention me to ensure it's OK to rearrange the comment or just thought I'd be interested in this pull request?

andreyv:
> By the way, did you mention me to ensure it's OK to rearrange the comments at the top of the file or just thought I'd be interested in this pull request?

First reason for sure, but if you have any comments, then it would be nice too.

alex65536:
> First reason for sure

I approve the idea of rewriting this part of the changelog; now it's much clearer.

andreyv:
I also investigated testlib interactors (what would be called a "manager" for an interactive task in CMS). In theory they can be ported to CMS too, but this requires redirecting `std::cout` to a file from within testlib. This can be done using `std::cout.rdbuf()`, but this is a separate task and I decided not to do it in this PR.

stefano-maggiolo:
Merged, thank you

PR #1085, opened 2018-11-27
Store hashed password in cookie to avoid re-hashing at every request

### LINKED ISSUE #1076: User password is hashed with bcrypt on every request
Steps to reproduce:
- Create a contest
- Add a new user with a bcrypt-hashed password; add participation
- Log in to CWS with the new user
- After login, every HTTP request to CWS, including notifications, takes about 400±100 ms

Workaround:
- Set the user's password type to plain text

The slowdown can be observed in CWS logs:
```
2018-11-15 17:23:35,559 - INFO [Contest,0 11 authentication::_authenticate_request_from_cookie] Successful cookie authentication as user 'test', on contest test, returning from 2018-11-15 15:23:34.671650, at 2018-11-15 15:23:35.218682
2018-11-15 17:23:35,568 - INFO [Contest,0 11 web::log_request] 200 GET /test (::ffff:127.0.0.1) 349.93ms
```

Relevant code seems to be here (starts at `_authenticate_request_from_cookie()`):

https://github.com/cms-dev/cms/blob/2e0910b43f05cfafedb6278d492e06c386c182e1/cms/server/contest/authentication.py#L344
https://github.com/cms-dev/cms/blob/2e0910b43f05cfafedb6278d492e06c386c182e1/cms/server/contest/authentication.py#L67
https://github.com/cms-dev/cms/blob/2e0910b43f05cfafedb6278d492e06c386c182e1/cmscommon/crypto.py#L194

lerks:
The only drawback is that if admins change a user's password from plaintext to hashed or vice versa that user will have to log in again. Attackers knowing hashed passwords shouldn't be a problem as the server only accepts it if it's signed (together with a recent timestamp) with its secret key.

Fixes #1076.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1085)

### DIFF
--- a/cms/server/contest/authentication.py
+++ b/cms/server/contest/authentication.py
@@ -45,7 +45,7 @@ logger = logging.getLogger(__name__)
 
 
 def get_password(participation):
-    """Find out what password a participation is expected to provide.
+    """Return the password the participation can log in with.
 
     participation (Participation): a participation.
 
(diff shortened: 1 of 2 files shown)


stefano-maggiolo:
Merged

PR #1067, opened 2018-11-05
Add basic registration form

wil93:
I think the absence of this feature is hindering CMS adoption to a too great extent.

This PR introduces a new setting (defaults to False) for contests: `allow_registration`. If enabled, it will let users autonomously choose a password, create an account and participate in the contest.

The admin is thus freed from the responsibility to:

- create the account
- communicate the credentials

Screenshots:

screenshot from 2018-11-05 03-36-10
screenshot from 2018-11-05 03-37-06
screenshot from 2018-11-05 03-37-30

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1067)

PR #1086, opened 2018-11-29
Implement support for task statement in HTML and MD

wil93:
I tested adding (from AWS and from CLI) and removing (from AWS), and tested the rendering.

If you like this change, I think the only change needed is to write the DB updater (and "vendorize" the new JS dependencies, I didn't do it here because I didn't want to push big files).

- [ ] make sure that AWS supports uploading the new statement types
- [ ] fix cmsAddStatement not fetching assets

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1086)


CristianCantoro:
FYI, we are already using Markdown for task statements. We using `pandoc` and a `.tex` template in the middle, so we generate the task statements separately with a `Makefile`.

stefano-maggiolo:
Let's try to merge #1067 first

PR #1087, opened 2018-12-03
Fix reporting of used memory

andreyv:
In previous psutil versions the "used" field of memory stats included also buffers and cache. However, this is no longer the case starting from psutil 4.4. Now the output matches the `free(1)` utility, which reports used memory without buffers and cache.

Reference: giampaolo/psutil@ec7957c7bfd2abef1042bf4f2f72f586400c0d24

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1087)


stefano-maggiolo:
LGTM

stefano-maggiolo:
Merged

PR #1088, opened 2018-12-04
add install dev-requirements

karaketir16:
python3 setup.py install give error messages without installing dev-requirements

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1088)

### DIFF
--- a/docs/Installation.rst
+++ b/docs/Installation.rst
@@ -147,6 +147,7 @@ Assuming you have ``pip`` installed, you can do this:
 .. sourcecode:: bash
 
     sudo pip3 install -r requirements.txt
+    sudo pip3 install -r dev-requirements.txt
     sudo python3 setup.py install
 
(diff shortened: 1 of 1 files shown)


wil93:
Maybe another solution could be to find which dependency is required in order to run the install command, and move it from dev-requirements to the normal requirements

karaketir16:
Probably these dependencies are required when installed development version. Maybe it could be add "if you are installing development version install dev-requirements too".

wil93:
Mmm, wait: I think the `dev-requirements.txt` file is the list of dependencies that you need "if you are a CMS developer" and want to run tests, release a new version and such. It's not intended for the development version of CMS.

The normal `requirements.txt` file, instead, is for people that just want to "run CMS".

If the `setup.py install` command is failing then it means that the normal `requirements.txt` is missing something.

stefano-maggiolo:
I confirm what @wil93 wrote - @karaketir16, would you please find out what's missing from requirements.txt? Thanks!

karaketir16:
I couldn't install because I missed "# Only if you are going to use pip/venv to install python dependencies"
This is not about requirements.txt

PR #1093, opened 2019-01-07
Fixed a minor typo in the comments

eljakim:
@stefano-maggiolo as promised, a minor pull request that fixes a single typo.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1093)


stefano-maggiolo:
Thanks! BTW, don't worry about the code coverage check, we have yet to understand why but it always gives some discrepancy.

stefano-maggiolo:
Merged.

PR #1094, opened 2019-01-07
Show number of days left for long lasting contests.

eljakim:
When a contest lasts for a couple of weeks, showing the exact number of seconds counting down is strange and distracting. This change keeps everything as it was if there are less than 72 hours left in a contest. However, when there is more time left it only displays the number of days left.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1094)

### DIFF
--- a/cms/server/contest/static/cws_utils.js
+++ b/cms/server/contest/static/cws_utils.js
@@ -223,7 +223,7 @@ CMS.CWSUtils.prototype.format_timedelta = function(timedelta) {
     }
     else {
 	// otherwise only return the number of days.
-        return days + " days ";
+        return days + $("#translation_days").text();
     }
 };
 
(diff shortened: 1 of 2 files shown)


karliss:
Not everyone speaks uses English language. "days" should use the same translation mechanism as everything else.

wil93:
Maybe a good solution could be to show DD:HH:MM:SS instead of HH:MM:SS

No translation mechanism needed, would still retain the "precise" feeling, while at the same time solving the problem of "72 hours left"...

OR... we could just use a library to handle both the "humanization" of the time and its localization (e.g. moment or luxon)

eljakim:
> Maybe a good solution could be to show DD:HH:MM:SS instead of HH:MM:SS
> 
> No translation mechanism needed, would still retain the "precise" feeling, while at the same time solving the problem of "72 hours left"...
> 
The issue I have with the current system (besides the many hours that are left for a contest that lasts for two weeks) is that the ticking hours, minutes and seconds distract from the fact that they have plenty of time left.

I can definitely live with using moment.js or something similar, but even then you'd have to figure out what you want to show when. The way the system works is great for short term contests, but a bit over precise for longer term things.

PR #1102, opened 2019-01-22
ProxyService support for different contest RWS

PPakalns:
ProxyService only supports one parallel contest when sending initial data to listed RWS. After that, it sends all submission changes (even from different contests). These changes are discarded by RWS because of data inconsistency.

This pull request adds support to ProxyService for sending scores to multiple RWS where each of them can show a different subset of contests.
This request fixes the problem where ProxyService sends unrelated contest submissions to RWS.

Additionally, this allows to easily set up RW servers for multiple contests that are run in parallel and keep their data unmerged if needed.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1102)

### DIFF
--- a/cms/service/ResourceService.py
+++ b/cms/service/ResourceService.py
@@ -179,10 +179,6 @@ class ResourceService(Service):
         self._prev_cpu_times = self._get_cpu_times()
         # Sorted list of ServiceCoord running in the same machine
         self._local_services = self._find_local_services()
-        if "ProxyService" in (s.name for s in self._local_services) and \
-                self.contest_id is None:
-            logger.warning("Will not run ProxyService "
-                           "since it requires a contest id.")
         # Dict service with bool to mark if we will restart them.
         self._will_restart = dict((service,
                                    None if not self.autorestart else True)
@@ -222,10 +218,7 @@ class ResourceService(Service):
         matcher = ProcessMatcher()
         for service in self._local_services:
             # We let the user start logservice and resourceservice.
-            if service.name == "LogService" or \
-                    service.name == "ResourceService" or \
-                    (self.contest_id is None and
-                     service.name == "ProxyService"):
+            if service.name in ("LogService", "ResourceService"):
                 continue
 
             # If the user specified not to restart some service, we
@@ -462,10 +455,6 @@ class ResourceService(Service):
         except ValueError:
             logger.error("Unable to decode service string.")
         name = service[:idx]
-
-        # ProxyService requires contest_id
-        if self.contest_id is None and name == "ProxyService":
-            return None
 
         try:
             shard = int(service[idx + 1:])
--- a/config/cms.conf.sample
+++ b/config/cms.conf.sample
@@ -151,12 +151,14 @@
 
     "_section": "ScoringService",
 
-    "_help": "List of URLs (with embedded username and password) of the",
+    "_help": "List of [contest id, URL] tuples.",
+    "_help": "Contest id specifies which contest scores to send.",
+    "_help": "URL (with embedded username and password) of the",
     "_help": "RWSs where the scores are to be sent. Don't include the",
     "_help": "load balancing proxy (if any), just the backends. If any",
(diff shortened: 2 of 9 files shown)


cms/service/ResourceService.py
```diff
@@ -222,10 +218,7 @@ def _restart_services(self):
         matcher = ProcessMatcher()
         for service in self._local_services:
             # We let the user start logservice and resourceservice.
-            if service.name == "LogService" or \
-                    service.name == "ResourceService" or \
-                    (self.contest_id is None and
-                     service.name == "ProxyService"):
+            if service.name in ("LogService", "ResourceService"):
```
andreyv: It should be more conventional to use a list here.

config/cms.conf.sample
```diff
@@ -151,12 +151,14 @@
...
-    "_help": "List of URLs (with embedded username and password) of the",
+    "_help": "List of [contest id, URL] tuples.",
+    "_help": "Contest id specifies which contest scores to send.",
+    "_help": "URL (with embedded username and password) of the",
     "_help": "RWSs where the scores are to be sent. Don't include the",
     "_help": "load balancing proxy (if any), just the backends. If any",
     "_help": "of them uses HTTPS specify a file with the certificates",
     "_help": "you trust.",
-    "rankings": ["http://usern4me:passw0rd@localhost:8890/"],
+    "rankings": [[1, "http://usern4me:passw0rd@localhost:8890/"]],
```
andreyv: This somewhat breaks the current workflow, because now you have to update `cms.conf` every time your contests change.

I suppose that the _real_ fix would be to make RWS contest-aware, that is, serve multiple results pages under a prefix when multi-contest mode is on, similarly to how CWS works now. (Although this would interfere with RWS's feature of joining "day 1" and "day 2" contests…)

Maybe we could use a default of `0` or `-1` here to indicate that the old behavior should be used, so that users with existing workflows don't need to do anything. I would appreciate opinions on this.

wil93:
Did you consider changing the DB model so that a contest can provide a ranking_url as well as other fields already provided like title and description? If the field is null we can fallback to the ranking specified in cms.conf

In this way, proxyservice could send contest data to the right URL for each contest.

If there are multi-day contests we can easily support that by setting the same ranking_url value for both days

andreyv:
Or leave the upload URL in the config file, but for the contest specify a "contest group" it belongs to, and use this group as the RWS prefix. Empty group would mean the current behavior. Very nice idea, @wil93 :)

andreyv:
There is one more problem with multiple RankingWebServers, if you run them on the same machine. Not taking into account that each needs a different port number, they also need different log and data directories. By default they all try to use `/var/local/log/cms/ranking` and `/var/local/lib/cms/ranking`.

This can be worked around by specifying different `log_dir` and `lib_dir` settings and having a separate config file for each RWS instance (overridden using `CMS_RANKING_CONFIG`).

config/cms.conf.sample
PPakalns: Additionally, with change to "rankings" property format in config file, `cms/cmscontrib/RWSHelper.py` retrieving of ranking server urls needs to be fixed.

andreyv:
To expand on the last idea:

We can add a new "group" field to the contest in DB. Contests in one group would be temporally related, for example, day 1 and day 2. When ProxyService sends data to RWS, it would also send the contest's group along the data.

When RWS receives the data, it would display it under the `/<group>` URL prefix. An empty group would mean the current behavior.

For example, for the Latvian olympiad with two days and junior/senior student groups, we would have
```
Contest       Group
===========   ======
junior_day1   junior
senior_day1   senior
junior_day2   junior
senior_day2   senior
```
and there would be two scoreboards at http://rws/junior and http://rws/senior.

This way there is no need to change `cms.conf`/`cms.ranking.conf` each time. Also, as shown above, running multiple RWSs is troublesome too.

A separate question is how to deal with the leaking of the task names from the second day. I think this can be handled in a different issue.

PR #1103, opened 2019-01-27
Call super().prepare() before raising HTTPError in choose_contest()

### LINKED ISSUE #1069: "url" is None during error template rendering
Steps to reproduce:

1. Assume CWS is running in multi-contest mode and listening on `localhost:8888`
2. Run `telnet localhost 8888`
3. Request a non-existing contest page:
```
GET /zzz HTTP/1.1
Host: localhost

```
4. Actual output:
```
HTTP/1.1 404 Not Found
Server: TornadoServer/4.5.3
Content-Type: text/html; charset=UTF-8
Date: Tue, 06 Nov 2018 18:25:50 GMT
Content-Length: 224
Set-Cookie: zzz_login=""; expires=Mon, 06 Nov 2017 18:25:50 GMT; Path=/
Set-Cookie: _xsrf=2|47f2b7e0|d92adbae5745d453d5769a29a88df401|1541528750; Path=/

<!DOCTYPE html>
<html lang="en">
    <head>
        <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
        <meta name="viewport" content="width=device-width, initial-scale=1.0">

        <title>Error 404***OUTPUT STOPS HERE***
```
5. After that the server is still listening and responds to further requests on the same connection.

Environment: Ubuntu 18.04 with Python dependencies installed from the system; CMS is built from master and runs with Python 3.

We can also see that the server tries to set cookies for the non-existent contest.

andreyv:
If `choose_contest()` raises an exception (for example, contest doesn't exist), then the server will try to render an error template in `write_error()`. The template needs an `url` parameter, which is assigned in `CommonRequestHandler.prepare()`, so we need to call this function beforehand.

Partially fixes #1069.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1103)

### DIFF
--- a/cms/server/contest/handlers/contest.py
+++ b/cms/server/contest/handlers/contest.py
@@ -106,6 +106,7 @@ class ContestHandler(BaseHandler):
                 # render_params in this class assumes the contest is loaded,
                 # so we cannot call it without a fully defined contest. Luckily
                 # the one from the base class is enough to display a 404 page.
+                super().prepare()
                 self.r_params = super().render_params()
                 raise tornado.web.HTTPError(404)
         else:


stefano-maggiolo:
There's a small problem, after this change one can bypass the localization limitation set in the contest (because setup_locale() is done when available_localization is not yet limited).

The "solution" (ugly, but what can you do) is already there since it was done for render_params(): just call the required methods before raising. I think calling super().prepare() before render_params() there works. Can you try and amend if works for you too?

andreyv:
Thanks, this makes sense. I tested that this fixes the crash.

stefano-maggiolo:
Merged! If you want to keep the bug open please reopen it.

PR #1105, opened 2019-02-02
Reload the admin/contest page only once once when the contest ends

### LINKED ISSUE #1104: AWS: Infinite refresh loop on contest end when the server is slow to respond
Normally, AWS counts down to the contest end and refreshes the page. Under certain conditions, when the server is slow to respond, this can cause an infinite refresh loop and spam the server with requests, further increasing the load and DoSing the server.

Steps to reproduce:
- Apply the following patch to the code:
  ```diff
  diff --git a/cms/server/admin/handlers/contestranking.py b/cms/server/admin/handlers/contestranking.py
  index b3ff30b6..e85132da 100644
  --- a/cms/server/admin/handlers/contestranking.py
  +++ b/cms/server/admin/handlers/contestranking.py
  @@ -28,6 +28,7 @@
   
   import csv
   import io
  +import time
   
   from sqlalchemy.orm import joinedload
   
  @@ -42,6 +43,7 @@ class RankingHandler(BaseHandler):
       """
       @require_permission(BaseHandler.AUTHENTICATED)
       def get(self, contest_id, format="online"):
  +        time.sleep(5)
           # This validates the contest id.
           self.safe_get_item(Contest, contest_id)
   
  ```
- Set an end time for the contest
- Open the contest's ranking page in AWS and wait
- When the contest ends, refresh loop occurs

It is possible that CWS has the same problem.

andreyv:
Previously the callback to update remaining time in both CWS and AWS would continue to fire every second even after the page reload was triggered. If the server was slow to respond, the reload would be issued again, resulting in an infinite loop and possibly DoS-ing the server.

Fix this by cancelling the update timer when a page reload is issued.

Fixes #1104. See the issue for additional discussion.

The return statement is removed so that the countdown stops at 00:00 and not at 00:01.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1105)

### DIFF
--- a/cms/server/admin/static/aws_utils.js
+++ b/cms/server/admin/static/aws_utils.js
@@ -445,8 +445,10 @@ CMS.AWSUtils.prototype.two_digits = function(n) {
 
 /**
  * Update the remaining time showed in the "remaining" div.
+ *
+ * timer (int): handle for the timer that called this function, or -1 if none
  */
-CMS.AWSUtils.prototype.update_remaining_time = function() {
+CMS.AWSUtils.prototype.update_remaining_time = function(timer = -1) {
     // We assume this.phase always is the correct phase (since this
     // method also refreshes the page when the phase changes).
     var relevant_timestamp = null;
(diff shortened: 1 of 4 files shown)


stefano-maggiolo:
Merged, thank you for fixing the bug

PR #1106, opened 2019-02-09
export_submissions: Introduce a new filename variable 'score'

wafrelka:
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1106)


stefano-maggiolo:
Merged!

PR #1108, opened 2019-02-09
Fix bug that misidentified submissions in CWS autorefresh feature

### LINKED ISSUE #1107: Unofficial submission updates the last "Official submissions" row as well
Steps to reproduce:

- In CWS, go to the Submissions page for a task
- Send some submissions while the contest is running
- Wait until analysis mode starts
- Send another submission
- Observe that the last item in "Official submissions" is updated with the new details too

stefano-maggiolo:
That feature relies on the submissions being named with their position
in the chronological list of submissions of the user. After the split
between official and unofficial lists, the positions were all wrong
since they were computed independently.

Fixes #1107.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1108)

### DIFF
--- a/cms/server/contest/templates/submission_table.html
+++ b/cms/server/contest/templates/submission_table.html
@@ -51,16 +51,18 @@
         </tr>
     </thead>
     <tbody>
-    {% if submission_table_submissions|length == 0 %}
+    {% if submissions|length == 0 %}
         <tr>
             <td colspan="{{ num_cols }}" class="no_submissions">{% trans %}no submissions{% endtrans %}</td>
         </tr>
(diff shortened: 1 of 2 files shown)


stefano-maggiolo:
@andreyv do you want to take a look?

stefano-maggiolo:
There's a minor bug when there are no official submissions but some unofficial (will not display "no submissions" message).

If you don't mind, I'm moving all of that code to macro (since the original bug would have been caught if we were using macros...) and fixing it then.

stefano-maggiolo:
Merged.

PR #1110, opened 2019-02-11
Added e-mail field to registration form.

rvisser7:
Made a simple addition to the registration form for an e-mail address.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1110)

### DIFF
--- a/cms/server/contest/handlers/main.py
+++ b/cms/server/contest/handlers/main.py
@@ -86,6 +86,7 @@ class RegistrationHandler(ContestHandler):
             last_name = self.get_argument("last_name")
             username = self.get_argument("username")
             password = self.get_argument("password")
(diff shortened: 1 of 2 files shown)


rvisser7:
> I think you can add `autocomplete="email"` as explained in https://developers.google.com/web/updates/2015/06/checkout-faster-with-autofill (similar to the password field later).

Done.

stefano-maggiolo:
Do you need the email as a required field? I would prefer to keep it optional.

rvisser7:
> Do you need the email as a required field? I would prefer to keep it optional.

Done (made it optional).  I'll keep it as a required field for our local version though.

stefano-maggiolo:
Merged (with a small further change, if email == "" then make it None).

PR #1111, opened 2019-02-12
Fix RWS with score mode MAX_SUBTASK

lucach:
Tasks with a score type that was not group (i.e., without subtasks)
were not handled correctly.

We consider them as having one subtask (as already done elsewhere in
CMS).

**Important note**
I think that the proposed change can work in the event of an unscored submission (which should have 0 as score), but please double-check what is being done inside those lines. Feel free to modify them as needed.
They were disputed since their introduction and have led to a number of subtle bugs in these last months. :joy:

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1111)

### DIFF
--- a/cmsranking/Scoring.py
+++ b/cmsranking/Scoring.py
@@ -4,6 +4,7 @@
 # Copyright © 2011-2013 Luca Wehrstedt <luca.wehrstedt@gmail.com>
 # Copyright © 2018 Stefano Maggiolo <s.maggiolo@gmail.com>
 # Copyright © 2018 William Di Luigi <williamdiluigi@gmail.com>
+# Copyright © 2019 Luca Chiodini <luca@chiodini.org>
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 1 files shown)


wil93:
Mmm I feel like a better solution would be to avoid altogether having tasks with score_mode = 'max_subtask' and no subtasks.

A DB constraint maybe?

lucach:
That is another option, it depends on how you watch the issue.

I chose this path because it is how CMS treats this situation elsewhere in the codebase (see https://github.com/cms-dev/cms/blob/master/cms/grading/scoring.py#L247-L249 ).

stefano-maggiolo:
Yeah, I see. Seems to make sense, but I'll take another look for the corner case you pointed out. But why does this bother you? Are you using this combination of configurations?

stefano-maggiolo:
Yes, I think it makes sense, since in reset_history the score is set to 0 by default (then changes are applied).

lucach:
> But why does this bother you? Are you using this combination of configurations?

Not intentionally (but it happened by mistake during the final round of the Italian Olympiad in Teams :laughing: )

Thanks for merging the fix, at least now CMS behaves properly. For the future some more profound changes, as those suggested by @wil93, can be implemented.

PR #1113, opened 2019-02-18
Regenerate cms.pot file

lerks:
I first reverted the `{%` -> `{%+` change as that doesn't appear to be compatible with Babel's extracter, then ran `./setup.py extract_messages`.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1113)


stefano-maggiolo:
So small? Nice. Since you're at it, let's do it on master as well?

stefano-maggiolo:
LGTM for this one.

PR #1114, opened 2019-02-18
Regenerate cms.pot file

lerks:
I first reverted the `{%` -> `{%+` change as that doesn't appear to be compatible with Babel's extracter, then ran `./setup.py extract_messages`.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1114)


lerks:
Rebased and updated this.

stefano-maggiolo:
LGTM

PR #1115, opened 2019-02-18
Follow up to the fix to #1107

stefano-maggiolo:
This transforms some include in macros, which is supposedly safer (as
they don't rely on state of the context, which is error prone). Also
fixes a small bug left from the previous fix.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1115)


stefano-maggiolo:
Sorry for the push -f. I think I addressed the failure AND all of the things you pointed out. If you want to take a look let me know, otherwise I'll push to master soonish (e.g., tomorrow).

PR #1117, opened 2019-02-24
Fix ProxyService team.code encoding for user

PPakalns:
Fix inconsistent team.code encoding in ProxyService.

It breaks if `team.code` has `_`  or any other non digit, non letter character.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1117)


andreyv:
Thanks, it works 🙂 I applied slight style changes.

PR #1120, opened 2019-04-20
Added macro templates files to babel mapping.

rvisser7:
As some of the template code has been moved to a macro file, everything in server/contest/templates/macro/*.html must also be localised.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1120)


stefano-maggiolo:
Merged, thank you

PR #1128, opened 2019-08-17
Replace most subprocess.call() invocations with check_call()

andreyv:
This ensures proper error handling. In some places manual error reporting is also replaced with Python chained exceptions.

Manual error handling is left in loaders because multiple other places also return `None` instead of raising an exception. It can be refactored too if needed.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1128)

### DIFF
--- a/cms/grading/Sandbox.py
+++ b/cms/grading/Sandbox.py
@@ -1412,11 +1412,11 @@ class IsolateSandbox(SandboxBase):
             [self.box_exec]
             + (["--cg"] if self.cgroup else [])
             + ["--box-id=%d" % self.box_id, "--init"])
-        ret = subprocess.call(init_cmd)
-        if ret != 0:
+        try:
+            subprocess.check_call(init_cmd)
+        except subprocess.CalledProcessError as e:
             raise SandboxInterfaceException(
-                "Failed to initialize sandbox with command: %s "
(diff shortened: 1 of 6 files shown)


andreyv:
One of the newly added checks fails:
```
ERROR [Worker,3 4 Worker::execute_job_group] Worker failed: Command '['isolate', '--cg', '--box-id=40', '--dir=/tmp=/tmp/cms-compile-50t1z18h/home:rw', '--run', '--', '/bin/chmod', '777', '-R', '/tmp']' returned non-zero exit status 1..
```

andreyv:
The output from isolate is something like
```
/bin/chmod: changing permissions of '/tmp': Operation not permitted
/bin/chmod: changing permissions of '/tmp/checker': Operation not permitted
/bin/chmod: changing permissions of '/tmp/user_output.txt': Operation not permitted
/bin/chmod: changing permissions of '/tmp/input.txt': Operation not permitted
/bin/chmod: changing permissions of '/tmp/correct_output.txt': Operation not permitted
Exited with error status 1
```

Looking at the `home` directory, the mentioned files are owned by the CMS user instead of the sandbox user. The `chmod` command runs as the sandbox user.

Looks like this behavior is intended. A comment can be added that the exit code should be ignored.

PR #1131, opened 2019-08-22
Show original IP address of requests in AWS logs

andreyv:
See also the previous discussion: https://gitter.im/cms-dev/cms?at=5bfa905597a8982b9a606c79

For now I implemented a bare-bones approach with the hidden config setting. It can be developed further:
- Either by documenting the new setting in `cms.conf`
- Or by refactoring `num_proxies_used` to include both CWS and AWS

I can rework this PR if needed.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1131)


andreyv:
Let's make the new setting separate. Initializing it from `num_proxies_used` by default may have security implications.

PR #1139, opened 2019-10-13
Add Language.executable_extension property

andreyv:
This property allows to simplify language compilation commands in multiple cases. The compilation commands presented to the user are also easier to understand.

The initial motivation for this change was to be able to use [`python3 -m zipapp`](https://docs.python.org/3/library/zipapp.html) instead of manual archiving for the Python 3 language. Currently this doesn't work, however, because from [PEP 441](https://www.python.org/dev/peps/pep-0441/#a-new-python-zip-application-extension) there must be a `__main__.py` file in the archive, and `zipapp` [enforces](https://github.com/python/cpython/blob/b16e382c446d76ede22780b15c75f43c5f132e25/Lib/zipapp.py#L112-L117) this. But these changes are still useful for multiple other languages.

A couple of notes:
1. The `.pyz` extension is specified in [PEP 441](https://www.python.org/dev/peps/pep-0441/).
1. This code doesn't work if there are multiple source files, but no grader:
   https://github.com/cms-dev/cms/blob/d4c9e926bd52d8022069c417b206b0882ef4d1ba/cms/grading/tasktypes/Batch.py#L256-L257
    I think it could query the task object instead. Perhaps a new issue can be opened.
1. The PHP language plugin shows that it may be useful to have an `InterpretedLanguage` subclass without compilation commands. Currently a language with empty compilation commands fails here:
    https://github.com/cms-dev/cms/blob/d4c9e926bd52d8022069c417b206b0882ef4d1ba/cms/grading/steps/utils.py#L88-L89
    because `stats` is `None`. On the other hand, it is not clear how this can be reconciled with the case of multiple source files.
1. Logically, each language type plugin should implement its own updater function, so that the updater script can query all plugins, including out-of-tree ones. As this is overkill, the updater just has a hardcoded list of languages instead.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1139)

### DIFF
--- a/cms/grading/languages/php.py
+++ b/cms/grading/languages/php.py
@@ -40,11 +40,20 @@ class Php(Language):
         """See Language.source_extensions."""
         return [".php"]
 
+    @property
+    def executable_extension(self):
+        """See Language.executable_extension."""
+        return ".php"
+
     def get_compilation_commands(self,
                                  source_filenames, executable_filename,
                                  for_evaluation=True):
         """See Language.get_compilation_commands."""
-        return [["/bin/cp", source_filenames[0], executable_filename]]
+        if source_filenames[0] != executable_filename:
+            return [["/bin/cp", source_filenames[0], executable_filename]]
+        else:
+            # We need at least one command to collect execution stats.
+            return [["/bin/true"]]
 
     def get_evaluation_commands(
             self, executable_filename, main=None, args=None):
--- a/cms/grading/languages/python2_cpython.py
+++ b/cms/grading/languages/python2_cpython.py
@@ -45,11 +45,15 @@ class Python2CPython(CompiledLanguage):
         """See Language.source_extensions."""
         return [".py"]
 
+    @property
+    def executable_extension(self):
+        """See Language.executable_extension."""
+        return ".zip"
+
     def get_compilation_commands(self,
                                  source_filenames, executable_filename,
                                  for_evaluation=True):
         """See Language.get_compilation_commands."""
-        zip_filename = "%s.zip" % executable_filename
 
         commands = []
         files_to_package = []
@@ -64,10 +68,8 @@ class Python2CPython(CompiledLanguage):
             else:
                 files_to_package.append(pyc_filename)
 
-        # zip does not support writing to a file without extension.
-        commands.append(["/usr/bin/zip", "-r", zip_filename]
+        commands.append(["/usr/bin/zip", executable_filename]
                         + files_to_package)
-        commands.append(["/bin/mv", zip_filename, executable_filename])
 
         return commands
 
--- a/cms/grading/languages/python3_cpython.py
+++ b/cms/grading/languages/python3_cpython.py
@@ -45,11 +45,16 @@ class Python3CPython(CompiledLanguage):
         """See Language.source_extensions."""
         return [".py"]
 
+    @property
+    def executable_extension(self):
+        """See Language.executable.extension."""
+        # Defined in PEP 441 (https://www.python.org/dev/peps/pep-0441/).
+        return ".pyz"
+
     def get_compilation_commands(self,
                                  source_filenames, executable_filename,
                                  for_evaluation=True):
         """See Language.get_compilation_commands."""
-        zip_filename = "%s.zip" % executable_filename
 
         commands = []
         files_to_package = []
@@ -64,10 +69,8 @@ class Python3CPython(CompiledLanguage):
             else:
                 files_to_package.append(pyc_filename)
 
-        # zip does not support writing to a file without extension.
-        commands.append(["/usr/bin/zip", "-r", zip_filename]
+        commands.append(["/usr/bin/zip", executable_filename]
                         + files_to_package)
-        commands.append(["/bin/mv", zip_filename, executable_filename])
 
         return commands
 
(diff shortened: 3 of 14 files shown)


cms/grading/languages/php.py
```diff
@@ -40,11 +40,19 @@ def source_extensions(self):
...
+
     def get_compilation_commands(self,
                                  source_filenames, executable_filename,
                                  for_evaluation=True):
         """See Language.get_compilation_commands."""
-        return [["/bin/cp", source_filenames[0], executable_filename]]
+        if source_filenames[0] != executable_filename:
+            return [["/bin/cp", source_filenames[0], executable_filename]]
+        else:
+            return [["/bin/true"]]
```
stefano-maggiolo, line 55: Can this be an empty list instead?

cms/grading/languages/python3_cpython.py
```diff
@@ -45,30 +45,25 @@ def source_extensions(self):
         """See Language.source_extensions."""
         return [".py"]
 
+    @property
+    def executable_extension(self):
+        """See Language.executable.extension."""
+        return ".pyz"
```
stefano-maggiolo, line 51: Comment with a link to the PEP

cms/grading/languages/python3_cpython.py
```diff
@@ -45,30 +45,25 @@ def source_extensions(self):
...
 
+        commands = [["/usr/bin/python3", "-m", "compileall", "-b", "."]]
+
+        # The file with the entry point must be in first position.
+        basename, _ = os.path.splitext(os.path.basename(source_filenames[0]))
+        pyc_filename = "%s.pyc" % basename
+        commands.append(["/bin/mv", pyc_filename, self.MAIN_FILENAME])
+
+        commands.append(["/usr/bin/zip", executable_filename,
+                         self.MAIN_FILENAME] + source_filenames[1:])
```
stefano-maggiolo, line 66: This zips the .py files, not the .pyc - is this intended?

cms/grading/languages/python2_cpython.py
```diff
@@ -45,30 +45,25 @@ def source_extensions(self):
...
 
+        commands = [["/usr/bin/python2", "-m", "compileall", "."]]
+
+        # The file with the entry point must be in first position.
+        basename, _ = os.path.splitext(os.path.basename(source_filenames[0]))
+        pyc_filename = "%s.pyc" % basename
+        commands.append(["/bin/mv", pyc_filename, self.MAIN_FILENAME])
+
+        commands.append(["/usr/bin/zip", executable_filename,
+                         self.MAIN_FILENAME] + source_filenames[1:])
```
stefano-maggiolo, line 66: This zips the .py files, not the .pyc - is this intended?

cmscontrib/updaters/update_43.py
```diff
@@ -0,0 +1,60 @@
...
+
+"""A class to update a dump created by CMS.
+
+Used by DumpImporter and DumpUpdater.
+
+This updater adds file name extensions to relevant executables.
+
+Note:
+  - File descriptions are not updated.
+  - Out-of-tree languages should be added to LANG_EXT manually.
```
stefano-maggiolo, line 27: Add "- Custom use of USE_JAR = False should be reflected in this updater."

stefano-maggiolo (commented):
What's the current status of TwoSteps? Are we letting it rot away?

cms/grading/languages/php.py
andreyv: There is a problem here: https://github.com/cms-dev/cms/blob/d4c9e926bd52d8022069c417b206b0882ef4d1ba/cms/grading/steps/utils.py#L88-L89

If there are no commands, then `stats` is `None`. I can see two possible ways: initialize `stats` with a "zero" object (or treat `None` as zero in relevant places), or skip compilation altogether if there are no commands. But the latter is likely to introduce special cases in multiple places.

This PR avoids the problem using `/bin/true` for now.

cms/grading/languages/python3_cpython.py
andreyv: Added.

cms/grading/languages/python3_cpython.py
andreyv: Nice catch! Fixed by returning the old implementation.

cms/grading/languages/python2_cpython.py
andreyv: Fixed.

cmscontrib/updaters/update_43.py
andreyv: Added.

andreyv:
> What's the current status of TwoSteps? Are we letting it rot away?

Thanks, I updated it as well (somewhat). Even in master, as I can see, TwoSteps won't work with Python etc., because it doesn't call `get_evaluation_commands()`, but rather just runs `./executable`. But at least now it is not broken more than it is in master.

cms/grading/languages/php.py
stefano-maggiolo: Ok, please add a comment saying why we need it.

stefano-maggiolo:
I had one more comment request then feel free to push to master. Thank you!

cms/grading/languages/php.py
andreyv: Added a comment.

andreyv:
Thanks! I'll push the changes soon.

andreyv:
Merged.

PR #1141, opened 2019-10-26
Fix support for services connected via IPv6

edomora97:
Use the correct address family when connecting to the various RPC services.

The solution is basically the same suggested here:
https://github.com/cms-dev/cms/issues/157#issuecomment-463427954

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1141)

### DIFF
--- a/cms/io/rpc.py
+++ b/cms/io/rpc.py
@@ -500,6 +500,7 @@ class RemoteServiceClient(RemoteServiceBase):
                              self._repr_remote(), host, port, error)
             else:
                 self.initialize(sock, self.remote_service_coord)
+                break
 
     def _run(self):
         """Maintain the connection up, if required.
--- a/cms/io/rpc.py
+++ b/cms/io/rpc.py
@@ -5,6 +5,7 @@
 # Copyright © 2010-2018 Stefano Maggiolo <s.maggiolo@gmail.com>
 # Copyright © 2010-2012 Matteo Boscariol <boscarim@hotmail.com>
 # Copyright © 2013-2017 Luca Wehrstedt <luca.wehrstedt@gmail.com>
+# Copyright © 2019 Edoardo Morassutto <edoardo.morassutto@gmail.com>
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as
@@ -478,8 +479,12 @@ class RemoteServiceClient(RemoteServiceBase):
 
         """
         try:
-            sock = gevent.socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-            sock.connect(self.remote_address)
+            # Find the family (AF_INET or AF_INET6) of the IPv4/6 address.
+            addresses = gevent.socket.getaddrinfo(
+                self.remote_address.ip, self.remote_address.port)
+            family, *rest, sockaddr = addresses[0]
+            sock = gevent.socket.socket(family, socket.SOCK_STREAM)
+            sock.connect(sockaddr)
         except OSError as error:
             logger.debug("Couldn't connect to %s: %s.",
                          self._repr_remote(), error)
--- a/cms/io/rpc.py
+++ b/cms/io/rpc.py
@@ -489,11 +489,11 @@ class RemoteServiceClient(RemoteServiceBase):
             logger.warning("Cannot resolve %s.", self.remote_address)
             raise
 
-        for family, *rest, sockaddr in addresses:
+        for family, type, proto, _canonname, sockaddr in addresses:
             try:
                 host, port, *rest = sockaddr
                 logger.debug("Trying to connect to %s at port %d.", host, port)
-                sock = gevent.socket.socket(family, socket.SOCK_STREAM)
+                sock = gevent.socket.socket(family, type, proto)
                 sock.connect(sockaddr)
             except OSError as error:
                 logger.debug("Couldn't connect to %s at %s port %d: %s.",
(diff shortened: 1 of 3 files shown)


cms/io/rpc.py
```diff
@@ -478,8 +479,12 @@ def _connect(self):
 
         """
         try:
-            sock = gevent.socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-            sock.connect(self.remote_address)
+            # Find the family (AF_INET or AF_INET6) of the IPv4/6 address.
+            addresses = gevent.socket.getaddrinfo(
+                self.remote_address.ip, self.remote_address.port)
+            family, *rest, sockaddr = addresses[0]
```
andreyv, line 485: It is slightly worrying to take only the first returned address. It would be more prudent to try all of them. From Linux `getaddrinfo(3)`:
> Normally, the application should try using the addresses in the order in which they are returned.

Can you please implement a loop that tries all addresses in succession? A related example is available at https://docs.python.org/3/library/socket.html#example.

cms/io/service.py
```diff
@@ -158,9 +159,11 @@ def _connection_handler(self, sock, address):
 
         """
         try:
-            ipaddr, port = address
-            ipaddr = gevent.socket.gethostbyname(ipaddr)
-            address = Address(ipaddr, port)
+            # In case of IPv6 address is (ip, port, flow info, scope id).
+            ipaddr, port = address[:2]
```
andreyv, line 163: In this place, the address is ultimately coming from `accept()`, and should have the numerical form already. So I think we can just create an `Address` directly from that without any address resolution.

andreyv (changes requested):
Thanks! Overall this looks good. Please see comments below.

edomora97:
Sure! I'll take a look again at this in few days, I'm pretty busy right now!

cms/io/rpc.py
```diff
@@ -478,8 +479,12 @@ def _connect(self):
 
         """
         try:
-            sock = gevent.socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-            sock.connect(self.remote_address)
+            # Find the family (AF_INET or AF_INET6) of the IPv4/6 address.
+            addresses = gevent.socket.getaddrinfo(
```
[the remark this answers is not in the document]
edomora97: Done.

cms/io/rpc.py
edomora97: Done, tell me if you like the loop like that.

cms/io/service.py
edomora97: Right, fixed.

edomora97:
Sorry for the _small_ delay, I've rebased on the current master and fixed the comments you left. The first two commits are the exact same, the third fixes the code review (hopefully).

cms/io/rpc.py
```diff
@@ -478,13 +479,27 @@ def _connect(self):
...
+            try:
+                host, port, *rest = sockaddr
+                logger.debug("Trying to connect to %s at port %d.", host, port)
+                sock = gevent.socket.socket(family, socket.SOCK_STREAM)
+                sock.connect(sockaddr)
+            except OSError as error:
+                logger.debug("Couldn't connect to %s at %s port %d: %s.",
+                             self._repr_remote(), host, port, error)
+            else:
+                self.initialize(sock, self.remote_service_coord)
```
andreyv, line 502: Is a `break` missing here?

andreyv (changes requested):
Hi,

Thanks for the update. Now, re-reading the changes, I started wondering: do we need `getaddrinfo()` at all? [`Address`](https://github.com/cms-dev/cms/blob/da7f2d26b821f4f90f24201adc920ab1ad4ced61/cms/conf.py#L37) already contains an IP address, so do we need to resolve it again?

It looks like it should be enough to just get `ipaddress.ip_address(address[0]).version` and select address family based on that. What do you think?

cms/io/rpc.py
edomora97: Yes. Fixing it immediately.

edomora97:
I don't think `ipaddress.ip_address(address[0]).version` is enough. That `Address` can also contain domain names (as you write them in `cms.conf`), and `ipaddress.ip_address` only parses plain IP addresses.

cms/io/rpc.py
```diff
@@ -478,13 +479,28 @@ def _connect(self):
...
+            # addresses, we'll try all of them.
+            addresses = gevent.socket.getaddrinfo(
+                self.remote_address.ip,
+                self.remote_address.port,
+                type=socket.SOCK_STREAM)
+        except socket.gaierror:
+            logger.warning("Cannot resolve %s.", self.remote_address)
+            raise
+
+        for family, *rest, sockaddr in addresses:
```
andreyv, line 492: Sorry, one more comment: you can get `type` and `proto` here and pass them to `socket()`, as recommended in the docs.

andreyv:
Thanks, the code looks good now and works fine with both IPv4 and IPv6.

andreyv:
I rebased this branch onto de5a0caf9b9ae02e4a90618748cbb46a66189b11.

One unit test fails, mock description needs to be adjusted:
```
AssertionError: expected call not found.
Expected: getaddrinfo('127.0.0.1', 36887)
Actual: getaddrinfo('127.0.0.1', 36887, type=<SocketKind.SOCK_STREAM: 1>)
```

andreyv (approved):

andreyv:
Looks good, thanks.

PR #1142, opened 2019-11-11
Ignore wrong contest submissions in ProxyService

edomora97:
When you are running cms with more than a contest (ie multicontest) ProxyService can be notified by ScoringService of new submissions of any contest, ignoring the `-c X` flag of PS. This will lead to `Inconsistent Data` errors in RWS due to submissions of the wrong contest (bad user, bad task, ...). This is especially bad because the wrong submission can be sent with other valid submissions, and RWS drop them all.

The proposed fix just ignores those submissions from ProxyService since ScoringService has no way to know which contest PS is bound to.

Related to https://github.com/cms-dev/cms/pull/1102 but less intrusive.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1142)

### DIFF
--- a/cms/service/ProxyService.py
+++ b/cms/service/ProxyService.py
@@ -480,9 +480,10 @@ class ProxyService(TriggeredService):
             # ScoringService sent us a submission of another contest, they
             # do not know about our contest_id in multicontest setup.
             if submission.task.contest_id != self.contest_id:
-                logger.debug("Ignoring submission %s of contest %d "
-                             "(this ProxyService considers contest %d only)",
-                             submission.id, submission.task.contest_id, self.contest_id)
+                logger.debug("Ignoring submission %d of contest %d "
+                             "(this ProxyService considers contest %d only).",
+                             submission.id, submission.task.contest_id,
+                             self.contest_id)
                 return
 
             if submission.participation.hidden:
@@ -523,9 +524,10 @@ class ProxyService(TriggeredService):
             # ScoringService sent us a submission of another contest, they
             # do not know about our contest_id in multicontest setup.
             if submission.task.contest_id != self.contest_id:
-                logger.debug("Ignoring submission %s of contest %d "
-                             "(this ProxyService considers contest %d only)",
-                             submission.id, submission.task.contest_id, self.contest_id)
+                logger.debug("Ignoring submission %d of contest %d "
+                             "(this ProxyService considers contest %d only).",
+                             submission.id, submission.task.contest_id,
+                             self.contest_id)
                 return
 
             if submission.participation.hidden:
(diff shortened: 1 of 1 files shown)


andreyv (changes requested):
Thanks, conceptually this looks good.

I think it is better placed in `submission_scored()`/`submission_tokened()`:
1. `_missing_operations()` filters by contest id anyway
2. You can put the code together with the other checks and log a message
3. `dataset_updated()` can have its own meaningful message

Could you please update the PR accordingly?

edomora97:
Sure! I'll take a look again at this in few days, I'm pretty busy right now!

edomora97:
Sorry for the small delay, I've rebased on the current master and fixed the comments you left.

Not sure why the CI doesn't work, but it seems quite unrelated.

cms/service/ProxyService.py
```diff
@@ -488,6 +489,11 @@ def submission_scored(self, submission_id):
                             submission_id)
                 return
 
+            # ScoringService sent us a submission of another contest, they
```
andreyv, line 492: It is better to put this before the `hidden`/`official` checks, otherwise they will get logged for other-contest submissions as well. Same for `submission_tokened()` below.

(Following the order: from most general to least general.)

cms/service/ProxyService.py
```diff
@@ -488,6 +489,11 @@ def submission_scored(self, submission_id):
...
+            # do not know about our contest_id in multicontest setup.
+            if submission.task.contest_id != self.contest_id:
+                return
```
andreyv, line 495: Would you like to put a `info`/`debug` message here? This could help with troubleshooting. Same for `submission_tokened()` below.

cms/service/ProxyService.py
```diff
@@ -9,6 +9,7 @@
 # Copyright © 2015 Luca Versari <veluca93@gmail.com>
 # Copyright © 2015 William Di Luigi <williamdiluigi@gmail.com>
 # Copyright © 2016 Amir Keivan Mohtashami <akmohtashami97@gmail.com>
+# Copyright © 2019 Edoardo Morassutto <edoardo.morassutto@gmail.com>
 #
```
andreyv, line 13: `dataset_updated()` should also check for the task's contest id in a similar way.

andreyv (changes requested):
Thanks for the update. Please see some comments below.

andreyv:
Thanks, now the changes look good.

Please adjust the code to 79 characters and add periods at the end of log messages. Also, you can print submission id with `%d`.

edomora97:
Done!

Btw, is there an automated tool for formatting the code?

andreyv:
Thanks, merging.

Well, not currently, but one could be used in the future.

andreyv (approved):

PR #1144, opened 2019-12-17
Add feedback_level to italy_yaml loader

lucach:
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1144)


andreyv:
Merged, thanks.

I also added an explicit check for `FEEDBACK_LEVEL_RESTRICTED`, as is similarly done in other parts of the code.

Issue #1099, opened 2019-01-19
Contest name with special characters causes HTTP 500 in CWS


andreyv:
There are two variants of the issue. How to reproduce:

1. Run CWS in multi-contest mode
2. Clear cookies
3. Navigate to given path

### Variant 1

Relative path: `/test"`
<details><summary>Result</summary>

```
jan 19 11:07:11 bionic cmsLogService[21243]: 2019-01-19 11:07:11,143 - INFO [Contest,0] Unsuccessful cookie authentication: no cookie provided
jan 19 11:07:11 bionic cmsLogService[21243]: 2019-01-19 11:07:11,145 - ERROR [Contest,0] Uncaught exception GET /test%22 (10.0.2.2)
jan 19 11:07:11 bionic cmsLogService[21243]: HTTPServerRequest(protocol='http', host='localhost:8080', method='GET', uri='/test%22', version='HTTP/1.1', remote_ip='10.0.2.2', headers={'Host': 'localhost:8080', 'X-Real-Ip': '10.0.2.2', 'X-Forwarded-For': '10.0.2.2', 'X-Forwarded-Proto': 'http', 'Connection': 'close', 'User-Agent': 'Mozilla/5.0 (X11; Linux x86_64; rv:64.0) Gecko/20100101 Firefox/64.0', 'Accept': 'text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8', 'Accept-Language': 'ru,en;q=0.7,lv;q=0.3', 'Accept-Encoding': 'gzip, deflate', 'Dnt': '1', 'Upgrade-Insecure-Requests': '1'})
jan 19 11:07:11 bionic cmsLogService[21243]: Traceback (most recent call last):
jan 19 11:07:11 bionic cmsLogService[21243]:   File "/usr/lib/python3/dist-packages/tornado/web.py", line 1489, in _execute
jan 19 11:07:11 bionic cmsLogService[21243]:     result = self.prepare()
jan 19 11:07:11 bionic cmsLogService[21243]:   File "/usr/local/lib/python3.6/dist-packages/cms-1.5.dev0-py3.6.egg/cms/server/contest/handlers/contest.py", line 67, in prepare
jan 19 11:07:11 bionic cmsLogService[21243]:     self.choose_contest()
[...]

PR #1146, opened 2020-03-03
Replacing spaces in cookies with _

eliasffyksen:
This solves, Issue #1099, contests with spaces gives server error 500.
This is because the the cookie can not contain spaces.
This is simply fixed by replacing all spaces with underscores in the name.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1146)

### DIFF
--- a/cms/server/contest/handlers/contest.py
+++ b/cms/server/contest/handlers/contest.py
@@ -138,7 +138,7 @@ class ContestHandler(BaseHandler):
             user logged in for the running contest.
 
         """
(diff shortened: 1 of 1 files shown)


andreyv:
Thanks for the PR.

Contest names cannot have spaces; validation for this was fixed in 9da2d146d8306464e49dc2bccc982f3fca9bfb8c. After this commit it shouldn't be possible to get an invalid cookie name in the first place.

Therefore I'm closing this pull request.

PR #1147, opened 2020-03-22
Fix cmsImportContest when multiple IPs are provided

wil93:
The cmsImportContest command fails when some participations specify multiple IPs separated by comma

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1147)


andreyv:
Thanks, looks good and works fine.

In the `italy_yaml` loader, the list syntax could be supported as well in the future:
```yaml
    ip:
      - "192.168.0.0/24"
      - "10.0.0.0/8"
```

PR #1149, opened 2020-03-27
Make clients revalidate files

magula:
We've noticed that when an admin changes a task statement during a contest (e.g. through the web interface), a contestant who has already downloaded the respective statement will regularly still get served the old version by his browser. The same probably holds for attachments as well.

This is because `FileServerMiddleware` sets the response headers such that a client should assume the file sent to be persistent, i.e. it doesn't change for a year and the client doesn't have to ask whether it's changed until then.

With this change, the client is told to always ask whether the file was changed (revalidate) instead of just using the cached file (by setting the `no-cache` flag and `max-age=0`, [reference](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#Requiring_revalidation)). This affects all files that are served through `FileServerMiddleware`.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1149)

### DIFF
--- a/cms/server/file_middleware.py
+++ b/cms/server/file_middleware.py
@@ -2,6 +2,7 @@
 
 # Contest Management System - http://cms-dev.github.io/
 # Copyright © 2018 Luca Wehrstedt <luca.wehrstedt@gmail.com>
+# Copyright © 2021 Manuel Gundlach <manuel.gundlach@gmail.com>
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as
@@ -21,9 +22,6 @@ from werkzeug.wrappers import Response, Request
 from werkzeug.wsgi import responder, wrap_file
 
 from cms.db.filecacher import FileCacher, TombstoneError
-
-
-SECONDS_IN_A_YEAR = 365 * 24 * 60 * 60
 
 
 class FileServerMiddleware:
@@ -105,7 +103,7 @@ class FileServerMiddleware:
             response.headers.add(
                 "Content-Disposition", "attachment", filename=filename)
         response.set_etag(digest)
-        response.cache_control.max_age = SECONDS_IN_A_YEAR
+        response.cache_control.no_cache = True
         response.cache_control.private = True
         response.response = \
             wrap_file(environ, fobj, buffer_size=FileCacher.CHUNK_SIZE)


wil93:
Related PR: #1138

But I like your solution a lot more! 😄 👍 

Do you know if it can be also adapted to specify `no-store` and `must-revalidate`? (I'm actually not entirely sure that they are necessary, I think I tested them at some point and I kind of remember them being necessary...)

andreyv:
Note that if the content hasn't changed, the client shouldn't re-download it. So I think `no-store` isn't needed, instead it should be better to set HTTP ETag.

magula:
I completely missed PR #1138, sorry about that.

I agree with @andreyv that we shouldn't use `no-store` as that would completely turn off caching.

It seems that an `ETag` is indeed set and used to revalidate the file.

As I understand it, `must-revalidate` would tell a browser to absolutely not use a file unless it has revalidated it, i.e. confirmed that it's still up-to-date. I think that would only make a difference when the internet connection is lost, in which case it would cause clicking the download button not to work, whereas without `must-revalidate` the browser might just yield the old file again, even though it knows the file could be stale and that it _should_ always revalidate (when it can). I'm not sure if that's what actually happens, though. Nor do I know which behavior would be better.

andreyv:
@magula I confirmed that files are revalidated and HTTP `ETag`/[`If-None-Match`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-None-Match) is used with these changes—in Chromium.

Firefox, for some reason, always issues a full request. However, the same behavior happens with master, and it is still compliant, I suppose.

From [RFC 7234](https://tools.ietf.org/html/rfc7234#section-5.2.2),  `no-cache` alone should work, so you can remove `max-age` :)

P.S. Not sure if `private` is compatible with `no-cache` either.

magula:
Indeed, I will remove the `max-age` directive.

I'm inclined to think `private` is fine, though I'm not certain. Judging from [RFC 7234 Section 3](https://tools.ietf.org/html/rfc7234#section-3) (when to store in a cache; affected by `private`) and [Section 4](https://tools.ietf.org/html/rfc7234#section-4) (when to use something from a cache; affected by `no-cache`), I don't see how the two would interfere. The standard does say ([Section 5.2.2.6](https://tools.ietf.org/html/rfc7234#section-5.2.2.6)),

> A private cache MAY store the response and reuse it for later requests, even if the response would normally be non-cacheable.

which does sound contradictory to `no-cache`, but here I believe "non-cacheable" refers to what is decribed for the `public` directive ([Section 5.2.2.5](https://tools.ietf.org/html/rfc7234#section-5.2.2.5)) as

> responses that would normally not be stored, due to their status codes not being defined as cacheable by default; see
   [Section 4.2.2](https://tools.ietf.org/html/rfc7234#section-4.2.2).

which in turns refers to [RFC 7231 Section 6.1](https://tools.ietf.org/html/rfc7231#section-6.1), which defines codes like 200 as "cacheable by default".

wil93:
FYI, I just found that MDN has a nice page explaining all the caching directive "use cases": https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#directives

From what it says, it seems that the only way to effectively request a new version everytime is to specify both `no-store` and `max-age=0`

> The `no-store` directive will prevent a new resource being cached, but it will not prevent the cache from responding with a non-stale resource that was cached as the result of an earlier request. Setting `max-age=0` as well forces the cache to revalidate (clears the cache).

magula:
@wil93 Yes, but I believe the `no-store` directive is not what we want. The browser _should_ cache files, and it should use them for subsequent requests after checking with the server whether they're still actual (haven't changed). This "asking for validation" (via telling the server the timestamp of the cached file) is what the browser is instructed to do by the `no-cache` directive. The `no-store` directive, however, would tell the browser to not cache the file, so it will always have to be transmitted again, even when it hasn't changed. This would seem less efficient, especially considering the files this affects rarely change.

andreyv:
Thanks everyone, merged.

PR #1150, opened 2020-03-27
Don't open title page input file in binary mode in PrintingService

magula:
`PrintingService` currently doesn't work. When someone uses it to print something, the compilation of (or rather the rendering of the template of) the title page will fail. As far as I can tell, the reason for this comes from the transition from Tornado to Jinja2:

Jinja2 templates render to strings (unlike the previously used Tornado templates, which render to bytes). So we shouldn't open the output file in binary mode.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1150)


andreyv:
Thanks, the description and the fix are good.

PR #1151, opened 2020-03-27
Let PdfFileMerger open the files to be merged itself

magula:
`PrintingService` currently doesn't work. With the fix mentioned in PR #1150, a file is printed, but it only contains empty pages.

It seems that `PdfFileMerger` is used in a way that raises unspecified behavior: The input files are opened and appended to the `pdfmerger`, and then closed again before `pdfmerger.write(...)` is called. This way, `pdfmerger.write` probably tries to read the appended files, but is unable to. I suppose that in previous versions of `PdfFileMerger` the appended files were already read into memory when `append` was called, and that the implementation has changed.

To fix this, we could keep the files open until calling `pdfmerger.write(...)`. But as `pdfmerger` is able to open files itself, the proposed change just gives it the paths to the files and calls `pdfmerger.close()` in the end, so `pdfmerger` should close all files again.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1151)

### DIFF
--- a/cms/service/PrintingService.py
+++ b/cms/service/PrintingService.py
@@ -190,13 +190,11 @@ class PrintingExecutor(Executor):
                     "(error %d)" % (pretty_print_cmdline(cmd), ret))
 
             pdfmerger = PdfFileMerger()
-            with open(title_pdf, "rb") as file_:
-                pdfmerger.append(file_)
-            with open(source_pdf, "rb") as file_:
-                pdfmerger.append(file_)
+            pdfmerger.append(title_pdf)
+            pdfmerger.append(source_pdf)
             result = os.path.join(directory, "document.pdf")
-            with open(result, "wb") as file_:
-                pdfmerger.write(file_)
+            pdfmerger.write(result)
+            pdfmerger.close()
 
             try:
                 printer_connection = cups.Connection()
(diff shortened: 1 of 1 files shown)


andreyv:
Thanks, very good analysis, and I can also reproduce the problem.

Looks like this broke after the Python 3 migration. See mstamy2/PyPDF2#293.

Letting `PdfFileMerger` open the files sounds good, but there is a catch.

`PdfFileMerger` [keeps the files open](https://github.com/mstamy2/PyPDF2/blob/master/PyPDF2/merger.py#L181) until the `.close()` method is called. If an exception is raised before that, the method won't be called and the open files will be left dangling. This may have been the reason the original code was written this way.

You could use [`contextlib.closing`](https://docs.python.org/3/library/contextlib.html#contextlib.closing) to manage the `PdfFileMerger` object in a `with` statement.

Note that the PyPDF2 code itself could still leak the file objects, depending on where exactly an exception is raised. But as the current implementation doesn't work at all, I think it's an acceptable tradeoff.

magula:
You're right -- I'll make a new commit implementing the change you proposed.

andreyv:
Tested and merged. Thanks!

PR #1152, opened 2020-03-27
Make PrintingService translatable

magula:
There are strings in `PrintingService` that appear in the contestant interface and are already marked translatable. However, `babel_mappings` doesn't recognize `service/` so far.

This change lets `babel_mappings` include `service/*.py`.

It might be better to only include `service/PrintingService.py` or to refactor `PrintingService` so the strings are somewhere else, in a more appropriate place.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1152)

### DIFF
--- a/babel_mapping.cfg
+++ b/babel_mapping.cfg
@@ -5,6 +5,7 @@
 [python: server/contest/*.py]
 [python: server/contest/handlers/*.py]
 [python: server/contest/submission/*.py]
+[python: service/*.py]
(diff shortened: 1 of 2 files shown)


andreyv:
Looks good, and thanks for updating the `.pot` file :)

I agree it feels slightly odd to have translatable messages in `services`. A better approach may be to have a status enum instead of text in `printjobs.status`. But anyway, the current variant is also good.

PR #1158, opened 2020-06-29
Fix task submission language option re-enable when reset

edisonhello:
Language list will be disable when a file is selected, but will not re-enable when `reset` button is pressed.
This commit add a event listener on the button which calls `CMS.CWSUtils.filter_languages` to reset language list.


andreyv:
Thanks, I can reproduce the issue.

I think there may be an even easier fix. Could you please check if this works:
1. Use the `onreset` event attribute (on `<form>`) instead of `onclick`.
2. Just `.find()` the `<option>` elements directly and call `.removeAttr()` on all of them at once.

PR #1162, opened 2020-09-25
Fix Haskell timeout-cputime test

wil93:
This PR should (hopefully) fix the error in the Travis logs

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1162)

### DIFF
--- a/cmstestsuite/code/timeout-cputime.hs
+++ b/cmstestsuite/code/timeout-cputime.hs
@@ -1,4 +1,7 @@
+import Control.Monad
+
+main :: IO ()
 main = do
     n <- (read :: String -> Int) `fmap` getLine
     putStrLn $ "correct " ++ show n
-    last [0 ..] `seq` return ()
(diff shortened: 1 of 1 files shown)


andreyv:
Looks like the test now fails even with the old Ubuntu version.

wil93:
Yeah 😕 we're looking into it

andreyv:
FWIW, I can't reproduce the failure on a local Ubuntu 18.04 installation. I'll continue later on my side.

I noticed that the program maps 1 TB of virtual memory, which may (or may not) be relevant.

wil93:
It should work now! I rebased the right commit on top of master.

andreyv:
Thanks :) Now it also works with Ubuntu 18.04.

andreyv:
Merged in fe15f9183d4fc6478c506fa18b5da83b2f1bedeb.

PR #1163, opened 2020-11-08
Support Ubuntu 20/Python 3.8

ranaldmiao:
This PR introduces a set of minimal changes required to run CMS on Ubuntu 20.
This combination of modules have been used in IOI 2020.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1163)


ranaldmiao:
@wil93 PR opened as requested :)

wil93 (approved):

docs/Installation.rst
```diff
@@ -16,7 +16,7 @@ These are our requirements (in particular we highlight those that are not usuall
 
 * `GNU compiler collection <https://gcc.gnu.org/>`_ (in particular the C compiler ``gcc``);
 
-* `Python <http://www.python.org/>`_ >= 3.6;
+* `Python <http://www.python.org/>`_ >= 3.8;
```
andreyv, line 19: Python 3.6 should remain supported, so the version should be kept here.

requirements.txt
```diff
@@ -6,13 +6,13 @@ tornado>=4.5,<4.6  # http://www.tornadoweb.org/en/stable/releases.html
 # http://initd.org/psycopg/articles/2018/02/08/psycopg-274-released/
 # http://initd.org/psycopg/docs/install.html#disabling-wheel-packages-for-psycopg-2-7
 # We will be able to remove the flag when upgrading to version 2.8.
-psycopg2>=2.7,<2.8 --no-binary psycopg2  # http://initd.org/psycopg/articles/tag/release/
+psycopg2>=2.8.4,<2.9 --no-binary psycopg2  # http://initd.org/psycopg/articles/tag/release/
```
andreyv, line 9: The comment above says that `--no-binary psycopg2` can be removed starting from version 2.8. Could you please implement this as well and remove the comment?

andreyv (changes requested):

andreyv:
Thanks!

Adding minimal changes sounds good as a first step. Please see a few comments above.

I see that some packages are newer than in Ubuntu 20.04 due to Python 3.8 support or wheel availability. I'll test the system packages on an Ubuntu 20.04 installation later, but so far these versions are fine.

requirements.txt
ranaldmiao: The documentation of `psycopg` seems to suggest that the `no-binary` build is recommended over the `binary` version for production environments.

From https://www.psycopg.org/docs/install.html#quick-install :

> 
> The psycopg2-binary package is meant for beginners to start playing with Python and PostgreSQL without the need to meet the build requirements.
> 
> If you are the maintainer of a published package depending on psycopg2 you shouldn’t use psycopg2-binary as a module dependency. For production use you are advised to use the source distribution.

andreyv: Yes, note that here we have the [`psycopg2`](https://pypi.org/project/psycopg2/) package and not the [`psycopg2-binary`](https://pypi.org/project/psycopg2-binary/) package. `psycopg2` contains the source distribution.

This is described [here](https://www.psycopg.org/docs/install.html#change-in-binary-packages-between-psycopg-2-7-and-2-8), as noted in the comment:
> In version 2.7.x, `pip install psycopg2` would have tried to install automatically the binary package of Psycopg. Because of concurrency problems binary packages have displayed, `psycopg2-binary` has become a separate package, and from 2.8 it has become the only way to install the binary package.

From version 2.8, installing `psycopg2` always builds from source, so the flag becomes redundant.

andreyv:
This is merged in 6a9c965e0f65942f712e763acc437a65e7a42acc, thanks.

requirements.txt
andreyv: The `--no-binary` option is removed in 20a365e84daf39bcfb2134412d18ce47c0ca556b.

Issue #1159, opened 2020-07-20
Field name clash in user custom testing results


ollpu:
**Description:**
![image](https://user-images.githubusercontent.com/7241014/87932160-b7bf6b00-ca93-11ea-9c7b-78804ee533a1.png)
**Time** reads execution time instead of the time the test was sent at.

**Steps to reproduce:**
Send a custom test and wait for it to complete without reloading the page. If the result is fetched via JS, the visual error occurs.

It seems both fields have `class="time"`, so the first field gets overwritten.

[cms/server/contest/templates/user_test_row.html](https://github.com/cms-dev/cms/blob/bfd0f82e7fab8753a5614c485c6de3583521dd2a/cms/server/contest/templates/user_test_row.html)
```html
<td class="time">{{ t.timestamp|format_time }}</td>
...
<td class="time undefined">
...
<td class="time">
```

PR #1164, opened 2020-11-08
Fix issue #1159

ranaldmiao:
This PR fixes issue #1159

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1164)

### DIFF
--- a/cms/server/contest/handlers/taskusertest.py
+++ b/cms/server/contest/handlers/taskusertest.py
@@ -197,10 +197,10 @@ class UserTestStatusHandler(ContestHandler):
                 self._("Executed"), self._("details"))
 
             if ur.execution_time is not None:
-                data["time"] = \
+                data["execution_time"] = \
                     self.translation.format_duration(ur.execution_time)
             else:
-                data["time"] = None
+                data["execution_time"] = None
 
             if ur.execution_memory is not None:
                 data["memory"] = \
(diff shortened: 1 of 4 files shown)


andreyv:
Thanks for the fix :)

From the code, it looks like the `datetime` and `time` classes are intended for styling the first column. See [cws_style.css](https://github.com/cms-dev/cms/blob/18721ff6bc326fd293e102e24aa2b2f76e7ae8b7/cms/server/contest/static/cws_style.css#L907-L955) and similar code in [submission.html](https://github.com/cms-dev/cms/blob/18721ff6bc326fd293e102e24aa2b2f76e7ae8b7/cms/server/contest/templates/macro/submission.html).

In particular, there seems to be another bug [here](https://github.com/cms-dev/cms/blob/18721ff6bc326fd293e102e24aa2b2f76e7ae8b7/cms/server/contest/static/cws_style.css#L907-L918):  the `col.time` selector appears twice and overrides its previous definition.

What do you think about keeping `datetime`/`time` for the first column and re-classing the third column, in both `colgroup` and `tbody`, to something like `execution_time`? This would fix both problems and keep consistency with the submissions table.

ranaldmiao:
> What do you think about keeping `datetime`/`time` for the first column and re-classing the third column, in both `colgroup` and `tbody`, to something like `execution_time`? This would fix both problems and keep consistency with the submissions table.

It's been a while, I have made the changes as you suggested and rebased it on the current master branch!

Thanks for the review too!

andreyv:
Thanks! The fix looks good now. I adjusted the key in JSON response from `time` to `execution_time` as well.

PR #1165, opened 2020-11-08
Add no-users option to DumpImporter and DumpExporter

ranaldmiao:
At the moment, when exporting a contest using `cmsDumpExporter`, there is no option to omit information related to users during the export.
Thus, any user which is related to the contest (participants, admin, etc..) will also be included in the dump file. When importing this dump file into another CMS instance, `cmsDumpImporter` will attempt to create these users. If existing users with the same username exists, this will result in database `IntegrityError`, causing the import to fail.

This PR introduces the option to export a contest without any user information via `cmsDumpExporter`. (i.e. Just the tasks itself).

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1165)

### DIFF
--- a/cms/db/util.py
+++ b/cms/db/util.py
@@ -274,8 +274,8 @@ def get_datasets_to_judge(task):
 
 def enumerate_files(
         session, contest=None,
-        skip_submissions=False, skip_user_tests=False, skip_print_jobs=False,
-        skip_generated=False):
+        skip_submissions=False, skip_user_tests=False, skip_users=False,
+        skip_print_jobs=False, skip_generated=False):
     """Enumerate all the files (by digest) referenced by the
     contest.
 
@@ -302,7 +302,7 @@ def enumerate_files(
     queries.append(dataset_q.join(Dataset.testcases)
                    .with_entities(Testcase.output))
 
-    if not skip_submissions:
+    if not skip_submissions and not skip_users:
         submission_q = task_q.join(Task.submissions)
         queries.append(submission_q.join(Submission.files)
                        .with_entities(File.digest))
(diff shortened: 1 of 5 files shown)


andreyv:
Thanks.

Observe that this feature is conceptually very similar to a contest loader: there is some task data and one want to import it. Just for clarification, do you use it for a similar purpose, e.g., prepare a contest on a staging server and import it on a testing/production server?

ranaldmiao:
Yes, this is for the purpose you mentioned. The 'staging' server as well as the testing/production server are both CMS.

At the moment, this is done by exporting the contest (using `cmsDumpExporter`) from the staging CMS instance and then re-importing them into the production instance using `cmsDumpImporter`.

Would you recommend looking into implementing a `ContestLoader` that reads from a cms dump exporter format?

ranaldmiao:
@andreyv Requested changes have been made, do review when you have the time. 

Thanks once again :)

andreyv:
Thanks, it looks good now.

I see that you added tests and also fixed a bug that could cause `DumpImporter` to import everything instead of skipping items as requested — nice!

I added some minor changes in f41a5d872198400dbd58ab0bb9ba2d554adde835:
* In `DumpExporter`, simplify the user-related class list and only check classes that can be reached from root (from `Contest` and `Task` classes). The other previously listed classes, including `User`, should not appear there.
* In `DumpImporter`, also skip `Announcement`. Otherwise it could indirectly import `Admin`s as well when importing a full dump with `-X`.
* Whitespace and indentation fixes.

I'll merge this MR in a few days if you have no further comments.

andreyv:
Merged!

Issue #1153, opened 2020-05-01
User testing does not pick up header files


CristianCantoro:
User testing does not pick up header files from the task, resulting in failing compilation. User submission functions normally (submitting only the source code).

**Steps to reproduce:**
1. Create a task with a grader and a header file
2. Go to user testing, upload the source code and the grader code, and an input file.

**Expected:** the test code is compiled with the grader and executed

**Actual:** Compilation files
```
grader.cpp:11:17: fatal error: taskname.h: No such file or directory
compilation terminated.
taskname.cpp:2:17: fatal error: taskname.h: No such file or directory
 #include "taskname.h"
                 ^
compilation terminated.
```

### Logs

N/A

### System Information

**CMS version:** 1.4rc1
**Was CMS installed:** yes
**Using a virtualenv:** no
**Postgres version:** 9.5
**Distribution and version:** Ubuntu 16.04.6 LTS

CristianCantoro:
One thing that I have realized is that in previous versions of CMS when testing tasks with a grader, it only asked for the source code and not for the grader source code.

## Old user testing interface
![CMS 1.3 user testing interface](https://i.imgur.com/a3OMUgH.png)

## Current (`cms1.4rc1`) user testing interface
![CMS 1.4rc1 user testing interface](https://i.imgur.com/wUFyNb1.png)

I am not really sure why his change was made since the grader is not something that contestants can change. We do provide the code of the grader for local testing, but in principle, this should not be necessary.

CristianCantoro:
It looks like that the breaking commit is https://github.com/cms-dev/cms/commit/a64cc71b84c0e18195b454db8a65d23befd0e9be, but I do not know if it is safe to revert this change, What do you think @stefano-maggiolo?

andreyv:
@CristianCantoro This is actually the expected behavior. See https://cms.readthedocs.io/en/latest/Task%20types.html#batch:
> Batch supports user tests; if a grader is used, the contestants must provide their own grader (a common practice is to provide a simple grader to contestants, that can be used for local testing and for server-side user tests).

As you noted, admins need to provide a grader to contestants for local testing, and so it can be submitted together with the solution.

andreyv:
The issue with headers should be fixed now.

CristianCantoro:
@andreyv said:
> The issue with headers should be fixed now.

thanks, @andreyv, will this fix be in the next release?

andreyv:
[...]

PR #1172, opened 2021-03-07
Fix misc issues regarding user tests with graders

ranaldmiao:
- This PR primarily fixes the issue described in #1153 where header files cannot be found for user tests

This implementation re-uses the same header files as that for actual submissions. (Assumption: Task authors typically do not have different implementations for public and private header files.)

- This PR also allows for graders for user tests to be downloaded. Previously, a HTTP 404 will be encountered when contestants select the option to download grader files from their user tests.

It seems that graders are not stored with the `.%l` suffix (Eg: `grader.%l`) but the original extension (Eg: `grader.cpp`) instead. Hence, this PR introduces an additional check for the original filename before declaring HTTP 404.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1172)

### DIFF
--- a/cms/grading/Job.py
+++ b/cms/grading/Job.py
@@ -378,6 +378,15 @@ class CompilationJob(Job):
                     managers[manager_filename] = \
                         dataset.managers[manager_filename]
 
+        # Copy header files from dataset.
+        # FIXME This bypasses get_auto_managers() logic
+        if language is not None:
+            for manager_filename in dataset.managers:
+                if any(manager_filename.endswith(header)
+                       for header in language.header_extensions):
+                    managers[manager_filename] = \
+                        dataset.managers[manager_filename]
+
         return CompilationJob(
             operation=operation,
             task_type=dataset.task_type,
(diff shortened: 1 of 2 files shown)


andreyv (changes requested):
Thanks for the fixes!

The first change is related to `TaskType.get_auto_managers()`, which is supposed to return managers to be taken from the task. But you can see that currently there is no way for `get_auto_managers()` to account for a non-mandatory, arbitrarily named `.h` manager (even less so for additional `.o` managers). One could make it work now by returning `None` instead of `[]` from `Batch.get_auto_managers()`, but I don't think this is intended.

So I think that for now the proposed change is a sufficient short-term fix until a general solution is worked out.

The second change looks good. Managers are in general stored without `.%l` in the database, so it is fitting to do the same here.

ranaldmiao:
Thanks for the review, @andreyv 

Made the requested changes.

andreyv (approved):

andreyv:
Thanks, merged. Also added tests in 81fb9df98e90e710c87c9e5c44458577fc69f2a8.

PR #1174, opened 2021-03-26
While starting to look at typing, I found some code we took from CPython 2.7 and was not updated. Briefly tested manually, it looks like it's working (but please check, this is my first time touching CMS in quite some time).

stefano-maggiolo:
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1174)

### DIFF
--- a/cms/log.py
+++ b/cms/log.py
@@ -2,7 +2,7 @@
 
 # Contest Management System - http://cms-dev.github.io/
 # Copyright © 2010-2013 Giovanni Mascellani <mascellani@poisson.phc.unipi.it>
-# Copyright © 2010-2016 Stefano Maggiolo <s.maggiolo@gmail.com>
+# Copyright © 2010-2021 Stefano Maggiolo <s.maggiolo@gmail.com>
 # Copyright © 2010-2012 Matteo Boscariol <boscarim@hotmail.com>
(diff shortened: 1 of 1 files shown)


cms/log.py
```diff
@@ -40,8 +40,8 @@
 # OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 #
 # You can find the original files at:
-# http://hg.python.org/cpython/file/69ee9b554eca/Lib/logging/__init__.py
-# http://hg.python.org/cpython/file/69ee9b554eca/Lib/logging/handlers.py
+# https://hg.python.org/cpython/file/f23fa1f7b68f/Lib/logging/__init__.py
+# https://hg.python.org/cpython/file/4243df51fe43/Lib/logging/handlers.py
```
andreyv, line 44: Minor note: The new links are for two different revisions. Is it intended?

andreyv (commented):
As far as I can see, the upstream changes are ported correctly, and the custom code is preserved. Looks good!

cms/log.py
stefano-maggiolo: Maybe - I think I clicked on one I clicked on the "last commit changing this file" to get a permalink instead of the "last commit", I'll change it.

stefano-maggiolo:
Thank you Andrey!

PR #1175, opened 2021-04-10
Make no-conf apply to isolate config as well

ranaldmiao:
Currently, when running `sudo python3 prerequisites.py install --no-conf`, configuration files for `isolate` will be re-installed.

This PR introduces a change for `--no-conf` to also be applied to `isolate` configuration files.
When `--no-conf` argument is provided, `isolate` configuration files will not be installed.

### DIFF
--- a/prerequisites.py
+++ b/prerequisites.py
@@ -224,6 +224,19 @@ def install_isolate():
              os.path.join(USR_ROOT, "bin", "isolate"),
              root, 0o4750, group=cmsuser_grp)
 
+def install_isolate_conf():
+    """This function installs the configuration files for isolate
+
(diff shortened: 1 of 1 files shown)


andreyv:
Thanks.

The general idea is good, but I think it breaks the
```shell
./prerequisites.py install_isolate
```
subcommand. This should be easy to fix, though.

ranaldmiao:
> Thanks.
> 
> The general idea is good, but I think it breaks the
> 
> ```shell
> ./prerequisites.py install_isolate
> ```
> 
> subcommand. This should be easy to fix, though.

Good point, let me fix that and update you

PR #1176, opened 2021-04-14
Migrate CI to GitHub Actions

fushar:
Migrating CI to GitHub's native CI tools (GitHub actions) so that we don't need to rely on external site (Travis) anymore.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1176)


fushar:
Sample CI job results: https://github.com/ioi-2022/cms/actions/runs/747025045

cmstestsuite/unit_tests/io/rpc_test.py
```diff
@@ -262,7 +262,7 @@ def test_autoreconnect1(self):
         self.sleep()
         self.assertTrue(client.connected)
         self.disconnect_servers()
-        gevent.sleep(0.01)
+        gevent.sleep(0.1)
```
fushar, line 265: Cannot get CI to pass without this change :( Please let me know if this is OK

.github/workflows/main.yml
```diff
@@ -0,0 +1,100 @@
...
+          psql --host=localhost --username=postgres --dbname=cmsdbfortesting --command='GRANT SELECT ON pg_largeobject TO cmsuser'
+
+          cp ./config/cms.conf.sample ./config/cms.conf
+          sed -e "s/\"cmsuser\": \"cmsuser\"/\"cmsuser\": \"$USER\"/" -i ./config/cms.conf
+
+          cmsInitDB
+
+      - name: Create cache and log directories for tests
+        run: | 
+          mkdir cache log
```
fushar, line 96: Need to create this in advance, otherwise the test will fail with permission error on this line:

https://github.com/cms-dev/cms/blob/master/cms/util.py#L60

I'm suspecting this is because the default group in GitHub Actions is `docker`. Do not have bandwidth to investigate further. Please advise if there is better approach.

.github/workflows/main.yml
```diff
@@ -0,0 +1,100 @@
...
+        run: |
+          pip install -r requirements.txt
+          pip install -r dev-requirements.txt
+
+          sudo $(which python) prerequisites.py -y --cmsuser=$USER install
+          python setup.py install
+
+      - name: Change group for working and temp directories to $USER
+        run: |
+          sudo chgrp $USER . /tmp
```
fushar, line 67: By default, the user group is `docker` for GitHub actions runner. Without this, somehow CMS cannot create temp files in tests (permission denied)

andreyv:
Hi,

Thanks for the PR, the idea sounds great. I'll take a look at the details a little bit later.

cmstestsuite/unit_tests/io/rpc_test.py
andreyv: I think this is fine :) The objective is to verify that the client eventually reconnects, and the `sleep()` is there to allow the client to run. See also daf332ff935a7b629dfdf063f7f720a5629202a1.

andreyv (commented):
Please see some comments below.

I noticed the following in the new build log, too:
* Multiple warnings such as
  ```
  Using legacy 'setup.py install' for netifaces, since package 'wheel' is not installed.
  ```
* Python packages, and CMS itself (!), are installed to, e.g., `/opt/hostedtoolcache/Python/3.8.8/x64/lib/python3.8/site-packages`, which doesn't really look like a user directory. Not sure if this is the intended way.

I will also be running some experiments with Github Actions to better understand the problems with the user group and `isolate`, and the warnings as well.

.github/workflows/main.yml
```diff
@@ -0,0 +1,100 @@
...
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        python-version:
+          - 3.8
+          - 3.6
+    
+    services:
+      postgres:
+        image: postgres:13
```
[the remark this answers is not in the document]
fushar: Yeah, using PG 12 also works: https://github.com/ioi-2022/cms/actions/runs/807190223

fushar:
Will take a look at other comments later this week!

.github/workflows/main.yml
```diff
@@ -0,0 +1,100 @@
...
+          submodules: true
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install dependencies
+        run: >-
+          sudo apt-get install -y
```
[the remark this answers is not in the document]
fushar: Done!

fushar:
Ah, thanks for investigating and fixing the issue!

I've cherry-picked your fix and included the commit in this branch. The CI passed here: https://github.com/ioi-2022/cms/actions/runs/810455275. Now this PR should be good to go 😄

andreyv:
Merged, thanks :)

PR #1178, opened 2021-05-13
Fix typo in the AddSubmission script

Grackins:
Edit help message of the `--contest-id` argument

The name of the script is "Add Submission" but there is "id of contest where to add the user" in the help message. So I thought that would've been a mistake.

---
This change is [Reviewable](https://reviewable.io/reviews/cms-dev/cms/1178)


andreyv:
Thanks, indeed this looks like a mistake. Merging.

PR #1180, opened 2021-05-15
fix #1179

### LINKED ISSUE #1179: Can't download source from tests with grader because the link to the file has the wrong extension
## Issue template

**Description:** 

Downloading the source code from tests with grader fails with Error 500 because they have the wrong extension, `task.%25l` instead of `task.cpp`. Note that changing manually the extension to the correct one (`%25l` -> `.cpp`) the file is downloaded.

**Steps to reproduce:**

1. apply the patch as in #1153.
2. upload a test for a task with a grader, `ìnput.txt`, `task.cpp`, `grader.cpp`.
3. try to download the source of the task.

Laboratoriodialgoritmi—MozillaFirefox_2021-05-15_00:38:51

**Expected:**

The source code of the test (`task.cpp`) is downloaded.

**Actual:**

The server responds with a 500 error because the link is incorrect. In the example:
```
http://<domain.tld>/tasks/tsp/tests/1/files/tsp.%25l
```

Error500—MozillaFirefox_2021-05-15_00:39:00

instead of:
```
http://<domain.tld>/tasks/tsp/tests/1/files/tsp.cpp
```

Changing the link works and the file is downloaded:

Workspace1_2021-05-15_00:43:31

### Logs

From `/var/local/log/cms/ContestWebServer-0`: [2021-05-15-00-04-07.log](https://github.com/cms-dev/cms/files/6481483/2021-05-15-00-04-07.log)

### System Information

**CMS version:** [master, or a specific version]
**Was CMS installed:** yes, system-wide
**Using a virtualenv:** no
**Postgres version:** not relevant
**Distribution and version:** Ubuntu 20.04.2 LTS

ranaldmiao:
Fixes issue in #1179

### DIFF
--- a/cms/server/contest/templates/user_test_row.html
+++ b/cms/server/contest/templates/user_test_row.html
@@ -87,7 +87,7 @@
                     {% set real_filename = filename %}
                 {% endif %}
                 <li>
-                    <a href="{{ contest_url("tasks", task.name, "tests", t_idx, "files", filename) }}">
+                    <a href="{{ contest_url("tasks", task.name, "tests", t_idx, "files", real_filename) }}">
                         {{ real_filename }}
                     </a>
                 </li>


andreyv:
Thanks for the quick fix!

Issue #1182, opened 2021-05-22
Use one file cache per machine instead of per service


fagu:
CMS currently uses one file cache per service (by default located in a directory below `/var/local/cache/cms/`). Each worker caches its own copy of each test case file, so if there are W workers on a machine and the total size of the test case files is S, the total disk space used is (eventually) at least W*S. On top of that, there are for example solution executables, etc. This recently caused us to run out of disk space in a contest with lots of workers (about 30) running on the same (powerful) machine. Our temporary solution during the contest was to repeatedly manually shut down some workers, delete their cache directories, and restart them.

Investigating this, @magula and I also discovered that there is code in the Workers https://github.com/cms-dev/cms/blob/dbd6645ed82cce95b8b2a9c99f328669a2040887/cms/service/Worker.py#L67 and the ES https://github.com/cms-dev/cms/blob/dbd6645ed82cce95b8b2a9c99f328669a2040887/cms/service/workerpool.py#L176 that supposedly precaches all files related to the configured contest when a worker connects. (This could speed up grading the first few submissions.) However, this code is never called because ES ignores the `contest_id`! The `ask_contest` parameter is not passed to `default_argument_parser`:
https://github.com/cms-dev/cms/blob/dbd6645ed82cce95b8b2a9c99f328669a2040887/scripts/cmsEvaluationService#L41 (We were in fact lucky that workers didn't precache files, as otherwise deleting a worker's cache directory would not have helped very much in our situation!)

Here are some approaches to reduce disk space consumption. I would be interested to hear your thoughts.
1. @t-lenz suggested a panic button that automates our temporary solution. ;)
2. Have all services on a machine share a cache directory. I think one would at least need to replace https://github.com/cms-dev/cms/blob/dbd6645ed82cce95b8b2a9c99f328669a2040887/cms/db/filecacher.py#L571 with something that atomically moves the temporary file fetched from the database to its final destination if that destination file doesn't exist, and otherwise deletes the temporary file. (Maybe `renameat2` with the `RENAME_NOREPLACE` flag?) I suspect that otherwise there could be issues if one service replaces the file while another is reading from it. Also, I didn't look for additional potential problems. In any case, this approach still wastes resources (network traffic and CPU) as multiple services may simultaneously fetch the same file from the database.
3. Have all services on a machine share a cache directory and use a locking mechanism to ensure no two services fetch the same file at the same time. One would need a way to deal with stale locks from services that crashed while fetching a file.
[...]

PR #1185, opened 2021-06-04
Enable automatic precaching

fagu:
Following the discussion in #1182, this enables automatic precaching of files for the contest specified when starting ES.

Automatic precaching had already been implemented, but didn't take place because ES ignored the `contest_id`. Not ignoring it has several side-effects. (See the commit message for ef99ab0189a1a90c2d576b2882f4f8ef79e15be0.) To re-initiate precaching, you should restart the workers.

Using a file lock, we ensure that on each machine only one worker precaches at any time. (All other workers skip precaching.)

Precaching can take a nontrivial amount of time. (My laptop needs ~1min for ~2.5GB. I don't know how the duration increases if there are workers on many different machines.)

### DIFF
--- a/cms/db/filecacher.py
+++ b/cms/db/filecacher.py
@@ -30,6 +30,7 @@ import io
 import logging
 import os
 import tempfile
+import fcntl
 from abc import ABCMeta, abstractmethod
 
 import gevent
@@ -542,6 +543,36 @@ class FileCacher:
             msg = "Cannot create required directory '%s'." % directory
             logger.error(msg)
             raise RuntimeError(msg)
+
+    def precache_lock(self):
+        """Lock the (shared) cache for precaching if it is currently unlocked.
+
+        Locking is optional: Any process can perform normal cache operations
+        at any time whether the cache is locked or not.
+
+        The locking mechanism's only purpose is to avoid wasting resources by
+        ensuring that on each machine, only one worker precaches at any time.
+
+        return (fileobj|None): The lock file if the cache was previously
+            unlocked. Closing the file object will release the lock.
+            None if the cache was already locked.
+
+        """
+        lock_file = os.path.join(self.file_dir, "cache_lock")
+        fobj = open(lock_file, 'w')
+        returned = False
+        try:
+            fcntl.flock(fobj, fcntl.LOCK_EX | fcntl.LOCK_NB)
+        except BlockingIOError:
+            # This exception is raised only if the errno is EWOULDBLOCK,
+            # which means that the file is already locked.
+            return None
+        else:
+            returned = True
+            return fobj
+        finally:
+            if not returned:
+                fobj.close()
 
     def _load(self, digest, cache_only):
         """Load a file into the cache and open it for reading.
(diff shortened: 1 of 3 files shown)


cms/db/filecacher.py
```diff
@@ -543,6 +545,33 @@ def _create_directory_or_die(directory):
...
+            None if the cache was already locked.
+
+        """
+        lock_file = os.path.join(self.file_dir, "cache_lock")
+        fobj = open(lock_file, 'wb')
+        returned = False
+        try:
+            fcntl.flock(fobj, fcntl.LOCK_EX | fcntl.LOCK_NB)
+        except BlockingIOError as e:
+            if e.errno != errno.EWOULDBLOCK:
```
andreyv, line 565: Is it needed to check `errno` here? We already got a subclass of `OSError` that is raised only for specific `errno` values.

cms/service/Worker.py
```diff
@@ -70,24 +71,33 @@ def precache_files(self, contest_id):
...
-                # No problem (at this stage) if we cannot find the
-                # file
-                pass
-
-        logger.info("Precaching finished.")
+        lock = self.file_cacher.precache_lock()
+        if lock is None:
+            # Another worker is already precaching. Hence, this worker doesn't
+            # need to do anything.
+            return
```
andreyv, line 78: Would you like to add a `logger.info()` or `logger.debug()` call here? It could be useful to find out someone is holding the lock.

cms/db/filecacher.py
```diff
@@ -543,6 +545,33 @@ def _create_directory_or_die(directory):
             logger.error(msg)
             raise RuntimeError(msg)
 
+    def precache_lock(self):
+        """Lock the (shared) cache for precaching if it is currently unlocked.
+
+        The purpose is to avoid wasting resources by ensuring that on each
+        machine, only one worker precaches at any time.
+
```
andreyv, line 553: Would you like to also document here that normal cache operations can be called while the lock is held (including even from the same `FileCacher` in a different greenlet)?

cms/db/filecacher.py
```diff
@@ -543,6 +545,33 @@ def _create_directory_or_die(directory):
...
+        return (fileobj|None): The lock file if the cache was previously
+            unlocked. Closing the file object will release the lock.
+            None if the cache was already locked.
+
+        """
+        lock_file = os.path.join(self.file_dir, "cache_lock")
+        fobj = open(lock_file, 'wb')
```
andreyv, line 560: Minor note: we don't need binary mode here.

andreyv (commented):
Thanks, I checked `contest_id` usage in ES, and it looks like it is correct in both single-contest and multi-contest modes.

Nice idea about using the file object itself as a context manager :)

I added some minor comments, but overall this looks good to merge.

cms/db/filecacher.py
fagu: You're right. Removed the check.

cms/service/Worker.py
fagu: Sure, I've added a `logger.info()` call.

cms/db/filecacher.py
fagu: Done. I clarified that locking is optional and just an optimization.

cms/db/filecacher.py
fagu: Okay, I changed it to 'w'.

fagu (commented):
Thanks again for your help and for the quick reviews!

andreyv:
Thanks, rebased and merged.

PR #1188, opened 2021-08-06
Update to jQuery 3.6.0

andreyv:
This update covers the following GitHub security advisories:
- GHSA-gxr4-xjj5-5px2
- GHSA-jpcq-cgw6-v4j6
- CVE-2019-11358
- CVE-2015-9251

After the update dismissing Bootstrap alerts (`alert` class), shown after submitting a file, on a notification etc., was failing with `Syntax error, unrecognized expression: #`. This is fixed by adding [jquery/jquery-migrate](/jquery/jquery-migrate) (official project) to CWS pages.

I looked into updating Bootstrap itself some time ago as well. It appears that the currently used distribution was customized using a web generator, which is no longer available, and loading a newer version with default settings produces multiple style differences. Thus updating Bootstrap looks to be a separate task, but until then we can use jquery-migrate.

During testing I also found 13abdd7b0d783567788edd21df4060f02b991331, which is not related to the update, but is fixed as well.

Otherwise CWS, AWS (including jqplot), RWS seem to work fine with the new jQuery version.


stefano-maggiolo:
Thanks a bunch, this was long overdue! Good on finding the way to ease the migrate path.

stefano-maggiolo (approved):

andreyv:
Thanks!

PR #1190, opened 2021-09-10
Update Czech translation

SvizelPritula:
The Czech translation was last updated 3 years ago and had many translation strings missing, so I added them.

I have gone over my changes twice and I don't think there are any mistakes. `gettext`'s `msgmerge` rearranged everything to be in the same order as the reference file, which adds quite a lot of changes to this commit.


andreyv:
Thanks for the update!

PR #1193, opened 2021-09-16
Italian contest loader is able to load contests with no participants

### LINKED ISSUE #1192: Italian contest loader fails with no participants
**Description:**

The italian contest loader fails when the contest has no participants.

**Steps to reproduce:**
1. Download the contest example (https://github.com/cms-dev/con_test)
2. Open `contest.yaml` and remove `users` section
3. Import the contest (using `cmsImportContest -L italy_yaml`)

**Expected:**

The contest is added to CMS with all the problems but no participants.

**Actual:**

The application crashes

### System Information

**CMS version:** tested in `v1.4` (but the bug remains in `master`)
**Was CMS installed:** yes
**Using a virtualenv:** no

MarcoUCM:
Fixes #1192 (Italian contest loader fails with no participants)

Please, consider applying the patch to v1.4

### DIFF
--- a/cmscontrib/loaders/italy_yaml.py
+++ b/cmscontrib/loaders/italy_yaml.py
@@ -219,7 +219,7 @@ class YamlLoader(ContestLoader, TaskLoader, UserLoader, TeamLoader):
 
         tasks = load(conf, None, ["tasks", "problemi"])
         participations = load(conf, None, ["users", "utenti"])
-        participations = [] if participations is None else participations;
+        participations = [] if participations is None else participations
         for p in participations:
             p["password"] = build_password(p["password"])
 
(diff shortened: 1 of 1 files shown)


cmscontrib/loaders/italy_yaml.py
```diff
@@ -219,6 +219,7 @@ def get_contest(self):
 
         tasks = load(conf, None, ["tasks", "problemi"])
         participations = load(conf, None, ["users", "utenti"])
+        participations = [] if participations is None else participations;
```
andreyv, line 222:
```suggestion
        participations = [] if participations is None else participations
```
Semicolon is not needed here.

andreyv (changes requested):
Thanks, the fix looks appropriate.

Please see one minor comment below — I don't have access to edit this branch.

cmscontrib/loaders/italy_yaml.py
MarcoUCM: Ups. Done.

andreyv (approved):
Thanks!

PR #1196, opened 2021-09-22
BugFix: Italian contest exporter and communication tasks

### LINKED ISSUE #1195: Italian contest exporter misconfigure some communication tasks
**Description:**

The italian contest exporter does not adjust some properties properly when exporting communication tasks. When the exporter infers that the tasks is a "communication task" it automatically assumes that "Submissions are compiled into a stub" and that "User process read from and write to fifos, whose paths are given as arguments".

However IMHO that assumption should only be made when at least a stub is provided in the task description. If no one is provided then the task should be configured as "Submissions are self-sufficient" and "User processes read from stdin and write to stdout" as in regular batch tasks.

**Steps to reproduce:**

1. Download the contest example (https://github.com/cms-dev/con_test)
2. Remove `con_test/communication/sol/stub*` in order to simulate a communication problem without stubs. Obviously, the task is not valid anymore because the statement, solution and manager expect those stubs, but it is just a quick-and-dirty way of having a communication task that (apparently) communicates via stdin/stdout with the validator.
3. Import the contest (using cmsImportContest -L italy_yaml)

**Expected:**

In the Dataset section of the task details `communication` in the front-end the configuration should be:

   - Task type = `communication`
   - Number of Processes =  `1`
   - Compilation: `Submissions are self-sufficient`
   - User I/O: `User processes read from stdin and write to stdout`

**Actual:**

The loaded configuration is

   - :heavy_check_mark: Task type = `communication` 
   - :heavy_check_mark: Number of Processes =  `1`
   - :x: Compilation: `Submissions are compiled into a stub`
   - :x: User I/O: `User process read from and write to fifos, whose paths are given as arguments`

### System Information

**CMS version:** tested in `v1.4` (but the bug remains in `master`)
**Was CMS installed:** yes
**Using a virtualenv:** no

MarcoUCM:
Fixes #1195 (Italian contest exporter misconfigure some communication tasks)

Please, consider applying the patch to v1.4

When the task is a communication task and it has no stubs,
exporter assumes "alone" and "std_io".
It only changes its mind to "stub" and "fifo_io" when a stub of
any language is found.

### DIFF
--- a/cmscontrib/loaders/italy_yaml.py
+++ b/cmscontrib/loaders/italy_yaml.py
@@ -629,7 +629,7 @@ class YamlLoader(ContestLoader, TaskLoader, UserLoader, TeamLoader):
                     logger.info("Task type Communication")
                     args["task_type"] = "Communication"
                     args["task_type_parameters"] = \
-                        [num_processes, "stub", "fifo_io"]
+                        [num_processes, "alone", "std_io"]
                     digest = self.file_cacher.put_file_from_path(
                         path,
                         "Manager for task %s" % task.name)
@@ -643,6 +643,8 @@ class YamlLoader(ContestLoader, TaskLoader, UserLoader, TeamLoader):
                                 stub_name,
                                 "Stub for task %s and language %s" % (
                                     task.name, lang.name))
+                            args["task_type_parameters"] = \
+                                [num_processes, "stub", "fifo_io"]
                             args["managers"] += [
                                 Manager(
                                     "stub%s" % lang.source_extension, digest)]


andreyv:
Thanks for the fix!

It makes sense to assume such defaults for I/O type.

But also note that I/O type is orthogonal to the stubs' presence. This is also mentioned in CMS documentation.

What do you think about enhancing the fix with one more task parameter, `user_io`, similar to how `num_processes` is already handled nearby? It could be used to override the default choice from your suggestion.

PR #1197, opened 2021-11-25
Complete polygon loader language list

Grackins:
Hi!

There was a TODO in code about adding other language codes to polygon loader.

I found a list of languages in polygon and just added them to the loader dict.

However there are some concerns:
1. Is there a cleaner approach compared to letting a big constant dictionary be in the middle of code?
2. What language code to choose for "other"? Current choice is blank string.

Regards

### DIFF
--- a/cmscontrib/loaders/polygon.py
+++ b/cmscontrib/loaders/polygon.py
@@ -5,6 +5,7 @@
 # Copyright © 2014-2017 Stefano Maggiolo <s.maggiolo@gmail.com>
 # Copyright © 2018 Edoardo Morassutto <edoardo.morassutto@gmail.com>
 # Copyright © 2018 Luca Wehrstedt <luca.wehrstedt@gmail.com>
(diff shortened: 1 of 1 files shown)


andreyv:
Hi,

Thanks for the improvement :)

1. Good question. I was looking at [Babel documentation](https://babel.pocoo.org/en/latest/index.html), and it doesn't seem to provide a straightforward way to convert a language name in English to a language code. The other direction is possible through [`Locale.english_name`](https://babel.pocoo.org/en/latest/api/core.html#babel.core.Locale.english_name).
   Anyway, unless the code is changed to glob `statements/.pdf/*/problem.pdf` instead of trying all languages, the custom list is needed either way to know languages supported by Polygon. So, I think, leaving the list as is is fine too.
2. The empty language name creates some artifacts in CWS when multiple statements are added (it is not possible to select it from the sidebar etc.). Another possible choice is literally mapping `other` to `Other`, which is allowed too.

andreyv:
I applied the above suggestion about `other` language. Code looks good, merging.

PR #1199, opened 2021-12-11
Fix CI

edomora97:
`apt install` may fail without a prior `apt update`.

If you accept this, this commit may also be cherry-picked in #1141 #1142 and #1198.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -38,8 +38,8 @@ jobs:
 
       - name: Install dependencies
         run: |
-          sudo apt update -y
-          sudo apt install -y \
+          sudo apt-get update -y
+          sudo apt-get install -y \
             build-essential \
             php7.4-cli \
             haskell-platform \
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,22 +37,23 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Install dependencies
-        run: >-
-          sudo apt install -y
-          build-essential
-          php7.4-cli
-          haskell-platform
-          fp-compiler
-          python2.7
-          cppreference-doc-en-html
-          cgroup-lite
-          libcap-dev
-          python-dev
-          libpq-dev
-          libcups2-dev
-          libyaml-dev
-          rustc
-          mono-mcs
+        run: |
+          sudo apt update -y
+          sudo apt install -y \
+            build-essential \
+            php7.4-cli \
+            haskell-platform \
+            fp-compiler \
+            python2.7 \
+            cppreference-doc-en-html \
+            cgroup-lite \
+            libcap-dev \
+            python-dev \
+            libpq-dev \
+            libcups2-dev \
+            libyaml-dev \
+            rustc \
+            mono-mcs
 
       - name: Install requirements
         run: |


edomora97:
I can't figure out why the tests now fail, in my fork they [seem to succeed](https://github.com/edomora97/cms/runs/4493751130?check_suite_focus=true). Are they flacky? 🙈

.github/workflows/main.yml
```diff
@@ -37,22 +37,23 @@ jobs:
...
-          libcap-dev
-          python-dev
-          libpq-dev
-          libcups2-dev
-          libyaml-dev
-          rustc
-          mono-mcs
+        run: |
+          sudo apt update -y
+          sudo apt install -y \
```
andreyv, line 42: You can still use the folded style (`>`) with multiple commands: separate the commands with an empty line.

.github/workflows/main.yml
```diff
@@ -37,22 +37,23 @@ jobs:
...
-          cgroup-lite
-          libcap-dev
-          python-dev
-          libpq-dev
-          libcups2-dev
-          libyaml-dev
-          rustc
-          mono-mcs
+        run: |
+          sudo apt update -y
```
andreyv, line 41: While we are changing this, could you please also replace `apt` with `apt-get`? `apt` gives
```
WARNING: apt does not have a stable CLI interface. Use with caution in scripts.
```

andreyv (commented):
For reference, official recommendation to use `apt-get update`: https://docs.github.com/en/actions/using-github-hosted-runners/customizing-github-hosted-runners#installing-software-on-ubuntu-runners

andreyv:
> I can't figure out why the tests now fail, in my fork they [seem to succeed](https://github.com/edomora97/cms/runs/4493751130?check_suite_focus=true). Are they flacky? see_no_evil

Hmm, I restarted the jobs and they now pass. I haven't seen these errors before. The failure messages hint at some pattern, but at a first look there's nothing out of line in the relevant tests.

.github/workflows/main.yml
edomora97: Ups, you're right, I'll fix this in the next commit.

.github/workflows/main.yml
edomora97: Wow, YAML has a bit too many string variants 😅

It's just a pity that

```yaml
run: >
  sudo apt-get update -y

  sudo apt-get install
    aaa
    bbb
    ccc
```

becomes `"sudo apt-get update -y\nsudo apt-get install\n  aaa\n  bbb\n  ccc"`.
I'm not a fan of not indenting the packages a bit, and I think that with the backslashes (while a bit more verbose) the command is much more clear (at least to the ones not that expert in YAML). What do you think?

andreyv: Hmm, I understand the reasoning, although `\` feels like it reduces the clarity, too. You could actually use the plain flow style:
```yaml
run:
  sudo apt-get update
  
  sudo apt-get install
    aaa
    bbb
    ccc
```
How does this variant look?

edomora97: The more I look at YAML, the less I feel to know about it.

For sure this looks the cleanest, but when I look at those CI steps I imagine them as "bash scripts" that are executed one after the other. Without `\` and with just YAML magic I felt a bit confused on how they should be read. This is the cleanest variant but the furthest to a "bash script", and I feel like that installing packages should be closer to bash territory.

In any case, I don't think this is super important and mine is not a strong opinion, either way is fine for me. I'll leave to you the final decision.

andreyv: Alright, let's leave it like that.

PR #1200, opened 2021-12-11
Prometheus exporter

edomora97:
Add a new command: `cmsPrometheusExporter`

This will start a new HTTP server that listens on a port (defaults to 8811) for requests to /metrics. At this endpoint many metrics about a contest (or all contests) are exposed for Prometheus to be collected. Those metrics consider both system values (queue status, connected workers, ...) and contest values (submissions, users, questions, ...)

This exporter works by executing simple queries to the database, and obtains the queue information asking Evaluation Service.

The metrics exposed may leak information about the contest, so it's important to secure this endpoint (for example using a reverse proxy, or binding localhost).

With this exporter it's possible to build interactive and real-time dashboard for monitoring the status of the contest. For example, you can find [**here**](https://snapshot.raintank.io/dashboard/snapshot/0J29w0KruEiymy6zV30beX98aRG6njiX?orgId=2) the dashboard we used at OII 2020, and [**here**](https://snapshot.raintank.io/dashboard/snapshot/JngCBt71ZOJhiyEUC7yxZhS3guZmZYYi?orgId=2) the one used at an OIS round.
Those dashboards are built using Grafana, and we configured it also for sending telegram notifications when something wrong was happening (e.g. queue too long, see image below).

image

### DIFF
--- a/cmscontrib/PrometheusExporter.py
+++ b/cmscontrib/PrometheusExporter.py
@@ -1,0 +1,293 @@
+#!/usr/bin/env python3
+
+# Contest Management System - http://cms-dev.github.io/
+# Copyright © 2021 Edoardo Morassutto <edoardo.morassutto@gmail.com>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# We enable monkey patching to make many libraries gevent-friendly
+# (for instance, urllib3, used by requests)
+import gevent.monkey
+
+gevent.monkey.patch_all()  # noqa
+
+import argparse
+
+from prometheus_client import start_http_server
+from prometheus_client.core import REGISTRY, CounterMetricFamily, \
+    GaugeMetricFamily
+from sqlalchemy import func, distinct
+
+from cms import ServiceCoord
+from cms.db import Announcement, Dataset, Message, Participation, Question, \
+    SessionGen, Submission, SubmissionResult, Task, base
+from cms.io.service import Service
+from cms.server.admin.server import AdminWebServer
+
+
+class PrometheusExporter(Service):
+    def __init__(self, args):
+        super().__init__()
+
+        self.contest_id = args.contest_id
+        self.export_subs = not args.no_submissions
+        self.export_workers = not args.no_workers
+        self.export_queue = not args.no_queue
+        self.export_communiactions = not args.no_communications
+        self.export_users = not args.no_users
+        self.evaluation_service = self.connect_to(
+            ServiceCoord("EvaluationService", 0))
+
+    def collect(self):
+        with SessionGen() as session:
+            if self.export_subs:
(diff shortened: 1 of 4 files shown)


cmscontrib/PrometheusExporter.py
```diff
@@ -0,0 +1,293 @@
...
+    def collect_workers(self):
+        status = self.evaluation_service.workers_status().get()
+        metric = GaugeMetricFamily(
+            "cms_workers",
+            "Number of cmsWorker instances",
+            labels=["status"],
+        )
+        metric.add_metric(["total"], len(status))
+
+        connected = len([1 for worker in status.values() if worker["connected"]])
```
andreyv, line 120: You can use generator expressions to avoid constructing the full list: `sum(1 for … in … if …)`. Similarly below.

cmscontrib/PrometheusExporter.py
```diff
@@ -0,0 +1,293 @@
...
+        action="store_true",
+    )
+
+    args = parser.parse_args()
+
+    service = PrometheusExporter(args)
+    REGISTRY.register(service)
+    start_http_server(args.port, addr=args.host)
+    print("Started at http://%s:%s/metric" % (args.host, args.port))
+    Service.run(service)
```
andreyv, line 289: Why not `service.run()`?

cmscontrib/PrometheusExporter.py
```diff
@@ -0,0 +1,293 @@
...
+
+        yield metric
+
+
+def main():
+    """Parse arguments and launch process."""
+    parser = argparse.ArgumentParser(description="Prometheus exporter.")
+    parser.add_argument(
+        "--host",
+        help="IP address to bind",
```
andreyv, line 242:
```suggestion
        help="IP address to bind to",
```

requirements.txt
```diff
@@ -25,3 +25,6 @@ pyyaml>=5.3,<5.4  # http://pyyaml.org/wiki/PyYAML
 # Only for printing:
 pycups>=1.9,<1.10  # https://pypi.python.org/pypi/pycups
 PyPDF2>=1.26,<1.27  # https://github.com/mstamy2/PyPDF2/blob/master/CHANGELOG
+
+# Only for cmsPrometheusExporter
+prometheus-client>=0.7,<0.8  # https://pypi.org/project/prometheus-client
```
andreyv, line 30: Documentation should also be updated at "Installing CMS and its Python dependencies". We need `python3-prometheus-client` for Ubuntu and `python-prometheus_client` for Arch Linux as optonal deps.

cmscontrib/PrometheusExporter.py
```diff
@@ -0,0 +1,293 @@
+#!/usr/bin/env python3
```
andreyv, line 1: The script file should have the executable permission.

andreyv (commented):
Thanks, this looks great! The dashboard examples are nice.

I see that `prometheus_client.start_http_server()` uses `threading`, which should be patched by gevent — OK.

I noticed that:
1. The service crashes on start if CMS (`EvaluationService`, to be precise) is not running.
2. The service returns HTTP 500 if CMS/`EvaluationService` is stopped.

As a status reporting service, it is probably good to make it more robust to the state of other services.

cmscontrib/PrometheusExporter.py
```diff
@@ -0,0 +1,293 @@
...
+        help="Do not export users metrics",
+        action="store_true",
+    )
+
+    args = parser.parse_args()
+
+    service = PrometheusExporter(args)
+    REGISTRY.register(service)
+    start_http_server(args.port, addr=args.host)
+    print("Started at http://%s:%s/metric" % (args.host, args.port))
```
andreyv, line 288: This code could be moved to a new method in `PrometheusExporter`. In fact, `PrometheusExporter.run()` could also be overridden and this new method be called from there.

veluca93:
Closed and continued in #1303

Issue #33, opened 2012-10-07
Decide how we round the scores


lw:
At the moment the ScoreTypes are rounding the score and the max_score (both the public and the private ones) at two decimal places and, where applicable, the subtask score isn't rounded at all (we're using the %g output format).

At the IOI we were required by the rules to round the score to the nearest integer and patched our code to do so. Since we had many variable-score subtasks we also decided to round their score at two decimal places.

I think it's now time to decide what to do: keep the current rounding behavior, use the IOI one or provide some kind of configuration options for it.

PR #1211, opened 2022-07-01
fix: pin MarkupSafe to 2.0.1 (#33)

SvizelPritula:
Cherry-pick commit e718c6206e1d1a08c34e3703587c14522e86b391 from ioi2022/cms, which should fix tests breaking due to beaking changes in MarkupSafe.


myungwoo:
⬆️

wil93:
Thanks! I'm taking a look into this issue, a quick feedback in the meantime: I think it would be preferable to avoid adding MarkupSafe as a dependency since it's not actually a dependency for us (it's a dependency of one of our dependencies, Jinja2) so maybe it would be cleaner to solve this by upgrading Jinja2 instead.

wil93:
Merged like this for now, as it seems that the only other solution would be pinning all dependencies (e.g. using pip-tools or pipenv) and we can do that at a later time. Thanks for the fix!

PR #1221, opened 2022-12-15
Use docker-compose to run tests

wil93:
Move the testing logic from the Github Actions format to a `docker-compose.test.yml` file. The advantage of this approach is that we can now run tests locally in a easy and reproducible way.

Moreover, instead of using `sed` at runtime to inject the test database string into `cms.conf`, we now always add a `cms-testdb.conf` file next to `cms.conf` in the `/usr/local/etc/` folder as part of the build step of the docker container. We then decide which config file to use based on the already-supported `CMS_CONFIG` environment variable. (I had to patch `testrunner.py` which was not supporting that variable yet.)

One possible next step in this direction could be using this approach also for local development, with a `cms-devdb.conf` file which likewise always gets copied in the etc folder, and a `docker-compose.dev.yml` which uses it to start the DB and the local CMS instance. It's also possible to re-use the `Dockerfile` in production, but I wouldn't try to support that use case seriously just yet (I think the contest admins can write a `docker-compose.yml` fitting their specific needs).


Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+
+WORKDIR /home/cmsuser/cms
+
+RUN sudo pip3 install -r requirements.txt
+RUN sudo pip3 install -r dev-requirements.txt
+RUN sudo python3 setup.py install
+
+RUN sudo python3 prerequisites.py --yes --cmsuser=cmsuser install
+
+RUN sudo cp config/cms-testdb.conf /usr/local/etc/cms-testdb.conf
```
stefano-maggiolo, line 48: IMHO using `sed` is better than keeping a second copy of the file almost identical, that we'll have to remember to keep in sync.

stefano-maggiolo (commented):
I don't know docker, so I'll let you test that it works - it seems you're also still tweaking codecov. For the moment I left a comment for you, the rest looks fine for as much I know.

Dockerfile
wil93: You're probably right, I changed it back to a sed. The main thing I wanted to achieve was to do "everything at build time" (i.e. in the Dockerfile) so that at runtime we just need to choose the config file to use. This is still possible with sed (running it in the Dockerfile) so we can do that.

Another possible approach here could be to remove the connection string from the config file and have it as an environment variable, which would make things a bit easier / cleaner, considering that IIRC in the future we want to replace the config files pretty much completely in favor of storing more and more settings in the DB.

cmstestsuite/RunFunctionalTests.py
```diff
@@ -173,8 +173,8 @@ def main():
         help="send coverage results to Codecov (requires --coverage)")
     g = parser.add_mutually_exclusive_group()
     g.add_argument(
-        "--coverage", action="store_true",
-        help="compute line coverage information")
+        "--coverage", action="store", type=utf8_decoder,
+        help="path to the XML coverage report file (if not specified, coverage is not computed)")
```
stefano-maggiolo, line 177: Nit: wrap? (would be nice if github showed the max line length)

cmstestsuite/RunUnitTests.py
```diff
@@ -149,8 +149,8 @@ def main():
         help="send coverage results to Codecov (requires --coverage)")
     g = parser.add_mutually_exclusive_group()
     g.add_argument(
-        "--coverage", action="store_true",
-        help="compute line coverage information")
+        "--coverage", action="store", type=utf8_decoder,
+        help="path to the XML coverage report file (if not specified, coverage is not computed)")
```
stefano-maggiolo, line 153: Nit: wrap

stefano-maggiolo (approved):
Thank you! Please look at those two nits and squash the commits as necessary for them to make sense before pushing!

PR #1222, opened 2022-12-18
Upgrade to isolate v2

### LINKED ISSUE #1257: Switch to Isolate v2
I just released v2.0 of Isolate, which is based on CGroup v2 and seamlessly compatible with systemd. Therefore it runs on current Linux distributions without need of adding obscure kernel switches.

Could you please update CMS to use it?

Little changes are needed, perhaps only not giving the `--cg-timing` switch, which is no longer supported. This timing mode was made default a long time ago and it is the only timing mode supported by Isolate v2.

wil93:
In order to upgrade to isolate v2 we need to:

- [x] Switch the Github Actions runner from 20.04 to at least 22.04, where ubuntu seems to drop the unsupported "unified cgroups" hierarchy and only provides the new cgroup v2. It might make sense to just go with 24.04 since anyway we only use this to call docker, but let's keep 22.04 for now to be conservative.
- [x] Add `cgroup: host` to the docker-compose file to run the container in the host's cgroup namespace ([docs](https://docs.docker.com/reference/cli/docker/container/create/#options:~:text=cgroupns)).
- [x] Currently isolate is installed by `prerequisites.py` with default settings. In the default settings, isolate expects `/run/isolate/cgroup` to exist. This normally gets created by `isolate-cg-keeper`, which is what the `isolate.service` systemd unit is meant for. If we want users to install and enable that unit, then we should add it to the documentation (and possibly make `prerequisites.py` install the unit). I'm not sure however if it's really needed to run that, since just running `echo /sys/fs/cgroup > /run/isolate/group` manually seems to also work. Maybe we can instead configure isolate to always use `/sys/fs/cgroup` (I think we can do it by changing the default configuration in `/usr/local/etc/isolate`) and document this instead? We need to figure out which of these approaches are OK, and add it to the documentation.
    **UPDATE:** After discussion with @veluca93, we decided to create a local `./config/isolate.conf.sample` file (alongside the `cms.conf.sample`, `nginx.conf.sample`, etc) with `cg_root` overridden to `/sys/fs/cgroup`, and install that file as part of the `prerequisites.py install` step instead of the default isolate config file.

Fixes #1257

### DIFF
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -1,5 +1,3 @@
-version: "3.3"
-
 services:
   testdb:
     image: postgres
@@ -19,6 +17,7 @@ services:
     volumes:
       - "./codecov:/home/cmsuser/cms/codecov"
     privileged: true
+    cgroup: host
     command: >
       wait-for-it testdb:5432 -- sh -c "
       dropdb --host=testdb --username=postgres cmsdbfortesting ;
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   test:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
 
     steps:
       - uses: actions/checkout@v2
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,7 @@
 # syntax=docker/dockerfile:1
 FROM ubuntu:20.04
 
-RUN apt-get update
-RUN apt-get install -y \
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y \
     build-essential \
     cgroup-lite \
     cppreference-doc-en-html \
--- a/cms/grading/Sandbox.py
+++ b/cms/grading/Sandbox.py
@@ -1076,7 +1076,7 @@ class IsolateSandbox(SandboxBase):
         if self.box_id is not None:
             res += ["--box-id=%d" % self.box_id]
         if self.cgroup:
-            res += ["--cg", "--cg-timing"]
+            res += ["--cg"]
         if self.chdir is not None:
             res += ["--chdir=%s" % self.chdir]
         for src, dest, options in self.dirs:
--- a/config/isolate.conf.sample
+++ b/config/isolate.conf.sample
@@ -1,0 +1,28 @@
+# This is a configuration file for Isolate
+
+# All sandboxes are created under this directory.
+# To avoid symlink attacks, this directory and all its ancestors
+# must be writeable only to root.
+box_root = /var/local/lib/isolate
+
+# Directory where lock files are created.
+lock_root = /run/isolate/locks
+
+# Control group under which we place our subgroups
+# Either an explicit path to a subdirectory in cgroupfs, or "auto:file" to read
+# the path from "file", where it is put by isolate-cg-helper.
+cg_root = /sys/fs/cgroup
+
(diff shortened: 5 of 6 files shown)


docker-compose.test.yml
```diff
@@ -1,5 +1,3 @@
-version: "3.3"
```
wil93, line 1: Removed because deprecated, and docker compose in ubuntu 24.04 warns against using it.

CristianCantoro:
To clarify, until this gets merged we should still use the kernel flags as detailed in https://github.com/ioi/isolate/issues/35, correct?

At the moment, using `v1.4.rc1`, I get the following cgroups-related error:
```python
Failed to create control group /sys/fs/cgroup/memory/box-0/: No such file or directory
  File "/home/cms/cms_venv/bin/cmsMake", line 11, in <module>
    load_entry_point('cms==1.4rc1', 'console_scripts', 'cmsMake')()
  File "/home/cms/cms_venv/lib/python3.6/site-packages/cms-1.4rc1-py3.6.egg/cmstaskenv/cmsMake.py", line 755, in main
    assume=assume)
  File "/home/cms/cms_venv/lib/python3.6/site-packages/cms-1.4rc1-py3.6.egg/cmstaskenv/cmsMake.py", line 692, in execute_multiple_targets
    already_executed, debug=debug, assume=assume)
  File "/home/cms/cms_venv/lib/python3.6/site-packages/cms-1.4rc1-py3.6.egg/cmstaskenv/cmsMake.py", line 682, in execute_target
    action(assume=assume)
  File "/home/cms/cms_venv/lib/python3.6/site-packages/cms-1.4rc1-py3.6.egg/cmstaskenv/cmsMake.py", line 269, in test_src
    assume=assume)
  File "/home/cms/cms_venv/lib/python3.6/site-packages/cms-1.4rc1-py3.6.egg/cmstaskenv/Test.py", line 163, in test_testcases
    tasktype.evaluate(job, file_cacher)
  File "/home/cms/cms_venv/lib/python3.6/site-packages/cms-1.4rc1-py3.6.egg/cms/grading/tasktypes/Batch.py", line 290, in evaluate
    sandbox = create_sandbox(file_cacher, name="evaluate")
  File "/home/cms/cms_venv/lib/python3.6/site-packages/cms-1.4rc1-py3.6.egg/cms/grading/tasktypes/util.py", line 71, in create_sandbox
    sandbox = Sandbox(file_cacher, name=name)
  File "/home/cms/cms_venv/lib/python3.6/site-packages/cms-1.4rc1-py3.6.egg/cms/grading/Sandbox.py", line 944, in __init__
    self.initialize_isolate()
  File "/home/cms/cms_venv/lib/python3.6/site-packages/cms-1.4rc1-py3.6.egg/cms/grading/Sandbox.py", line 1420, in initialize_isolate
    "(error %d)" % (pretty_print_cmdline(init_cmd), ret))
cms.grading.Sandbox.SandboxInterfaceException: Failed to initialize sandbox with command: isolate --cg --box-id=0 --init (error 2)
```

veluca93 (approved):

gollux:
I completely agree that overriding `cg_root` on systems with systemd is wrong because it's in direct violation of systemd's cgroup delegation principles. Please don't do that.

I would very much like CMS to stop installing isolate. It's a separate project and users should run its own installation mechanism. I would like to provide Debian/Ubuntu packages of Isolate, which install the systemd service following Debian conventions. I welcome other people to contribute packaging for their favorite distributions.

Issue #119, opened 2013-03-06
Consider using another data format for configuration files


lw:
JSON is cool because it's an intuitive and concise language to describe simple data structures. But I don't think it's suited for configuration files, mainly because it lacks the ability to add comments. We worked around this by using underscore-prefixed keys (and abusing JSON's flexibility on not giving errors when the same key is defined many times?), but they're quite ugly, long, counter-intuitive and can lead to stupid mistakes (see issue #117). I therefore suggest to move to another format, better suited for that purpose.

This [1] could be a good candidate, since it's in the standard library and it's a common format (used, among others, by Python's setup.cfg files, git, openssl, etc.). Yet, I'm not sure it supports list values (which we're using a lot). Unfortunately there's no formal description of the language, so no easy way to discover it but trying...

Another option could be YAML. It's a far more powerful markup language that JSON, we're already using it in YamlImporter and I also like it aesthetically because it resembles Python (because of the indentation, I guess...).

The final option I see is XML, but I think this is too much complexity.

What do you think? Do you also feel that JSON is limited? What format would you prefer to use instead?

[1] http://docs.python.org/2/library/configparser.html

gollux:
BTW: JSON5 could be a nice compromise between usability and simplicity.

wil93:
I think nowadays the format of choice would be TOML: https://toml.io/en/

But if we'd like to keep the JSON while adding comments, there's [this recent one by Microsoft](https://code.visualstudio.com/docs/languages/json#_json-with-comments). It looks less intrusive than JSON5 (which would allow, among other things, to use either single or double quotes, which is probably not really an issue that needed to be solved as urgently as the comments)

gollux:
For me, absence of comments is the second worst problem. The one that annoys me the most is illegality of traling commas (I lost count of how many times I was burned by this!).

magula:
If CMS would switch to a standard made for configuration files, like TOML, it would indeed make configuration files a lot more readable and ease configuration significantly. See [this snippet](https://gist.github.com/magula/b5bd7a3451bb35595c693b6a12574b71) for what it may look like.

That would require people to rewrite their existing configuration files (or use a script provided for that). It would be interesting to know if people say they would be affected by that so much it would not be worth it?

If CMS does decide to switch to TOML, there would be some things to consider:

_TOML supports sections._
[...]

PR #1224, opened 2023-02-01
Change configuration format from JSON to TOML

magula:
In light of #119, this would change the format of the `cms.conf` and `cms.ranking.conf` configuration files from JSON to TOML. Apart from the obvious, the changes include:

* Transforms the `_section` headers of the previous JSON format into TOML sections.
The TOML sections of `cms.conf` correspond to predefined `dataclasses` in Python, i.e. there is an object for each TOML section that contains its attributes. E.g., `keep_sandbox` can be accessed as `config.worker.keep_sandbox`, while previously it would have just been `config.keep_sandbox`. Some variable names also change: `config.admin_cookie_duration` becomes `config.aws.cookie_duration`, and `config.cookie_duration` becomes `config.cws.cookie_duration`.
This should structure the attributes better and provide more consistent key names.

* The structure of `cms.ranking.conf` is left untouched, the file is just translated to TOML.

* Includes a check if the config file that would be used is in JSON format. If so, the user is warned about the change and advised to tranform the config into the new format. He is also presented with an attempt at translating the file.
This works by injecting values taken from the JSON config into a Jinja template of the bare TOML config.
If an unexpected key is present in the JSON config, the key-value pair is put in a section called `stray`. When loading a TOML config, everything in `stray` is stored as a field of `Config`, so it can still be accessed as `config.[keyname]`.
The automatic translation works only heuristically. The attributes are rendered in the Jinja template using `tojson`, so there is no guarantee the outcome will adhere to the TOML specification.
One could use a TOML writer instead, but that would add another package dependency.

Note that these changes have not been tested thoroughly.

Any feedback is very welcome!


setup.py
```diff
@@ -36,6 +37,9 @@
 
 
 PACKAGE_DATA = {
+    "cms": [
+        "cms_conf_legacy_mapping.toml",
```
magula, line 41: This is a strange path to have this template, but I don't know where it would fit in. Perhaps the whole translation function should be transfered somewhere else (away from `cms/Config.py`) anyway?

wil93: I think it's fine here, but maybe calling it `cms_conf_legacy_mapping.toml.jinja` would make more sense? What do you think?

wil93 (commented):
Thank you so much for taking care of this important task!

A thought about the use of sections: I get the desire of organizing keys into related groups, however maybe it would be better to keep all the keys under the same "namespace" as they were in the JSON, as to reduce the necessary changes to the absolute minimum (thus minimizing the risk of new bugs) especially given that we don't have a type system helping us to verify with certainty that all config usages are correctly migrated. Doing this would also allow us to continue supporting JSON for backwards-compatibility (I like the "heuristically generated" TOML version, it will be very helpful to users 😄 however I think that for some time can still support the JSON file when a TOML version is missing).

Also, I remember that some time ago there was some discussion about wanting to move many of the configuration options away from `cms.conf` and into the database: ideally, the `cms.conf` file would almost exclusively be used to specify the database connection string, while most of the options would be stored in the database itself (this would make it easier to change the values, e.g. "keep_sandbox" could be a simple checkbox in the AWS panel) and with this scenario in mind, I think it would make more sense to not organize options into sections.

Another thing: maybe we can update the file name to include `.toml` as the extension? I think there could be some value in indicating this extension (it clearly communicates that it's a TOML file so users, and most importantly their IDE / editors, will expect to find TOML syntax inside of it). For example we could change `cms.conf` to `cms.toml`, or if we really want to keep `conf` we could use`cms-conf.toml` or `cms.conf.toml`.

What do you think?

wil93:
> Would you mind if we still renamed
> 
> * `cookie_duration` -> `contest_cookie_duration` and
> * `num_proxies_used` -> `contest_num_proxies_used`,
> 
> but to preserve compatibility used the old names as aliases when loading the config? These changes should be small enough to reasonably test manually.

Yeah that sounds reasonable (I assume these aliases would work both in the legacy JSON file and the new TOML one?)



> Should we keep
> 
> * the attributes' type annotations, and
> * the check in the config loader for each attribute key if that config attribute exists (up until now, a typo in a key would have gone unnoticed and the default value been used inadvertently)?

The type annotations are a big plus in my opinion 😄 

As for the second point, one concern I have is that (if I remember correctly) sometimes for a contest we would remove keys from the configuration file since anyway they were set to the default value, thus creating a "minimal" configuration file that only ovverrides what is strictly necessary. Would this change break such "minimal" configuration files?


> PS: I am not sure I would be happy about moving much of the configuration to the database, actually. We tend to use the same basic configuration for new setups, and then only adapt where it is necessary.

That's a great input, I will keep this use-case in mind. I think for now it's fine to continue using a configuration file, and if we move to DB-configuration then we should look into having a file representation anyway to support this use-case (e.g. exporting the settings to a toml file, and being able to restore them)

cms/conf.py
```diff
@@ -178,74 +214,107 @@ def __init__(self):
...
                           os.path.dirname(__file__),
-                          '..', 'config', 'cms.conf'))]
-            paths += [os.path.join("/", "usr", "local", "etc", "cms.conf"),
-                      os.path.join("/", "etc", "cms.conf")]
+                          '..', 'config'))]
+            etc_paths += [os.path.join("/", "usr", "local", "etc"),
+                      os.path.join("/", "etc")]
+
+        legacy_paths = [os.path.join(p, "cms.conf") for p in etc_paths]
+        paths = [os.path.join(p, "cms.toml") for p in etc_paths]
```
wil93, line 233: I was thinking that maybe it would make sense here to not differentiate between `legacy_paths` and normal `paths`, we could just have `paths` (with the legacy ones appended at the end) and simply try to read sequentially each one of them: as soon as we find a file that we manage to read we return the configuration found there. We would try to read in JSON basically only when reading in TOML raises some kind of format exception. I think the code would be simpler. What do you think?

config/cms.ranking.toml.sample
```diff
@@ -0,0 +1,11 @@
...
+bind_address = ""
+
+# Listening port for RankingWebServer.
+http_port = 8890
+
+# Login information for adding and editing data.
+username = "usern4me"
+password = "passw0rd"
+
+# This is the end of this file.
```
wil93, line 11: nit: I think this line is not really useful, we could remove it (same comment for other TOML files)

magula:
> Yeah that sounds reasonable (I assume these aliases would work both in the legacy JSON file and the new TOML one?)

Yes, they will work in both.

> As for the second point, one concern I have is that (if I remember correctly) sometimes for a contest we would remove keys from the configuration file since anyway they were set to the default value, thus creating a "minimal" configuration file that only ovverrides what is strictly necessary. Would this change break such "minimal" configuration files?

It would not enforce keys to be present in the configuration file, so such a minimal configuration file will still work just the same. It just warns about config keys that it does not know about, instead of loading them silently. (A warning on missing keys is also included right now, but I will remove that again when I revert the changes of the config structure.)

> That's a great input, I will keep this use-case in mind. I think for now it's fine to continue using a configuration file, and if we move to DB-configuration then we should look into having a file representation anyway to support this use-case (e.g. exporting the settings to a toml file, and being able to restore them)

Sounds good, thanks!

magula:
I removed the section headers (or table headers, as TOML calls it) in the last commit, so the structure is as before.

However, there seems to be no (legible) way to structure `core_services` and `other_services` if not as tables, because [inline tables](https://toml.io/en/v1.0.0#inline-table) can not contain newlines. This is a problem because the top-level table ends with the first table, so nothing after that could be top-level again. Thus I moved these to the end of the TOML file in the last commit. They are now the only parts contained in a table, and they have to be at the end of the file, which seems unelegant. Do you see a better way to do this?

cms/conf.py
magula: I agree. I put the legacy paths first though, in case the new sample config is copied during installation: Otherwise, this sample config would be used inadvertently and the message about the old config file would not appear. So now,
* a warning is going to be raised as soon as one of the files can be read in JSON format before one can be read in TOML format, and
* this JSON config will then proceed to be loaded.

Let me know if I should change this.

wil93:
Now that we have pytest as test runner, adding tests should be easier so it would be nice to add a few tests, for example:

- a test to make sure that the values from the TOML / JSON file indeed are reflected in the config object
- a test to verify the logic of choosing TOML over JSON when available, and falling back to JSON when not available

Let me know if you would be down to add these, if not I can try to spend some time on it when I get free.

cms/conf.py
```diff
@@ -215,69 +227,158 @@ def _load(self, paths):
...
+                for shard_number, shard in \
+                        enumerate(data[part][service]):
+                    coord = ServiceCoord(service, shard_number)
+                    getattr(self.async_config, part)[coord] = Address(*shard)
+            del data[part]
+
+        # These keys have been renamed. If the old key name is still used, it
+        # is still regarded.
+        for key in ("cookie_duration", "num_proxies_used"):
+            if key in data:
```
veluca93, line 292: We should probably print a deprecation warning here.

cms/conf.py
```diff
@@ -215,69 +227,158 @@ def _load(self, paths):
...
+                continue
+            # Warn about unknown keys.
+            if not hasattr(self, key):
+                logger.warning("Key %s unknown.", key)
             setattr(self, key, value)
 
         return True
 
+    def _suggest_updated_legacy_config(self, path, legacy_data):
+        logger.error("Legacy json config file found at %s. "
```
veluca93, line 319: Why is this an error and not a warning?

requirements.txt
```diff
@@ -17,6 +17,7 @@ chardet>=3.0,<3.1  # https://pypi.python.org/pypi/chardet
 babel>=2.6,<2.7  # http://babel.pocoo.org/en/latest/changelog.html
 pyxdg>=0.26,<0.27  # https://freedesktop.org/wiki/Software/pyxdg/
 Jinja2>=2.10,<2.11  # http://jinja.pocoo.org/docs/latest/changelog/
+tomli
```
veluca93, line 20: This should probably come with a version.

wil93: This can now be removed as we upgraded to Python 3.12 which (since 3.11) supports TOML natively: https://docs.python.org/3/library/tomllib.html

PR #1228, opened 2023-02-21
add docker compose for dev and prod env

peterxcli:
I wonder why this repo has no docker-compose for production(to run the admin-webServer just after you up a docker-compose) is there any issue cause there is no container for deployment config file

So I have added two docker compose file for development and production respectively.
I'm seeking your opinion for this commit, thanks

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -1,0 +1,56 @@
+# syntax=docker/dockerfile:1
+FROM ubuntu:20.04
+
+RUN apt-get update
+RUN apt-get install -y \
+    build-essential \
+    cgroup-lite \
+    cppreference-doc-en-html \
+    fp-compiler \
+    git \
+    haskell-platform \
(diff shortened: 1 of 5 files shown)


wil93:
Hello, and sorry for the delay. Thanks for the PR 😄 

The `docker-compose.test.yml` file was added in #1221, where I mentioned that it would make sense to make a development version of it, but maybe not a production version just yet, since each situation can be slightly different I guess (contest admins can write a yml file themselves, and reuse just the Dockerfile).

As for having separate Dockerfiles, I think it would be best to try as much as we can to always keep only one image that supports all use cases (testing, development, production).

In summary, I would not add a prod yml file yet, and for a dev yml file we should understand what we need for local development. For example storing the DB in /tmp is probably not a good idea for local development, it would be better to store it in a temporary folder within the repo so that the data is preserved after reboot.

PR #1232, opened 2023-04-08
Adding C++17 to the default list of allowed programming language

davinci-tech:
C++17 offers a variety of great features which would benefit competitive programmers in the long run (such as Structured bindings, std::optional, std::clamp, std::gcd and std::lcm among other). Unfortunately, the default list of allowed programming languages to be used in a contest does not contain the option to use C++17. This commit will simply add the C++17 option.

### DIFF
--- a/cms/db/contest.py
+++ b/cms/db/contest.py
@@ -79,7 +79,7 @@ class Contest(Base):
     languages = Column(
         ARRAY(String),
         nullable=False,
-        default=["C11 / gcc", "C++11 / g++", "Pascal / fpc"])
+        default=["C11 / gcc", "C++11 / g++", "Pascal / fpc", "C++17 / g++"])
 
     # Whether contestants allowed to download their submissions.
     submissions_download_allowed = Column(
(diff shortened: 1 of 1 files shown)


cms/db/contest.py
```diff
@@ -79,7 +79,7 @@ class Contest(Base):
     languages = Column(
         ARRAY(String),
         nullable=False,
-        default=["C11 / gcc", "C++11 / g++", "Pascal / fpc"])
+        default=["C11 / gcc", "C++11 / g++", "Pascal / fpc", "C++17 / g++"])
```
wil93, line 82: Maybe it would make more sense to replace C++11 with C++17, what do you think?

It might make sense to allow both in some contests, but a more sensible default probably is to allow only one.

wil93 (changes requested):

cms/db/contest.py
davinci-tech: I think you are right. Especially since some contestants do not know the difference between the c++11 and the c++17 standards and will usually choose the later one.

davinci-tech: I changed the default setting for g++ from C++11 to C++17.

PR #1233, opened 2023-05-02
Simplify testing with pytest

wil93:
In this PR we:

- get rid of `cmsRunUnitTests.py` entirely, replacing it with `pytest` which has a powerful auto-discovery feature (see `pytest.ini` file), is compatible with the `unittest`-style tests we're already using, and produces a far more readable output which makes it easy to see what's failing
- fix a bunch of errors that came up while testing with `pytest`, it seems that our unittest runner was missing some tests because **not all of them are executable** (should we `chmod +x` them? I'm almost tempted to `chmod -x` them, considering that pytest doesn't have this requirement)
- fix a bunch of warnings of deprecated methods (e.g. `assertEquals` -> `assertEqual`) which I saw in the output thanks to `pytest` highlighting them (there are still a few more warnings but can't be fixed as easily yet)
- split `docker build` and `docker run` in two separate workflow steps (to reduce the output size in the github action and make it more readable)
- improve the `docker (re)build` speed by better using the cache (we first copy the `requirements.txt`, then install, then copy the rest of the folder, which means we won't invalidate the cached `pip install` step unless we really need to)


cms/io/web_service.py
```diff
@@ -27,7 +27,8 @@
     import tornado.wsgi as tornado_wsgi
 from gevent.pywsgi import WSGIServer
 from werkzeug.contrib.fixers import ProxyFix
-from werkzeug.wsgi import DispatcherMiddleware, SharedDataMiddleware
+from werkzeug.middleware.dispatcher import DispatcherMiddleware
+from werkzeug.middleware.shared_data import SharedDataMiddleware
```
stefano-maggiolo, line 31: Unrelated to this commit?

cmstestsuite/RunTests.py
```diff
@@ -22,34 +22,9 @@
...
-
-
 def main():
-    test_suite = get_test_suite()
     try:
-        if test_suite == UNITTESTS or len(test_suite) == 0:
-            sh(["./cmstestsuite/RunUnitTests.py"] + sys.argv[1:])
-        if test_suite == FUNCTIONALTESTS or len(test_suite) == 0:
-            sh(["./cmstestsuite/RunFunctionalTests.py"] + sys.argv[1:])
+        sh(["./cmstestsuite/RunFunctionalTests.py"] + sys.argv[1:])
```
stefano-maggiolo, line 27: If all RunTests does now is to call RunFunctionalTests, maybe we should get rid of it entirely, and move the only tiny additional functionality (dumping logs) into the other file?

cmstestsuite/unit_tests/db/filecacher_test.py
```diff
@@ -361,18 +364,15 @@ class TestFileCacherDB(TestFileCacherBase, DatabaseMixin, unittest.TestCase):
...
 
 
 class TestFileCacherFS(TestFileCacherBase, unittest.TestCase):
     """Tests for the FileCacher service with a filesystem backend."""
 
     def setUp(self):
-        super().setUp()
-        file_cacher = FileCacher(path="fs-storage")
-        self._setUp(file_cacher)
+        super().setUp(FileCacher(path="fs-storage"))
```
stefano-maggiolo, line 375: It's a bit suspicious codecov says this line (and other similar ones) are not covered by tests.

cmstestsuite/unit_tests/grading/steps/utils_test.py
```diff
@@ -122,7 +122,7 @@ def test_multiple_commands_success(self):
                              collect_output=True)
 
         # 2 commands executed, with exec_num 0 and 1
-        self.assertEquals(self.sandbox.exec_num, 1)
+        self.assertEqual(self.sandbox.exec_num, 1)
```
stefano-maggiolo, line 125: Optional: consider moving the normalization of assertEquals to a different commit.

cmstestsuite/unit_tests/server/file_middleware_test.py
```diff
@@ -83,7 +83,7 @@ def test_success(self):
             "attachment; filename=%s" % quote_header_value(self.filename))
         self.assertTupleEqual(response.get_etag(), (self.digest, False))
         self.assertEqual(response.accept_ranges, "bytes")
-        self.assertGreater(response.cache_control.max_age, 0)
+        # self.assertGreater(response.cache_control.max_age, 0)  # It seems that "max_age" is None
```
stefano-maggiolo, line 86: Out of curiosity was this already failing? Or what changed to make it fail?

stefano-maggiolo (commented):
Please make sure you copy the PR comments in the commit, it is useful to have them there.

Is there a place where we explain how to run tests? If not, consider creating it.

cmstestsuite/RunTests.py
wil93: Done

cms/io/web_service.py
wil93: Moved to `821e1b2b474d885f910f519101d06f6b97133576`

cms/server/contest/server.py
```diff
@@ -39,7 +39,7 @@
 
 import logging
 
-from werkzeug.wsgi import SharedDataMiddleware
+from werkzeug.middleware.shared_data import SharedDataMiddleware
```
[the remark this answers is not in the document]
wil93: Moved to `821e1b2b474d885f910f519101d06f6b97133576`

cmstestsuite/unit_tests/grading/steps/utils_test.py
wil93: Moved to `821e1b2b474d885f910f519101d06f6b97133576`

cmstestsuite/unit_tests/server/file_middleware_test.py
wil93: I think this might have changed with a werkzeug upgrade, so it was already failing for some time but the test was not being executed so we didn't notice.

cmstestsuite/unit_tests/db/filecacher_test.py
wil93: Indeed. I'm looking into it.

wil93 (commented):
I split the commits in a more sensible way (BTW: I was anyway planning to squash-merge at the end, and put all the comments in a single commit message, are you OK with it?)

I just added some documentation on how to run tests: I created a "Docker image" section which I plan to expand as soon as we will have a `docker-compose.dev.yml` file for the development use case.

cmstestsuite/RunTests.py
wil93: Done

cmstestsuite/unit_tests/db/filecacher_test.py
wil93: I think this was caused by `__test__ = False` on the parent class (TestFileCacherBase) being inherited by TestFileCacherDB and TestFileCacherFS. Fixing this made me realize that `super(ParentClass).setUp()` doesn't work like I though it did 😅  so now I'm using `ParentClass.setUp(self)` which although ugly seems to work.

stefano-maggiolo (approved):
If you can take a look whether changing the super() stuff works easily, please change it here, otherwise let's keep what you have done.

It's ok to submit all in once, but consider splitting the assertEquals change.

Issue #1216, opened 2022-08-21
CMSEvaluationService gets overloaded with too many workers


akmohtashami:
**Description:** When there are many workers, ES gets overloaded and can not assign a new task fast enough. This reduces the efficiency so much that having more workers after a certain point hurts performance. This happens because ES will be busy saving the results of the tasks and won't be available to assign new tasks.

**Steps to reproduce:**
1. Add ~80 workers
2. Make many submissions (~25x number of workers)
3. Rejudge all submissions to fill the queue (best to recompile so all tasks will take roughly the same amount of time)

**Expected:** A worker should be assigned a new task immediately after it finishes one.

**Actual:** Workers will end up waiting for a noticeable amount of time for the next task. This can be seen from admin page. One second all workers will be busy and the next all of them will be idle while the queue is still full. 

**Describe the solution you'd like**
In IOI 2017, we used a separate service to maintain the queue. The queue remains responsible for assigning new tasks to the workers and getting the results from them. However, when it receives task results, it sends them separately to ES to save them on the database. This allows the worker to be reused immediately and removes any bottleneck on the queue. Also,  in this way, ES can be scaled which is currently not possible since it is also handling the queue.   I suggest adapting the code from 2017.

**Describe alternatives you've considered**

* One alternative is that we will have workers connected to different ESes. The ES would be the bridge between the worker and the queue. It would obtain the results, save it to the database and then ask for a new job from the queue for the worker. However, in this case, workers for an ES would still be idle while it is saving results. While each ES gets less workers, still database can be stressed and so the save operation can be slower which makes the idleness longer. Also we would need to handle what would happen if one ES goes down (Workers would need to connect to a different ES, etc.). The approach is also different from the current approach as it relies on ES asking from the queue for new jobs. Right now, the queue assigns the jobs to idle workers. 
* Workers can be responsible for writing the results directly to the database. But this might be undesirable since it would require workers to have direct access to the database. 
* ES can prioritize assigning tasks over saving the results. But this would still not allow it to scale which I think is a very important problem. Also, when it is in the middle of saving a task it can not abandon it (though probably not that important).
* We can also have several queues and put tasks in one of them when enqueuing (for example in a round-robin fashion). However, this means that if workers for one of the ESes die for some reason, the tasks will be stuck (or some tasks might get done slower). Also, it would make recovering from failure of one ES harder (now we can just add all the remaining tasks to the queue but then we would have to check what is in other ESes, etc.)

**Additional context**
[...]

PR #1236, opened 2023-07-09
[WIP] Split Queue handling and DB writing to address #1216

akmohtashami:
To address #1216, this PR splits the ES into two separate services:

1- QueueService: This service is solely responsible for maintaining the queue. It assigns tasks to the worker. Once the results are ready it receives them and assigns it to an ES to write to database. Once done, QueueService asks ES to also create next step operations and push them to the queue. In this way, there is no time consuming process being done in the QueueService itself which allows it to always be available. The only exception currently is finding missing operations and invalidating a submission. But these happen less regularly. 

2- EvaluationService: This service handles the time consuming tasks needed for maintaining the queue (e.g. writing to the database and computing new operations). The current implementation also allows for multiple ES to be running. I am considering renaming this to QueueWorkerService.

The PR is currently WIP because there are some tasks left:
- [ ] Making sure dequeuing an operation fails when it is being written by ES (there is no way to cancel this so we can only fail)
- [ ] Some final clean ups

But I wanted to see if there are any comments in advance before finalizing.


wil93:
> I am considering renaming this to QueueWorkerService.

Maybe even just QueueWorker? 🙂

PR #1238, opened 2023-07-29
Update setup.py

Apetree100122:
I think this might be good can you assign me to bugs or whichever you deem proper


wil93:
Thanks for the contribution, can you explain what changed? It looks like the file was only reformatted (and is now much less readable)

PR #1239, opened 2023-07-29
Update dev-requirements.txt

Apetree100122:
See if works or if needed send review


Apetree100122:
Let Mr me know what needs change

wil93:
Again this is just reformatting the file in a less readable way. Please do not send formatting only pull requests, it's better to make a useful change and in the same PR fix some formatting issues.

PR #1240, opened 2023-09-11
AWS: do not depend on ordering of `UNION ALL`

pobrn:
Apparently neither postgres nor the SQL standard guarantees that the order of the results will match the order the of the queries<sup>[0]</sup>.

If the results are not returned in order, then AWS will confuse which result belongs to which query, producing incomprehensible results on the admin page.

[0]: https://dba.stackexchange.com/questions/316818/are-results-from-union-all-clauses-always-appended-in-order

### DIFF
--- a/cms/server/admin/server.py
+++ b/cms/server/admin/server.py
@@ -27,7 +27,7 @@
 
 import logging
 
-from sqlalchemy import func, not_
+from sqlalchemy import func, not_, literal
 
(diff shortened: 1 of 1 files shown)


wil93:
Hi @pobrn, thanks for the PR. I wonder if this is similar to what was implemented in #1201?

wil93:
Thank you for the ping. Merged #1201 with a small change (to replace the deprecated add_column)

PR #1244, opened 2023-10-29
ProxyService bugfix

CawaAlreadyTaken:


cms/service/ProxyService.py
```diff
@@ -562,28 +562,31 @@ def dataset_updated(self, task_id):
...
-
-            for submission in task.submissions:
-                # Update RWS.
-                if not submission.participation.hidden and \
-                        submission.official and \
-                        submission.get_result() is not None and \
-                        submission.get_result().scored():
-                    for operation in self.operations_for_score(submission):
-                        self.enqueue(operation)
+            if task is not None:
```
wil93, line 565: It might be more readable to invert the check here (`if task is None`), returning early in that case. It would avoid having to indent the rest of the code.

cms/service/ProxyService.py
```diff
@@ -562,28 +562,31 @@ def dataset_updated(self, task_id):
...
+                for submission in task.submissions:
+                    # Update RWS.
+                    if not submission.participation.hidden and \
+                            submission.official and \
+                            submission.get_result() is not None and \
+                            submission.get_result().scored():
+                        for operation in self.operations_for_score(submission):
+                            self.enqueue(operation)
+            else:
+                logger.warning("Dataset update for unexistent task %d.", task_id)
```
wil93, line 592: nit: `unexistent` -> `non-existent`

wil93 (changes requested):
Thanks!

I left a couple comments. I wonder also: how to reproduce this bug?

cms/service/ProxyService.py
CawaAlreadyTaken: You are right it would make it more readable. I think I fixed it and pushed, I am not sure it's the first time I open a pull request, sorry for any github-wise misbehaviour :D

cms/service/ProxyService.py
CawaAlreadyTaken: I just corrected each occurrence of it in the file

wil93:
Can you explain how this bug gets triggered? It could be useful to know it so we can write a test for it.

CawaAlreadyTaken:
While a contest is running, if the dataset for a task that does not belong to the running contest gets updated, the error is triggered.
Se system keeps working, but the python Traceback for the Exception is written in the logs, visible from the admin Dashboard.
I think `task.contest_id` should be used instead of `task.contest.id`.

PR #1250, opened 2023-11-20
fix missing spaces between sentences in task_submissions.html

prandla:
Jinja removes whitespace between template tags by default. This causes some sentences on the task submissions page to not have a space between them. Fixed by using `{%+` instead of `{%` to preserve some of the whitespace, as is done in some other places in CWS.


wil93:
Thanks! Do you by any chance have a screenshot to showcase the issue?

Issue #639, opened 2016-08-15
Forcing ScoreTypeGroup


qnighy:
Now all scoretypes except `Sum` are `ScoreTypeGroup`. Moreover, all non-group scoretypes can be regarded as scoretypes with one group.

Regarding the rule of IOI 2016, I think we will have to write more subtask-oriented code. Forcing all scoretypes to be `ScoreTypeGroup` may help this. For example, moving `Sum` to `GroupSum` like in #620 makes `Sum` unnecessary and we can concentrate on `ScoreTypeGroup`.

Moreover we may introduce different scoretypes for subtasks in a task, if it is useful... (Actually Japanese OI committee have made one or two such tasks, if I remember correctly.)

PR #1251, opened 2023-11-20
Add scoring types GroupSum, GroupSumCond, GroupSumCheck

prandla:
These are 3 scoring types that are frequently used in Estonian olympiads, especially for more beginner-oriented tasks.

GroupSum gives points for each test proportionally for each subgroup. It doesn't have many functional advantages over Sum (or even just GroupMin and splitting each testcase into its own group), but it makes the contestant-facing UI consistent with GroupMin tasks (we generally have a few GroupSum tasks and then GroupMin for the rest). Also, it allows assigning different weights to different testcases in a somewhat intuitive manner (i.e. you give it the amount of points for the entire group, and the points get distributed evenly between the testcases).

GroupSumCond is useful for tasks where the question is of the form "find X, or if no such exist, print NONE". In these tasks, with GroupSum it'd be easy to get a lot of points by just printing NONE unconditionally. GroupSumCond allows designating some groups as "conditional", with the contestant only getting points for such groups if they also got some points from the unconditional groups. This allows moving all testcases where the answer is NONE into a conditional group, with the contestant only getting points if they also solve at least one of the non-NONE test cases correctly.

GroupSumCheck allows setting the whole group's points to zero in case of some particularly egregious error in one testcase. Admittedly less useful than the other 2 (I think we only needed it for BOI once).

I am aware that this PR is missing tests and documentation, but I'd want to first get feedback as to whether this feature is desired for CMS master.

### DIFF
--- a/cms/grading/scoretypes/GroupSum.py
+++ b/cms/grading/scoretypes/GroupSum.py
@@ -1,0 +1,48 @@
+#!/usr/bin/env python3
+
+# Contest Management System - http://cms-dev.github.io/
+# Copyright © 2010-2012 Giovanni Mascellani <mascellani@poisson.phc.unipi.it>
+# Copyright © 2010-2018 Stefano Maggiolo <s.maggiolo@gmail.com>
+# Copyright © 2010-2012 Matteo Boscariol <boscarim@hotmail.com>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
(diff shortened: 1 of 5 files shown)


wil93:
Related prior discussions: #639 #620

prandla:
regarding lack of demand, we use GroupSum at least a few times per year, and it really isn't that big of a maintenance burden IMO (well, GroupSumCheck might be a bit of a different story as far as demand goes, but it too is tiny code-wise). issue 639 is interesting, i assume it would allow some refactoring and simplification of the scoring logic, but i don't know the codebase well enough to say how useful that would be.

wil93:
I think regardless of whether these scoretypes are useful enough to ship by default, it would make sense to have a separate PR for each of them. I also think GroupSum probably now has enough demand, so it can be added (with tests).

@stefano-maggiolo what do you think about adding GroupSum now, and specifically about the idea in #639 to basically drop the standard Sum in favor of GroupSum?

stefano-maggiolo:
GroupSum: seems uncontroversial.
GroupSumCheck: can't this be implemented with GroupThreshold? Actually I think it would already work (as long as you set a very high threshold) as GT checks outcome > 0
GroupSomeCond: this seems quite involved to me (hence not generic enough to be in the repo); the idea to deal with those tasks was to use GroupMin/Mul to make sure that each subtask has a representation of None and non-None answers.

Regarding #639, the only objection I have is whether we can keep a simple contestant interface for "tasks that are effectively Sum tasks" even if we represent as GroupSum.

PR #1253, opened 2024-01-17
Make the yaml importer load multiple languages.

veluca93:


cmscontrib/loaders/italy_yaml.py
```diff
@@ -371,12 +371,28 @@ def get_task(self, get_statement=True):
...
-                primary_language: Statement(primary_language, digest)
+                primary_language: Statement(
+                    primary_language, digest)
             }
-
             args["primary_statements"] = [primary_language]
 
+            for (lang, lang_code) in LANGUAGE_MAP.items():
+                if lang_code == primary_language:
+                    continue
```
wil93, line 383: IIUC this would mean that when no `primary_language` is specified in the `task.yaml` and the following files are present in the statement folder:

- english.pdf
- french.pdf
- italian.pdf
- spanish.pdf

We would actually only import 3 files (skipping italian.pdf). Is it correct?

veluca93: Feel free to make that change.

veluca93: Modified the code so that if there is a statement in the primary language, importing succeeds even if `statement.pdf` is not present.

wil93 (approved):

PR #1255, opened 2024-02-08
Upgrade gevent and greenlet

wil93:
Updating these two to the oldest versions that still successfully build on a Mac M1, mostly just for ease of testing, so I don't have to spin up a linux VM every time. (I still have issues with `fp-compiler` but that isn't critical).

### DIFF
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -1,4 +1,3 @@
-name: cms
 version: "3.3"
 
 services:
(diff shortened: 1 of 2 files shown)


stefano-maggiolo:
I vaguely remember some reason to keep gevent pinned to 1.5, but tbh I can't remember which. If tests are passing I guess that's fine

PR #1271, opened 2024-11-16
Compatibility with Python 3.12.

veluca93:
Inspired by https://github.com/ioi-2024/cms/commit/0f044fd92c3d0b79e1588df015d6b0fdca930dee

### DIFF
--- a/cmscontrib/loaders/polygon.py
+++ b/cmscontrib/loaders/polygon.py
@@ -20,7 +20,6 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-import imp
 import logging
 import os
 import subprocess
@@ -145,10 +144,11 @@ class PolygonTaskLoader(TaskLoader):
         task_cms_conf = None
         if os.path.exists(task_cms_conf_path):
             logger.info("Found additional CMS options for task %s.", name)
-            with open(task_cms_conf_path, 'rb') as f:
-                task_cms_conf = imp.load_module('cms_conf', f,
-                                                task_cms_conf_path,
-                                                ('.py', 'r', imp.PY_SOURCE))
+            import importlib.util
+            spec = importlib.util.spec_from_file_location(
+                'cms_conf', task_cms_conf_path)
+            task_cms_conf = importlib.util.module_from_spec(spec)
+            spec.loader.exec_module(task_cms_conf)
         if task_cms_conf is not None and hasattr(task_cms_conf, "general"):
             args.update(task_cms_conf.general)
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,20 +1,18 @@
-# We are currently targeting the versions shipped in Ubuntu 20.04.
-
-tornado>=4.5,<4.6  # http://www.tornadoweb.org/en/stable/releases.html
-psycopg2>=2.8,<2.9  # http://initd.org/psycopg/articles/tag/release/
+tornado==4.5.3  # http://www.tornadoweb.org/en/stable/releases.html
+psycopg2==2.9.7  # http://initd.org/psycopg/articles/tag/release/
 sqlalchemy>=1.3,<1.4  # http://docs.sqlalchemy.org/en/latest/changelog/index.html
 netifaces>=0.10,<0.11  # https://bitbucket.org/al45tair/netifaces/src/
-pycryptodomex>=3.6,<3.7  # https://github.com/Legrandin/pycryptodome/blob/master/Changelog.rst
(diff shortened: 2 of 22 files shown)


cmscontrib/loaders/polygon.py
```diff
@@ -145,10 +144,11 @@ def get_task(self, get_statement=True):
...
             logger.info("Found additional CMS options for task %s.", name)
-            with open(task_cms_conf_path, 'rb') as f:
-                task_cms_conf = imp.load_module('cms_conf', f,
-                                                task_cms_conf_path,
-                                                ('.py', 'r', imp.PY_SOURCE))
+            import importlib.util
+            spec = importlib.util.spec_from_file_location(
+                'cms_conf', task_cms_conf_path)
+            task_cms_conf = importlib.util.module_from_spec(spec)
+            spec.loader.exec_module(task_cms_conf)
```
wil93, line 151: Have we tried running this on a polygon task by any chance? (TODO for the future: add a test with a polygon task)

requirements.txt
```diff
@@ -1,20 +1,19 @@
...
 sqlalchemy>=1.3,<1.4  # http://docs.sqlalchemy.org/en/latest/changelog/index.html
 netifaces>=0.10,<0.11  # https://bitbucket.org/al45tair/netifaces/src/
-pycryptodomex>=3.6,<3.7  # https://github.com/Legrandin/pycryptodome/blob/master/Changelog.rst
+pycryptodomex==3.19.0  # https://github.com/Legrandin/pycryptodome/blob/master/Changelog.rst
 psutil>=5.5,<5.6  # https://github.com/giampaolo/psutil/blob/master/HISTORY.rst
-requests>=2.22,<2.23  # https://pypi.python.org/pypi/requests
-gevent==20.12.0  # http://www.gevent.org/changelog.html
-# Limit greenlet version for binary compatibility with gevent 20.12 wheels
-greenlet==1.0.0
-werkzeug>=0.16,<0.17  # https://github.com/pallets/werkzeug/blob/master/CHANGES
```
wil93, line 13: Let's keep the comment here?

requirements.txt
```diff
@@ -1,20 +1,19 @@
...
+requests==2.32.3  # https://pypi.python.org/pypi/requests
+gevent==23.9.0.post1  # http://www.gevent.org/changelog.html
+werkzeug<1.0
+backports.ssl-match-hostname==3.7.0.1
```
wil93, line 11: What is this? 🤔

requirements.txt
```diff
@@ -1,20 +1,19 @@
-# We are currently targeting the versions shipped in Ubuntu 20.04.
-
-tornado>=4.5,<4.6  # http://www.tornadoweb.org/en/stable/releases.html
-psycopg2>=2.8,<2.9  # http://initd.org/psycopg/articles/tag/release/
+# Fix tornado version to be sure that it will work with our patch
```
wil93, line 1: This comment seems out of place, should we remove it?

cmscontrib/loaders/polygon.py
veluca93: No, and I don't think it's *too* important as this is a corner-case even in the polygon loader. Anyway, this is documented as equivalent in the python 3.12 docs.

requirements.txt
veluca93: Let's try if it works without it...

veluca93: Tornado appears to import it but not depend on it (?)

cmscontrib/loaders/polygon.py
wil93: Indeed: https://docs.python.org/3.12/library/importlib.html#importing-a-source-file-directly

requirements.txt
```diff
@@ -1,20 +1,18 @@
...
-requests>=2.22,<2.23  # https://pypi.python.org/pypi/requests
-gevent==20.12.0  # http://www.gevent.org/changelog.html
-# Limit greenlet version for binary compatibility with gevent 20.12 wheels
-greenlet==1.0.0
-werkzeug>=0.16,<0.17  # https://github.com/pallets/werkzeug/blob/master/CHANGES
+requests==2.32.3  # https://pypi.python.org/pypi/requests
+gevent==23.9.0.post1  # http://www.gevent.org/changelog.html
+werkzeug<1.0 # https://github.com/pallets/werkzeug/blob/master/CHANGES
+backports.ssl-match-hostname==3.7.0.1 # required by tornado<5.0
+greenlet>=3.0rc1
```
wil93, line 11: nit: Let's keep the same order? i.e. keep greenlet and gevent together

wil93 (approved):

PR #1272, opened 2024-11-16
Remove python2.

veluca93:
It is not supported on the latest Ubuntu LTS, and has been unsupported for almost 5 years.


wil93 (commented):
Can we also remove `python2` from the Dockerfile?

wil93 (approved):

PR #1275, opened 2024-11-16
Support for c++20

Virv12:
This PR adds support for evaluating c++20 submissions. The default is also upgraded from c++17 to c++20.

### DIFF
--- a/cms/db/contest.py
+++ b/cms/db/contest.py
@@ -79,7 +79,7 @@ class Contest(Base):
     languages = Column(
         ARRAY(String),
         nullable=False,
-        default=["C11 / gcc", "C++17 / g++", "Pascal / fpc"])
+        default=["C11 / gcc", "C++20 / g++", "Pascal / fpc"])
 
     # Whether contestants allowed to download their submissions.
     submissions_download_allowed = Column(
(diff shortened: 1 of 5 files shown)


cms/db/contest.py
```diff
@@ -79,7 +79,7 @@ class Contest(Base):
     languages = Column(
         ARRAY(String),
         nullable=False,
-        default=["C11 / gcc", "C++17 / g++", "Pascal / fpc"])
```
wil93, line 82: I'm thinking this technically requires a schema updater, but to be fair we didn't write one for #1232 either, so... 🤷‍♂️

cms/grading/languages/cpp20_gpp.py
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env python3
+
+# Contest Management System - http://cms-dev.github.io/
+# Copyright © 2016 Stefano Maggiolo <s.maggiolo@gmail.com>
+# Copyright © 2020 Andrey Vihrov <andrey.vihrov@gmail.com>
```
wil93, line 5: Change these lines to your name?

wil93 (approved):

PR #1280, opened 2024-11-17
Test more language variants in communication

veluca93:


cmstestsuite/Tests.py
```diff
@@ -251,54 +254,52 @@
...
-                    LANG_C_SHARP),
+         languages=COMMUNICATION_LANGUAGES + (LANG_C_SHARP,),
          checks=[CheckOverallScore(0, 100)]),
 
     # Communication tasks. PHP is not yet supported.
 
     Test('communication-fifoio-correct',
          task=communication_fifoio_stubbed,
          filenames=['communication-stubbed-correct.%l'],
-         languages=(LANG_C, LANG_CPP, LANG_PASCAL, LANG_PYTHON3, LANG_JAVA),
```
wil93, line 271: Not having C# here is on purpose or maybe was it forgotten at the time?

wil93: If it's not too tedious it would probably be good to implement `communication` solutions for all languages.

wil93 (commented):
It would be nice to re-order the lists of languages so they're sorted alphabetically. Also for "COMPILED_LANGUAGES" maybe there's a better name, since right now Python is in it too?

Lastly, maybe we can avoid having to add new languages in multiple lists by specifying languages to remove instead of to add, something like:

```
LIST_OF_LANGUAGES = removeFromList(ALL_LANGUAGES, [LANG_PYTHON3])
```

veluca93:
> It would be nice to re-order the lists of languages so they're sorted alphabetically. Also for "COMPILED_LANGUAGES" maybe there's a better name, since right now Python is in it too?
> 
> Lastly, maybe we can avoid having to add new languages in multiple lists by specifying languages to remove instead of to add, something like:
> 
> ```
> LIST_OF_LANGUAGES = removeFromList(ALL_LANGUAGES, [LANG_PYTHON3])
> ```

In CMS, Python *is* a compiled language ;) (it has a separate compilation step)

cmstestsuite/Tests.py
veluca93: agreed, left a todo for it.

cmstestsuite/Tests.py
```diff
@@ -53,18 +53,32 @@
...
+    LANG_RUST,
 )
 
+# Languages which support compilation with a manager/grader.
+# TODO: this should be the same as `ALL_LANGUAGES`.
+MANAGER_LANGUAGES = [
+    lang for lang in ALL_LANGUAGES if lang not in (LANG_HS, LANG_PHP, LANG_RUST)
+]
+
+# Languages for which solutions to be linked to a stub for communication tasks is
```
wil93, line 75: nit: `should` be linked?

wil93 (approved):

cmstestsuite/Tests.py
```diff
@@ -239,66 +254,65 @@
 
     Test('oom-static',
          task=batch_stdio, filenames=['oom-static.%l'],
-         languages=NON_INTERPRETED_LANGUAGES,
+         languages=(LANG_C, LANG_CPP, LANG_CPP14,
```
wil93, line 257: If we add LANG_CPP23 in the future, how do we ensure we'll remember to put it here too?

veluca93: Honestly it doesn't really matter, having this test in a single language would probably be more than enough...

cmstestsuite/Tests.py
```diff
@@ -172,7 +186,8 @@
 
     Test('compile-fail',
          task=batch_fileio, filenames=['compile-fail.%l'],
-         languages=COMPILED_LANGUAGES,
+         # PHP's compilation step cannot fail.
```
wil93, line 189: I would elaborate here and say: PHP's compilation step cannot fail as it's just a `cp` command.

cmstestsuite/Tests.py
veluca93: No, this is correct (except I should have written `are` instead of `is`, fixed)

cmstestsuite/Tests.py
wil93: That can be said for many of these tests (after all we're really testing the behavior of the sandbox). Still it makes sense to keep them uniform: maybe we add a new language where it's quite natural to implement `oom-static.%l` and I think it makes sense to do it.

Personally I would go with the same exclusion approach you did for MANAGER and COMMUNICATION.

veluca93: I disagree, and I would rather modify this test to just test C than follow that approach.

It's also very much not true that we're just testing the behaviour of the sandbox in these tests.

veluca93: Said otherwise: *some* tests (such as this one) test the behaviour of the sandbox.
Other tests don't.

wil93: Yep that's what I said :)

Using only LANG_C isn't a bad idea, also for some other tests, but I guess for now we can keep it as it is.

I would still prefer to at least have this list defined above instead of inlined here if we keep it as it is, but it's not a strong opinion.

wil93 (approved):

PR #1300, opened 2024-12-05
Customizable documentation

Virv12:
Allows documentation for multiple languges, when docs_path is not set the legacy stl_path variable is used mantaining retrocompatibility.

### DIFF
--- a/cms/conf.py
+++ b/cms/conf.py
@@ -136,6 +136,7 @@ class Config:
         self.max_submission_length = 100_000  # 100 KB
         self.max_input_length = 5_000_000  # 5 MB
         self.stl_path = "/usr/share/cppreference/doc/html/"
+        self.docs_path = None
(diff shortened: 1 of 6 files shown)


cms/conf.py
```diff
@@ -136,6 +136,7 @@ def __init__(self):
         self.max_submission_length = 100_000  # 100 KB
         self.max_input_length = 5_000_000  # 5 MB
         self.stl_path = "/usr/share/cppreference/doc/html/"
```
wil93, line 138: I think we can remove this line

veluca93 (approved):

cms/conf.py
Virv12: This would break retrocompatibility.

wil93: Ah, got it

PR #1301, opened 2024-12-06
Mark installation through system packages as unsupported.

veluca93:


stefano-maggiolo:
Agree to simplify instructions

stefano-maggiolo (approved):

PR #1302, opened 2024-12-06
Add support for ephemeral services.

veluca93:
Ephemeral services are services that are not fixed in the configuration file, but dynamically added as they connect. This is especially useful in a setup in which cmsWorker/cmsContestWebServer are scaled dynamically, as one might do when configuring CMS for running on cloud services.

### DIFF
--- a/cms/conf.py
+++ b/cms/conf.py
@@ -22,11 +22,14 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import errno
+import ipaddress
 import json
 import logging
 import os
+import socket
 import sys
 from collections import namedtuple
+from contextlib import closing
 
 from .log import set_detailed_logs
 
@@ -44,6 +47,7 @@ class ServiceCoord(namedtuple("ServiceCoord", "name shard")):
     service (thus identifying it).
 
     """
+
     def __repr__(self):
         return "%s,%d" % (self.name, self.shard)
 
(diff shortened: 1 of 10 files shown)


cms/conf.py
```diff
@@ -274,6 +349,18 @@ def _load_unique(self, path):
...
+            if service_name.startswith("_"):
+                continue
+            service = data["ephemeral_services"][service_name]
+            self.async_config.ephemeral_services[service_name] = \
+                EphemeralServiceConfig(
+                    service["subnet"],
+                    service["min_port"],
+                    service["max_port"],
+            )
+        del data["ephemeral_services"]
```
Virv12, line 362: `ephemeral_services` could not exists.

Virv12 (changes requested):

wil93 (changes requested):
I suggest expanding the documentation to add some information about this new kind of service.

wil93:
I guess the updates to the cms.conf.sample file can also be considered part of the documentation, in a way. However I am still worried it might be unclear how to this new feature works... Some questions:

Is "not providing a shard number" the one and only way to opt in to ephemeral services? If I have both the static and the ephemeral configuration in place in cms.conf, what happens if I launch `cmsResourceService -a`? Also, can any service can be ephemeral, or is it just Worker and ContestWebServer? If it's any service, couldn't we simplify the definition to something like the following?

```
"ephemeral_services":
    {
            "subnet": "127.0.0.0/8",
            "min_port": 26000,
            "max_port": 26999
    },
```

For the contest_listen_address we now say "When using ephemeral services only the first address and port are used", what does this mean?

wil93:
The benefit of having a clean-cut "switch" between hardcoded and ephemeral services is that it's less logic to maintain (we don't need to even test the case where we run with a mix of hardcoded and ephemeral services, since that would not be possible) and it also allows us (if we so choose) to deprecate the hardcoded style later on, and only support a single way of declaring services.

Also, would something like this help us to implement ephemeral services with less code? https://github.com/python-zeroconf/python-zeroconf

gollux:
I would appreciate better documentation. In particular, it should define what an ephemeral service is (instead of just mentioning that it can be used in some cases) and what is the meaning of shard numbers for such services. Also, it should specify which services can be ephemeral (I guess it's workers and the CWS now).

PR #1303, opened 2024-12-06
Prometheus Exporter

eliasol:
Add a new command: cmsPrometheusExporter

This will start a new HTTP server that listens on a port (defaults to 8811) for requests to /metrics. At this endpoint many metrics about a contest (or all contests) are exposed for Prometheus to be collected. Those metrics consider both system values (queue status, connected workers, ...) and contest values (submissions, users, questions, ...)

This exporter works by executing simple queries to the database, and obtains the queue information asking Evaluation Service.

The metrics exposed may leak information about the contest, so it's important to secure this endpoint (for example using a reverse proxy, or binding localhost).

With this exporter it's possible to build interactive and real-time dashboard for monitoring the status of the contest.

### DIFF
--- a/config/cms.conf.sample
+++ b/config/cms.conf.sample
@@ -194,7 +194,7 @@
 
     "_help": "Listening HTTP address and port for the exporter. If exposed",
     "_help": "this may leak private information, make sure to secure this endpoint.",
-    "prometheus_listen_address": "0.0.0.0",
+    "prometheus_listen_address": "127.0.0.1",
     "prometheus_listen_port": 8811,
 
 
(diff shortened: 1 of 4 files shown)


config/cms.conf.sample
```diff
@@ -187,6 +188,13 @@
     "pdf_printing_allowed": false,
 
 
+    "_section": "PrometheusExporter",
+
+    "_help": "Listening HTTP address and port for the exporter. If exposed",
+    "_help": "this may leak private information, make sure to secure this endpoint.",
+    "prometheus_listen_address": "0.0.0.0",
```
veluca93, line 195: Can we default this to 127.0.0.1?

veluca93 (dismissed):

veluca93 (approved):

PR #1304, opened 2024-12-07
Have ranking listen on localhost by default

veluca93:


wil93 (commented):
I'm thinking it might be unclear for CMS users how to revert to the current behavior (listen on any address) if they see this pre-filled in the documentation. Maybe we could expand the comment above this field to specify leaving the value empty? Or maybe specifying "0.0.0.0" also works?

veluca93:
0.0.0.0 also works, and it's IMO exactly what one would expect, so I don't see the need to add extra documentation...

wil93 (approved):
Ah, good to know. I would still lean towards specifying something like "change it to 0.0.0.0 to listen on all addresses" for extra clarity.

PR #1305, opened 2024-12-08
Add a Telegram bot to send announcements and reply to questions.

veluca93:
This bot is inspired by #1198 and by the ioi-germany implementation in https://github.com/ioi-germany/cms/blob/main/cms/service/TelegramBotService.py. It is implemented as a separate service, and does not modify the database.

The contest admin must configure a chat ID in which the bot will send questions as they arrive; they will be kept up to date as they are replied to or ignored. Announcements are also sent to this channel.

Members of the channel can reply to questions, either with pre-baked answers (with the inline keyboard) or by replying to the messages (and specifying `/answer`); they can also send announcements.

The bot supports running for all the contests, in which case it only interacts with questions.

### DIFF
--- a/cmscontrib/TelegramBot.py
+++ b/cmscontrib/TelegramBot.py
@@ -1,0 +1,461 @@
+#!/usr/bin/env python3
+
+# Contest Management System - http://cms-dev.github.io/
+# Copyright © 2024 Luca Versari <veluca93@gmail.com>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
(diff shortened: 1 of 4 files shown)


Virv12:
A few notes:
- We should document why we are using yaml instead of json, it's not obvious.
- It would be useful to support multi-contest mode with `-c ALL`.
- Maybe we want to use `default_argument_parser`.

veluca93:
> A few notes:
> 
> * We should document why we are using yaml instead of json, it's not obvious.

Added a comment

> * It would be useful to support multi-contest mode with `-c ALL`.

Done. In that case, announcements are not supported.

> * Maybe we want to use `default_argument_parser`.

Mostly done: this uses `contest_id_from_args` and `ask_for_contest`, but not `default_argument_parser` as that assumes it is passed a Service.

wil93 (changes requested):
Let's also have the bot send the username of who asked the question, maybe we can replace the question ID (which doesn't sound very helpful, correct me if I'm wrong) with it.

veluca93:
> Let's also have the bot send the username of who asked the question, maybe we can replace the question ID (which doesn't sound very helpful, correct me if I'm wrong) with it.

Done.

wil93 (approved):
I would really like to add some test coverage for this in the future. As agreed offline this is a contrib script so for the time being we can keep it as it is.

prandla:
> > * We should document why we are using yaml instead of json, it's not obvious.
> 
> Added a comment

@veluca93 the comment you added reads `# We use yaml instead of json or similar because`. Do you remember how you intended to finish that sentence? :)

PR #1307, opened 2024-12-16
Update docs in preparation for release

wil93:


docs/Installation.rst
```diff
@@ -136,7 +137,12 @@ Installing CMS and its Python dependencies
 
 There are a number of ways to install CMS and its Python dependencies:
 
-Method 1: Global installation with pip
+Method 1: Using the Docker image
```
veluca93, line 140: I would not recommend the docker image as the first method.

I'd recommend putting virtual environment first, global installation second, and docker image third.

Once the docker image is a bit more complete (IIRC @Virv12 had some issues with it), we can move it further up.

wil93: As discussed offline, now we recommend venv method, followed by docker method

veluca93 (approved):

docs/Installation.rst
Virv12: Well...

- In the container there are multiple config files
- The database password is hardcoded
- You cannot change the config without rebuilding the image
- There is no entrypoint
- We are installing the dev-dependencies, but I don't know python enough to judge this.

veluca93 (approved):

docs/Installation.rst
wil93: Thanks @Virv12 indeed our Dockerfile seems to be currently focused on the local test / development use case, we should expand it in the future to better support passing custom configuration etc.

PR #1309, opened 2024-12-23
Pull new translations and add script to map files

wil93:
There might be some string deletions, this is probably because the source text was changed and thus OneSky refused to import those translations so they are lost now. We could fix this manually, but it's probably easier to just wait until translators will fix it on Onesky.


scripts/copy_translations.py
veluca93, line 1: I'm not sure the scripts folder is the best place for this - it seems to be for scripts that are meant to be used by whoever installs CMS.
I'd rather have it be top-level (where i.e. the test scripts are), or make a dev-scripts folder

wil93: Moved it to top-level

veluca93 (approved):

PR #1317, opened 2025-02-08
Always include max_score in submission data

hezicyan:
Even if a token is not played, the participant should be able to see the `max_score`.

This fixes a UI issue where, in certain cases (below), the "Score of tokened submissions" box displays the wrong color after a submission.

Bug Reproduction Steps:

1. Submit a correct solution and play a token.
2. Submit an incorrect solution without refreshing the page.
3. After evaluation and scoring, the "Score of tokened submissions" box should remain green, but it incorrectly turns yellow.

The root cause of this bug is that the frontend JavaScript receives `data["max_score"]` as `undefined`.


wil93 (approved):

wil93:
Thank you!

PR #1318, opened 2025-02-12
Fix correct orders of parameters in ES

zesty-fox:
I think it is obvious that don't need to explain.
When I use `AddSubmission` to add a submission manually, ES will find it by `_missing_operation`, and it reports error when trying to enqueue the operation.
ps: Anyone know why it is useless to call `maybe_send_notification` to call a rpc method? Like AddSubmission won't trigger ES's `new_evaluation()` method.

### DIFF
--- a/cms/service/EvaluationService.py
+++ b/cms/service/EvaluationService.py
@@ -356,14 +356,14 @@ class EvaluationService(TriggeredService):
         counter = 0
         with SessionGen() as session:
 
-            for operation, timestamp, priority in \
+            for operation, priority, timestamp in \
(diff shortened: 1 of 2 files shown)


wil93 (commented):
Thank you, can you also fix the doc-string here? https://github.com/cms-dev/cms/blob/a4f06f2e3d2f77c17c73920c58f7e3127a02ad5b/cms/service/esoperations.py#L293C1-L294C22

zesty-fox:
Sorry for I forgot that. Now it's fixed.

wil93 (approved):

PR #1324, opened 2025-03-05
Fix type error in StressTest

zesty-fox:
StressTest.py reported TypeError because float number can't be used in `range()`, due to `//` in Python doesn't always return a integer. Now it's fixed.

### DIFF
--- a/cmstestsuite/StressTest.py
+++ b/cmstestsuite/StressTest.py
@@ -179,7 +179,7 @@ class Actor(threading.Thread):
         SLEEP_PERIOD = 0.1
         time_to_wait = self.metrics['time_coeff'] * \
             random.expovariate(self.metrics['time_lambda'])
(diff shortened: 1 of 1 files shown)


wil93 (approved):

wil93:
Thank you!

It's interesting though, that probably this script was probably not used since the Python 2 -> 3 migration which happened many years ago.

Out of curiosity: did you manage to run the script correctly? Maybe it could make sense to run it periodically as a Github action in that case.

zesty-fox:
> Thank you!
> 
> It's interesting though, that probably this script was probably not used since the Python 2 -> 3 migration which happened many years ago.
> 
> Out of curiosity: did you manage to run the script correctly? Maybe it could make sense to run it periodically as a Github action in that case.

Yes. It works well after solving this problem. I have run it for a stress test on my local machine, and it helps a lot.

PR #1330, opened 2025-03-27
Add a new batch-and-output task type.

veluca93:
This task type allows mixing explicitly-uploaded outputs and batch-style outputs generated by a submitted source file.

Tested locally.

### DIFF
--- a/cms/grading/steps/trusted.py
+++ b/cms/grading/steps/trusted.py
@@ -182,7 +182,7 @@ def trusted_step(sandbox, commands):
 
 
 def checker_step(sandbox, checker_digest, input_digest, correct_output_digest,
-                 output_filename):
+                 output_filename, extra_args=None):
     """Run the explicit checker given by the admins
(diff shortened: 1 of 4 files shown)


prandla:
* needs to be rebased on new main.
* there's a lot of code duplication between this and Batch.py. maybe we should try to refactor the common parts out.
* we should add some tests for this, in the functional test suite maybe?

cms/server/contest/submission/workflow.py
```diff
@@ -179,6 +179,7 @@ def accept_submission(
                 sql_session, participation, task, language,
                 missing_codenames)
         else:
+            print("???")
```
wil93: Is this intended?

gollux:
I would add a comment to `Batch` that it is used as a base for `BatchAndOutput`. Otherwise OK.

gollux (approved):

PR #1335, opened 2025-04-27
Add a simple API for automated interaction with CMS.

veluca93:
This adds 4 API endpoints (not CSRF protected):
- A login endpoint that does not return HTML
- An endpoint for programmatic access to the list of tasks, their statements, and the submission format
- An endpoint for programmatic submission
- An endpoint to get the total number of submissions on a task.

Together with the existing submission details APIs, this should allow automatic interaction with CMS.

### DIFF
--- a/cms/server/contest/handlers/__init__.py
+++ b/cms/server/contest/handlers/__init__.py
@@ -23,9 +23,24 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-from .communication import \
-    CommunicationHandler, \
-    QuestionHandler
+from .taskusertest import \
+    UserTestInterfaceHandler, \
+    UserTestHandler, \
+    UserTestStatusHandler, \
+    UserTestDetailsHandler, \
+    UserTestIOHandler, \
+    UserTestFileHandler
+from .tasksubmission import \
+    SubmitHandler, \
+    TaskSubmissionsHandler, \
+    SubmissionStatusHandler, \
+    SubmissionDetailsHandler, \
+    SubmissionFileHandler, \
+    UseTokenHandler
+from .task import \
+    TaskDescriptionHandler, \
+    TaskStatementViewHandler, \
+    TaskAttachmentViewHandler
 from .main import \
     LoginHandler, \
     LogoutHandler, \
@@ -34,24 +49,14 @@ from .main import \
     NotificationsHandler, \
     PrintingHandler, \
     DocumentationHandler
-from .task import \
-    TaskDescriptionHandler, \
-    TaskStatementViewHandler, \
-    TaskAttachmentViewHandler
-from .tasksubmission import \
-    SubmitHandler, \
-    TaskSubmissionsHandler, \
-    SubmissionStatusHandler, \
-    SubmissionDetailsHandler, \
-    SubmissionFileHandler, \
-    UseTokenHandler
-from .taskusertest import \
-    UserTestInterfaceHandler, \
-    UserTestHandler, \
-    UserTestStatusHandler, \
-    UserTestDetailsHandler, \
-    UserTestIOHandler, \
-    UserTestFileHandler
+from .communication import \
+    CommunicationHandler, \
+    QuestionHandler
+from .api import \
+    ApiLoginHandler, \
+    ApiSubmissionListHandler, \
+    ApiSubmitHandler, \
+    ApiTaskListHandler
 
 
 HANDLERS = [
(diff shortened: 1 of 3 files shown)


cms/server/contest/handlers/contest.py
```diff
@@ -256,6 +258,20 @@ def get_submission(self, task, submission_num):
...
+
+        task (Task): a task for the contest that is being served.
+
+        return (int): the number of submissions from this user on this task.
+
+        """
+        return self.sql_session.query(Submission) \
+            .filter(Submission.participation == self.current_user) \
+            .filter(Submission.task == task) \
+            .order_by(Submission.timestamp) \
```
gollux: Do we need to order the results to count them?

cms/server/contest/handlers/api.py
```diff
@@ -0,0 +1,145 @@
...
+        participation, cookie = validate_login(
+            self.sql_session, self.contest, self.timestamp, username, password,
+            ip_address)
+
+        cookie_name = self.contest.name + "_login"
+        if cookie is None:
+            self.clear_cookie(cookie_name)
+        else:
+            self.set_secure_cookie(cookie_name, cookie, expires_days=None)
+
```
gollux: I still don't think that using cookies for API authentication is reasonable, especially if their name varies with the contest. Authentication tokens should be as explicit as possible.

We should use `Authentication:` with a bearer token instead. The token can have the same format as the cookie. Login endpoint would return the token inside a JSON reply. We could modify `get_current_user` in `ContestHandler` to look for the authentication header and if there is one, use its value instead of the cookie.

This would be a very minimalistic change, easy to reason about.

cms/server/contest/handlers/__init__.py
```diff
@@ -97,6 +102,12 @@
     (r"/communication", CommunicationHandler),
     (r"/question", QuestionHandler),
 
+    # API
+    (r"/api/login", ApiLoginHandler),
+    (r"/api/task_list", ApiTaskListHandler),
+    (r"/api/(.*)/submit", ApiSubmitHandler),
+    (r"/api/(.*)/submission_list", ApiSubmissionListHandler),
```
gollux, line 109: I suspect that `(.*)` with eat even a slash. It would be better to use a narrower regex.

cms/server/contest/handlers/api.py
```diff
@@ -0,0 +1,147 @@
...
+    """Handler to list all tasks and their statements.
+
+    """
+    @tornado_web.authenticated
+    @actual_phase_required(0, 3)
+    @multi_contest
+    def get(self):
+        contest = self.contest
+        tasks = dict()
+        for task in contest.tasks:
```
gollux, line 85: The tasks should be presented in the same order as in the web UI. Let us sort by `task.num` here.

cms/server/contest/handlers/api.py
```diff
@@ -0,0 +1,147 @@
...
+        if task is None:
+            self.json({"error": "Not found"}, 404)
+            return
+        submissions: list[Submission] = (
+            self.sql_session.query(Submission)
+            .filter(Submission.participation == self.current_user)
+            .filter(Submission.task == task)
+            .all()
+        )
+        self.json({'list': [s.opaque_id for s in submissions]})
```
gollux: I would return an array of objects, even though the objects will contain just the ID. It would make it easy to add more attributes of submissions in the future.

gollux (changes requested):

cms/server/contest/handlers/api.py
gollux: In the future, I would like to support `Authentication`, but not necessarily in the initial version of the API.

cms/server/contest/handlers/__init__.py
veluca93: This matches the regex used everywhere else for task names, so I assume it should be fine

cms/server/contest/handlers/api.py
veluca93: They are already sorted by task.num in the sqlalchemy relationship.

cms/server/contest/handlers/api.py
veluca93: Done.

cms/server/contest/handlers/api.py
gollux: Ah, I didn't expect sorted relationships :) Good.

cms/server/contest/handlers/__init__.py
gollux: OK, we can keep the `(.*)` for now and fix it everywhere later.

cms/server/contest/handlers/api.py
prandla: the json might be emitted in the right order, but json itself doesn't guarantee ordering of keys in objects, so they might get shuffled by something else later. (e.g. it seems that javascript's JSON.stringify doesn't always preserve key order.)

veluca93: Good point, this should probably be a list.

PR #1343, opened 2025-05-28
add type hints

prandla:
Converting old docstring-based type hints to PEP 484 type hints. Current progress: all of cms/ except for cms/server/, and all of cmscommon/.

Notes:

* SQLAlchemy models are typed mostly correctly (required a few hacks, see diff of cms/db/base.py). SQLAlchemy queries are still untyped, requires upgrading to SQLAlchemy 2.0. It's also possible to manually type-hint the result of queries though.
* RPC method calls aren't type-checked. I couldn't figure out a way to do that without changing the method call syntax; I think it would be possible if RPC calls looked like `other_service.rpc.method(args)` (as opposed to the current `other_service.method(args)`).
* Minimum Python version is now 3.12 (for the new generic syntax). It would be pretty easy to port to python 3.11, but i don't want to go below that. I think this is not a huge issue, most distros have >=3.12 already, Debian is the last holdout and a new debian release should be happening in a few months.
* all of our jinja templates will forever be non-type-checked. this makes me sad :(


prandla:
> What feature from 3.12 do you need?

the new [type parameter syntax](https://docs.python.org/3/whatsnew/3.12.html#whatsnew312-pep695). i also wanted to use typing.override, but couldn't be bothered with it (there's a few functions where the return type depends on a boolean argument, using `override` allows typing them more precisely than a simple union).

it's okay, i'll just replace it with manual typing.TypeVar usage.

we might want to also run tests on the oldest version of python we want to support.

prandla:
Tests now pass on python 3.11 aswell. I'll make a separate PR later to automatically test this on github actions, for now i'm testing on 3.11 locally.

prandla:
some more notes:

for cmsranking and cmscontrib i tried to add type hints to all functions, even where they weren't present before. for cmstestsuite i only converted the old type hints to new-style ones.

there's a few modules where it would be possible to have much better static typing if the code was refactored a bit. currently i tried to keep the existing code untouched. (though i added a few asserts that should always be true, in cases where there's some complicated control flow and the type checker can't figure out that something is always not-None.)

the web server parts had less type hints than i expected; this is mostly because a lot of it is just passing parameters into jinja templates which are, again, completely untyped.

A lot of the remaining type errors are about sqlalchemy queries, which could be fixed by upgrading to sqlalchemy 2.0 (i think). I might try that at some point. The rest are in tornado/werkzeug/jinja, and as i understand, upgrading those is quite the ordeal. But I suppose it must still happen some day. And there's also a lot of type errors about missing None checks, most of them false positives, some of them quite certainly real :)

cms/db/__init__.py
```diff
@@ -45,6 +45,7 @@
     "version", "engine",
     # session
     "Session", "ScopedSession", "SessionGen", "custom_psycopg2_connection",
+    "Session",
```
veluca93: Was this intentional?

cms/db/__init__.py
```diff
@@ -88,7 +89,7 @@
 
 metadata = MetaData(engine)
 
-from .session import Session, ScopedSession, SessionGen, \
+from .session import Session, Session, ScopedSession, SessionGen, \
```
veluca93: Same

cms/grading/Sandbox.py
```diff
@@ -1451,7 +1488,13 @@ def cleanup(self, delete=False):
             rmtree(self._outer_dir)
 
 
-Sandbox = {
-    'stupid': StupidSandbox,
-    'isolate': IsolateSandbox,
+# StupidSandbox hasn't been tested in a while. It should probably be removed,
```
veluca93, line 1491: Can you open an issue for removing it?

cms/grading/scoretypes/abc.py
```diff
@@ -138,32 +148,34 @@ def get_html_details(self, score_details,
...
+            testcases; ranking headers.
 
         """
         pass
 
     @abstractmethod
-    def compute_score(self, unused_submission_result):
+    def compute_score(
+        self, submission_result: SubmissionResult
+    ) -> tuple[float, object, float, object, list[str]]:
```
veluca93, line 167: For later, probably we should make a dataclass for this...

cms/grading/steps/stats.py
```diff
@@ -26,22 +26,34 @@
...
 from cms.grading.Sandbox import Sandbox
+import typing
 
 
 # TODO: stats grew enough to justify having a proper object representing them.
 
 
-def execution_stats(sandbox, collect_output=False):
+# is this a proper enough object? :)
+class StatsDict(typing.TypedDict):
```
veluca93, line 36: why not a dataclass? :-)

cms/server/contest/submission/file_retrieval.py
```diff
@@ -96,18 +101,20 @@ def extract_files_from_archive(data):
...
+) -> list[ReceivedFile]:
     """Transform some files as received by Tornado into our format.
 
     Given the files as provided by Tornado on the HTTPServerRequest's
     files attribute, produce a result in our own format. Also, if the
     files look like they consist of just a compressed archive, extract
     it and return its contents instead.
 
-    tornado_files ({str: [tornado.httputil.HTTPFile]}): a bunch of
+    tornado_files: a bunch of
```
veluca93: single line? :-)

cms/service/LogService.py
```diff
@@ -78,8 +78,8 @@ def Log(self, **kwargs):
 
         Receive the attributes of a LogRecord, rebuild and handle it.
         The given keyword arguments will contain:
-        msg (string): the message to log.
-        levelname (string): the name of the level, one of "DEBUG",
+        msg (str): the message to log.
```
veluca93, line 81: Can we type this the same way as everything else?

cms/util.py
```diff
@@ -285,16 +297,15 @@ def _find_local_addresses():
...
 
-def _get_shard_from_addresses(service, addrs):
+def _get_shard_from_addresses(service: str, addrs: list[tuple[int, str]]) -> int | None:
     """Returns the first shard of a service that listens at one of the
     specified addresses.
 
-    service (string): the name of the service.
-    addrs ([(int, str)]): a list like the one returned by
+    service: the name of the service.
+    addrs: a list like the one returned by
```
veluca93: also probably should be single line

cms/service/LogService.py
prandla: these are given within kwargs, i'm not sure there's a nice way to annotate that. and having proper type hints wouldn't help as much either, the only caller is in cms/log.py which passes a dict of values into kwargs (and that's even ignoring the fact that this is an rpc call, so it doesn't get type-checked anyways), and the only thing that's done with them is passing them straight into logging.makeLogRecord.

cms/grading/steps/stats.py
prandla: I wanted to avoid changing existing code. it can be refactored along with some other dataclasses later.

cms/grading/scoretypes/abc.py
prandla: Indeed.

cms/db/__init__.py
prandla: Whoops. This used to be called SessionType, before i figured out i can do the trick with `if typing.TYPE_CHECKING`, apparently the rename didn't delete the now-duplicate imports. Will fix.

veluca93 (approved):

veluca93 (approved):

PR #1378, opened 2025-06-07
Add new restricted feedback mode

### LINKED ISSUE #1361: [IOI fork parity] better restricted feedback modes
https://github.com/ioi-2024/cms/commit/f52f91d187d9078cd6aef1222f40e88ad2cf62e5

changes restricted mode so that it shows until the first subtask with the lowest score, as opposed to showing until the first subtask with a non-full score. adds a new restricted mode that only shows the lowest subtask, instead of all subtasks up to it.

veluca93:
Closes #1361


gollux:
I would appreciate some documentation on what the restriction types mean.

veluca93:
> I would appreciate some documentation on what the restriction types mean.

Added documentation.

birka0:
I think this commit should be included as well:
https://github.com/ioi-2023/cms/pull/3/commits/709b37a18416644cd4e2396529810dc343dc74c2

It's probably not a realistic use case, but a bug is a bug -- sorry about that!

gollux (approved):

PR #1386, opened 2025-06-07
Use the filenames of the contestant as a default submission comment

### LINKED ISSUE #1371: [IOI fork parity] Use the filenames of the contestant as a default submission comment
https://github.com/ioi-2024/cms/commit/a4eec05049bdea7e3ff03c20d9574648bc351eb3

in my mind this is mostly useful for jury-submitted solutions, but i guess enabling it for all submissions isn't that bad.

i guess there could also be a secret input-box in CWS that allows setting the comment directly, i think that would also make the comment feature more ergonomic.

veluca93:
It is very useful for the scientific commitee to see it in the Comment column of the submissions page in the admin website. This column was mostly unused and empty in the past.

Closes #1371

### DIFF
--- a/cms/server/contest/submission/workflow.py
+++ b/cms/server/contest/submission/workflow.py
@@ -220,11 +220,15 @@ def accept_submission(
     logger.info("All files stored for submission sent by %s",
                 participation.user.username)
 
+    # Use the filenames of the contestant as a default submission comment
+    received_filenames_joined = ",".join([file.filename for file in received_files])
+
     submission = Submission(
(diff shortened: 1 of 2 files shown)


veluca93:
> This is technically another avenue for zipbombs to abuse (because this field shows all files that were in an archive), but given that we don't have any zip bomb protection at all currently, i think it's fine for now. I'll just have to handle this in the PR that fixes the other zip bombs.

IMO a zipbomb shouldn't even get to the point where this code gets called :-)

PR #1388, opened 2025-06-07
Increase file cache chunk size from 16 KB to 1 MB

### LINKED ISSUE #1368: [IOI fork parity] increase file cacher chunk size
https://github.com/ioi-2024/cms/commit/34f6079839cc4b10a3483ed90881775f06b2ea71

it would probably make the most sense to move this to a cms.conf setting.

or maybe fix whatever synchronization issues caused this to be a bottleneck in the first place :)

veluca93:
To improve performance when connecting to worker instances over high latency network

Fixes #1368

### DIFF
--- a/cms/db/filecacher.py
+++ b/cms/db/filecacher.py
@@ -478,8 +478,7 @@ class FileCacher:
     """
 
     # This value is very arbitrary, and in this case we want it to be a
-    # one-size-fits-all, since we use it for many conversions. It has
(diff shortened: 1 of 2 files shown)


veluca93:
I'm also fairly sure this will be fine at 1mb - and on low latency networks I imagine it won't cause problems.

On high latency networks, I think this was tested enough at IOI :D

prandla (approved):

PR #1391, opened 2025-06-07
feat: add config for removing submission rate limit

### LINKED ISSUE #1365: [IOI fork parity] add grace period near end of contest
https://github.com/ioi-2024/cms/commit/6bfb9ee7c68fca2ad0d09ac65e0870a343fff607
adds a grace period at the end of the contest where the usual submission rate limit is not applied.

veluca93:
Fixes #1365

### DIFF
--- a/cms/db/contest.py
+++ b/cms/db/contest.py
@@ -42,6 +42,7 @@ import typing
 if typing.TYPE_CHECKING:
     from . import Task, Participation
 
+
 class Contest(Base):
     """Class to store a contest (which is a single day of a
     programming competition).
@@ -252,6 +253,10 @@ class Contest(Base):
     min_submission_interval: timedelta | None = Column(
         Interval,
         CheckConstraint("min_submission_interval > '0 seconds'"),
+        nullable=True)
+    min_submission_interval_grace_period: timedelta | None = Column(
+        Interval,
+        CheckConstraint("min_submission_interval_grace_period > '0 seconds'"),
         nullable=True)
     min_user_test_interval: timedelta | None = Column(
         Interval,
--- a/cms/server/contest/submission/check.py
+++ b/cms/server/contest/submission/check.py
@@ -33,7 +33,6 @@ the current situation. Also include some support functions, which are
 exported as they may be of general interest.
 
 """
-
 from datetime import datetime, timedelta
 from sqlalchemy import desc, func
 from sqlalchemy.orm import Query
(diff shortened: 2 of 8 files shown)


cms/db/contest.py
```diff
@@ -247,6 +248,10 @@ class Contest(Base):
         Interval,
         CheckConstraint("min_submission_interval > '0 seconds'"),
         nullable=True)
+    min_submission_interval_grace_period: timedelta | None = Column(
```
prandla, line 257: why is this nullable with the check of >0, as opposed to not nullable, check >=0, and default 0? 0-second grace is logically equivalent to no grace, this might simplify some logic.

cms/server/contest/submission/check.py
```diff
@@ -33,6 +33,7 @@
 exported as they may be of general interest.
 
 """
+from datetime import datetime
```
prandla: redundant import

cms/server/contest/submission/check.py
```diff
@@ -207,3 +208,27 @@ def check_min_interval(
...
+    if participation.unrestricted \
+            or participation.contest.min_submission_interval_grace_period is None:
+        return False
+
+    if participation.contest.per_user_time is None:
+        end_time = participation.contest.stop
+    else:
+        end_time = participation.starting_time + participation.contest.per_user_time
+
+    end_time += participation.delay_time + participation.extra_time
```
prandla, line 228: is it good to duplicate this logic here? compute_actual_phase can already return the time the current phase will end.

cms/server/contest/submission/check.py
```diff
@@ -207,3 +208,27 @@ def check_min_interval(
...
             or timestamp - submission.timestamp >= min_interval)
+
+
+def is_last_minutes(timestamp: datetime, participation: Participation):
+    """
+    timestamp (datetime): the current timestamp.
+    participation (Participation): the participation to be checked.
+    delta (timedelta): length of the last time section to be checked.
+
+    return (bool): whether it is the last `delta` of the participation.
```
prandla: old-style type hints should be removed

prandla (changes requested):

cms/db/contest.py
veluca93: I am very reluctanct to change the logic here compared to what is known to match IOI rules.

We can perhaps consider some simplification as a follow-up. However, I would at that point reconsider the entire phase management of CMS.

cms/server/contest/submission/check.py
veluca93: See above

prandla (approved):

wil93:
This is also missing the updaters, can you please add them?

PR #1392, opened 2025-06-07
Allow contestants to view submissions (and possibly submit unofficially) after contest end

### LINKED ISSUE #1186: Let users see results (and possibly create unofficial submissions) after the end of the contest
**Is your feature request related to a problem? Please describe.**

There are two problems:

1. As soon as the contest ends, we are blocking access to the list and status of submissions. This means that users cannot see the result for submissions that are sent very close to the end of the contest. This in turn leads to a series of "clarification requests" where people are asking if the last submission they just sent will be counted or not.
2. As it sometimes happens, the server is overloaded near the end of the contest since many people try to submit. This can sometimes lead to not being able to submit a solution in time.

**Describe the solution you'd like**

We can solve both problems:

1. By continuing to show the list and status of submissions (i.e. in read-only mode) users won't need to ask if the submission is going to be counted or not: they can simply wait and refresh the page until they see the score.
2. By letting users create new "unofficial" submissions (i.e. in read-and-write mode) we can greatly mitigate the server load issue: as soon as the contest ends, people that were not able to submit can still do it unofficially and formally request to switch their last submission to "official", and this request can be later evaluated and approved/rejected by the admins.

**Describe alternatives you've considered**

The most glaring alternative would be to simply start the Analysis Mode as soon as the contest ends. This would indeed solve the 2 problems listed here, **however**, it would also introduce unwanted features:

- During analysis mode, CMS automatically shows all execution results for the testcases (instead of respecting the "restricted mode" where only the first failing testcases are shown). This means that submissions sent after the end of the contests have more feedback from the system.
- Also, if I remember correctly, there is a concept of "hidden attachment" in CMS where some attachments are visible only during analysis mode (e.g. a PDF with the explanation of the problem or a ZIP with all the testcases).

For these reasons it would be better to implement this as a standard behavior of CMS instead of having to rely on analysis mode.

**Additional context**

Possibly related to #61

### LINKED ISSUE #1363: [IOI fork parity] allow viewing submissions after end of contest
https://github.com/ioi-2024/cms/commit/e884a30a29c04ea24862525fbcafec3ebca2bb7c

### LINKED ISSUE #1364: [IOI fork parity] allow unofficial submissions after end of contest
https://github.com/ioi-2024/cms/commit/bba8607246cb7c8afa904ba3666e938c923f7df8

veluca93:
Fixes #1186
Fixes #1364
Fixes #1363

### DIFF
--- a/cms/db/contest.py
+++ b/cms/db/contest.py
@@ -101,6 +101,12 @@ class Contest(Base):
         nullable=False,
         default=True)
 
+    # Allow unofficial submission before analysis mode
+    allow_unofficial_submission_before_analysis_mode = Column(
+        Boolean,
+        nullable=False,
+        default=False)
+
     # Whether to prevent hidden participations to log in.
     block_hidden_participations: bool = Column(
         Boolean,
--- a/cms/server/admin/handlers/contest.py
+++ b/cms/server/admin/handlers/contest.py
@@ -96,6 +96,7 @@ class ContestHandler(SimpleContestHandler("contest.html")):
             self.get_bool(attrs, "submissions_download_allowed")
             self.get_bool(attrs, "allow_questions")
             self.get_bool(attrs, "allow_user_tests")
+            self.get_bool(attrs, "allow_unofficial_submission_before_analysis_mode")
             self.get_bool(attrs, "block_hidden_participations")
             self.get_bool(attrs, "allow_password_authentication")
             self.get_bool(attrs, "allow_registration")
--- a/cms/server/admin/templates/contest.html
+++ b/cms/server/admin/templates/contest.html
@@ -81,6 +81,15 @@
       </td>
       <td><input type="text" name="score_precision" value="{{ contest.score_precision }}"></td>
     </tr>
+    <tr>
+      <td>
+        <span class="info" title="Allow contestant to make unofficial submission before analysis mode"></span>
+        Allow unofficial submission before analysis mode
+      </td>
+      <td>
+        <input type="checkbox" id="allow_unofficial_submission_before_analysis_mode" name="allow_unofficial_submission_before_analysis_mode" {{ "checked" if contest.allow_unofficial_submission_before_analysis_mode else "" }}/>
+      </td>
+    </tr>
 
     <tr><td colspan=2><h2>Logging in</h2></td></tr>
     <tr>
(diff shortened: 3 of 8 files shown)


prandla:
after the end of the contest, the contest overview looks like this:
image
i think it'd make sense for it to also display the "task overview" table, like it does during the contest.


Also, the "allow unofficial submission before analysis" button really does only work "before analysis", by which i mean, if analysis mode is disabled for a contest then it does not work.

veluca93:
> after the end of the contest, the contest overview looks like this: image i think it'd make sense for it to also display the "task overview" table, like it does during the contest.
> 

Done.

> Also, the "allow unofficial submission before analysis" button really does only work "before analysis", by which i mean, if analysis mode is disabled for a contest then it does not work.

That makes sense to me.

wil93:
We're missing the SQL updater for this, @veluca93 could you add it?

wil93:
We have been providing SQL updaters until now, I would prefer to continue to do so, and possibly think about another easy upgrade path for CMS users in the future. We did some tricky SQL updaters in the past, is there a reason to not do it for smaller / easier changes like this one?

prandla:
thr last sql file in cmscontrib/updaters is for update_25, so i'm not very sure about the "until now".

in principle i'm not opposed to having them, but i think it leads to duplicated work (for basic changes duplication between sqlalchemy model definitions and the sql files, for complicated changes duplication between the python updater and sql updater), and duplicated work can lead to mistakes which would be hard to debug (forgetting a new constraint in the sql updater will cause some people to have a DB where a constraint is missing). this could be mitigated by somehow automatically generating the migrations, but i don't know any tools for that (though i haven't looked much either).

veluca93:
IMO the correct option is just not to support updates for now. The vast majority of CMS instances are not long-running anyway, and it's by far too easy to introduce inconsistent situations with upgrades; in some cases supporting database updates would require more complex code, and I don't believe that's the correct tradeoff for CMS.

veluca93:
> In one way or the other we must support updating from 1.5 to (say) 1.6, and in the release notes we will have to write some instructions on how to update. I put two options for this in my previous comment (export vs in-place)

I disagree with the premise.

wil93:
Do you suggest to release a new version with no documentation on how to upgrade? (i.e. like we did for 1.5)

I think it's not very nice to let CMS users figure out the upgrade path, we did this for 1.5 because there were no releases for so long that it was not really feasible to provide a proper list of "what changed" and upgrade path. For next releases it would be good to do that.

Issue #1362, opened 2025-06-06
[IOI fork parity] wording changes to precompiled question answers


prandla:
https://github.com/ioi-2024/cms/blob/363a804627067a74f44f9cc349b9d20ded058465/cms/server/admin/handlers/contestquestion.py#L71 this seems to be the current set of quick answers. also, these strings are duplicated in at least 3 files (admin/handlers/contestquestion.py, admin/templates/participation.html, admin/templates/questions.html), they should probably be refactored a bit.

PR #1394, opened 2025-06-07
Update wording of quick answers.

veluca93:
See #1362.

We might want to consider refactoring these as a follow-up.


prandla:
LGTM

PR #1395, opened 2025-06-07
RankingWebServer improvements

### LINKED ISSUE #1360: [IOI fork parity] RWS appearance changes
* score gradient is white to green instead of red to green
* at some point, ranking showed username as a separate column. (i don't know what happened to this; it seems to still be in the code that was used at 2024, but it's not in the archived ranking. maybe someone deleted it from the archived version manually.)
* ranking shows flags as a separate column.

veluca93:
Fixes #1360

### DIFF
--- a/cmsranking/static/Ranking.css
+++ b/cmsranking/static/Ranking.css
@@ -913,43 +913,31 @@ abbr {
     min-height: 242px;
 }
 
-#UserDetail_f_name,
-#UserDetail_l_name,
-#UserDetail_team,
-#UserDetail_flag {
-    position: absolute;
-    right: 470px;
+#UserDetail_summary {
+    display: flex;
+    flex-direction: column;
     font-size: 2em;
     line-height: 1.5em;
-    max-width: 63px;
+    text-align: center;
 }
 
-#UserDetail_f_name {
-    top: 0;
-}
-
-#UserDetail_l_name {
-    top: 1.5em;
-}
-
-#UserDetail_team {
-    bottom: 100px;
+#UserDetail_spacer {
+    flex-grow: 1;
+    min-height: 2em;
 }
 
 #UserDetail_flag {
-    bottom: 0;
+    max-width: 144px;
+    max-height: 96px;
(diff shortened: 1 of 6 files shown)


prandla:
i feel like showing the user id column should be toggleable somewhere. for most contests it's probably not very useful. (also, i wonder if anyone is going to miss the red-green color scheme...)

veluca93:
> i feel like showing the user id column should be toggleable somewhere. for most contests it's probably not very useful. (also, i wonder if anyone is going to miss the red-green color scheme...)

I feel like it can be useful for quite a few contest, but I can see wanting to disable it too.

IMO it's fine to enable it unconditionally for now and make it more configurable later.

cmsranking/static/Ranking.css
```diff
@@ -25,7 +25,7 @@ a {
 }
 
 abbr {
-    border-bottom: 1px dotted black;
+    text-decoration: none;
```
prandla, line 28: i'm not sure I like this change. the new style makes it very non-obvious that these elements have a `<title>` at all.

cmsranking/static/Ranking.css
```diff
@@ -385,6 +390,10 @@ abbr {
     width: 48px;
 }
 
+#Scoreboard_cols col.user_id {
+    width: 48px;
```
prandla, line 394: this width hardcodes the assumption that user IDs are IOI-style contestant identifiers (i.e. are 4 letters long). In other use cases, they might be longer (e.g. in estonia we use first initial + last name).

But the layout of these columns seems quite janky anyways, so I don't know how easy it is to improve this.

cmsranking/static/Ranking.css
veluca93: I don't think it's *too* important to notice that the problem names also have the title.

cmsranking/static/Ranking.css
veluca93: Agreed that this could be probably improved, but let's go with this for now.

PR #1396, opened 2025-06-08
Handle percent signs in manager output properly

gollux:
Evaluation outcome is stored as a formatted string with "%" serving as a formatting control character. We therefore need to escape outcomes output by managers.

Before we merge this fix, I would like to discuss sanitization of manager output in general. Currently, there is a mysterious function `_filter_ansi_escape` that tries to remove certain ANSI escape sequences (the original goal is probably to skip color switching sequences, but it doesn't match their syntax properly). I don't think we should do it - no program should ever emit color switching sequences without checking that the output is connected to a terminal. If it does, it should be fixed instead of CMS working around the issue.

I advocate for kicking out this trickery and outright refusing any message that contains control characters, possibly except TAB.

### DIFF
--- a/cms/grading/steps/trusted.py
+++ b/cms/grading/steps/trusted.py
@@ -36,6 +36,7 @@ can be translated by writing "translate:x" where x is "success", "partial" or
 """
 
 import logging
+import re
 
 from cms import config
 from cms.grading.Sandbox import Sandbox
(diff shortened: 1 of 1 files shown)


veluca93:
> Currently, there is a mysterious function `_filter_ansi_escape` that tries to remove certain ANSI escape sequences (the original goal is probably to skip color switching sequences, but it doesn't match their syntax properly). I don't think we should do it - no program should ever emit color switching sequences without checking that the output is connected to a terminal. If it does, it should be fixed instead of CMS working around the issue.
> 
> I advocate for kicking out this trickery and outright refusing any message that contains control characters, possibly except TAB.

100% agree here.

gollux:
We are already refusing messages that are not proper UTF-8 (instead of ignoring the offending bytes) and I don't see why control characters should be treated differently.

Also, if somebody outputs a control sequence and we drop the initial escape character, the result will be gibberish.

veluca93 (approved):

PR #1397, opened 2025-06-08
Set RankingWebServer listen address to "" in sample config

prandla:
This matches the defaults in cms.conf.sample. In particular, this is useful because when using cms-dev.sh, the default config leaves RWS listening at 127.0.0.1, which is not accessible outside docker, even though docker-compose.dev.yml specifically exposes port 8890 for it.

Also, I changed AWS from 0.0.0.0 to `""`. `""` seems to listen on both ipv6 and ipv4, as opposed to 0.0.0.0 which is ipv4 only, so it feels like a slightly better choice here.

### DIFF
--- a/config/cms.conf.sample
+++ b/config/cms.conf.sample
@@ -142,7 +142,7 @@
     "_help": "Listening HTTP address and port for the AWS. If you access",
     "_help": "it through a proxy running on the same host you could put",
     "_help": "127.0.0.1 here for additional security.",
-    "admin_listen_address": "0.0.0.0",
+    "admin_listen_address": "",
     "admin_listen_port":    8889,
 
     "_help": "Login cookie duration for admins in seconds.",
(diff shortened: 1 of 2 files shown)


veluca93:
I'm not sure this is a great idea... I think I'd like having everything listen on 127.0.0.1 by default, and have cms-dev.sh use nginx or similar to expose things, would be better.

prandla:
also, i don't like the idea of using nginx for cms-dev.sh, i don't think it provides any benefit really. (and suffers from the exact same problem: what address do you set nginx to listen on? if you set it to listen on 0.0.0.0 and forward all requests to cms, then it's identical to just having cms itself listen on 0.0.0.0)

veluca93:
> also, i don't like the idea of using nginx for cms-dev.sh, i don't think it provides any benefit really. (and suffers from the exact same problem: what address do you set nginx to listen on? if you set it to listen on 0.0.0.0 and forward all requests to cms, then it's identical to just having cms itself listen on 0.0.0.0)

I agree it is identical. However, doing so makes the default configuration not expose rws/aws to everyone, which is IMO valuable.

prandla:
In that case i'll just implement the rewrites in Dockerfile.

PR #1398, opened 2025-06-09
Round (sub)task scores to correct precision when displaying

prandla:
* The headings of subtasks in the submission details view were hardcoded to a precision of 2 decimal places. Round them to the correct precision.
* Send correctly rounded scores to RWS (instead of hardcoding 2 decimal places). RWS itself already handles score_precision correctly.
* Compute task score from rounded subtask scores.

First two points are clearly bugs. The last point is a little more debatable; it would theoretically be possible to have higher precision for subtasks than for the whole task. But I feel like that could be confusing to contestants, and rarely useful.

### DIFF
--- a/cms/grading/scoretypes/abc.py
+++ b/cms/grading/scoretypes/abc.py
@@ -238,10 +238,9 @@ class ScoreTypeGroup(ScoreTypeAlone):
         <span class="title">
             {% trans index=st["idx"] %}Subtask {{ index }}{% endtrans %}
         </span>
-    {% if "score_fraction" in st and "max_score" in st %}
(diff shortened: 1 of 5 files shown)


veluca93:
On the first two I agree.

The third however is problematic, as it may make it impossible to get 100 points -- for example if for whatever reason someone decided to make 9 subtasks each worth 11.(1) points.

IMO it should at the very least be in a different PR, and I am not sure we want it at all.

prandla:
actually, the scores in the ranking are still computed from rounded subtask scores (when using max over subtask scoring). fixing that requires more refactoring though; it's not a 2-line change.

PR #1399, opened 2025-06-09
Set listen addresses in sample config to 127.0.0.1

prandla:
See #1397.

the cms-test config file isn't altered, as the test machine doesn't need to be accessible from outside.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -49,10 +49,11 @@ RUN sudo pip3 install --break-system-packages .
 
 RUN sudo python3 prerequisites.py --yes --cmsuser=cmsuser install
 
-RUN sudo sed 's|/cmsuser:your_password_here@localhost:5432/cmsdb"|/postgres@testdb:5432/cmsdbfortesting"|' ./config/cms.conf.sample \
+RUN sed 's|/cmsuser:your_password_here@localhost:5432/cmsdb"|/postgres@testdb:5432/cmsdbfortesting"|' ./config/cms.conf.sample \
     | sudo tee /usr/local/etc/cms-testdb.conf
-RUN sudo sed 's|/cmsuser:your_password_here@localhost:5432/cmsdb"|/postgres@devdb:5432/cmsdb"|' ./config/cms.conf.sample \
+RUN sed -e 's|/cmsuser:your_password_here@localhost:5432/cmsdb"|/postgres@devdb:5432/cmsdb"|' -e 's/127.0.0.1/0.0.0.0/' ./config/cms.conf.sample \
     | sudo tee /usr/local/etc/cms-devdb.conf
+RUN sed 's/127.0.0.1/0.0.0.0/' ./config/cms.ranking.conf.sample | sudo tee /usr/local/etc/cms.ranking.conf
 
 ENV LANG C.UTF-8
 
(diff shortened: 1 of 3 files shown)


config/cms.conf.sample
```diff
@@ -105,7 +105,7 @@
     "_help": "in core_services. If you access them through a proxy (acting",
     "_help": "as a load balancer) running on the same host you could put",
     "_help": "127.0.0.1 here for additional security.",
```
veluca93: Can you update the "comment" here?

config/cms.conf.sample
```diff
@@ -142,7 +142,7 @@
     "_help": "Listening HTTP address and port for the AWS. If you access",
     "_help": "it through a proxy running on the same host you could put",
     "_help": "127.0.0.1 here for additional security.",
```
veluca93: same

veluca93 (approved):

Dockerfile
```diff
@@ -49,10 +49,11 @@ RUN sudo pip3 install --break-system-packages .
 
 RUN sudo python3 prerequisites.py --yes --cmsuser=cmsuser install
 
-RUN sudo sed 's|/cmsuser:your_password_here@localhost:5432/cmsdb"|/postgres@testdb:5432/cmsdbfortesting"|' ./config/cms.conf.sample \
+RUN sudo sed -e 's|/cmsuser:your_password_here@localhost:5432/cmsdb"|/postgres@testdb:5432/cmsdbfortesting"|' ./config/cms.conf.sample \
```
wil93: nit: we can probably remove the first sudo from these 3 lines? (I think we only need it for `tee`, not for `sed`)

Dockerfile
```diff
@@ -49,10 +49,11 @@ RUN sudo pip3 install --break-system-packages .
...
     | sudo tee /usr/local/etc/cms-testdb.conf
-RUN sudo sed 's|/cmsuser:your_password_here@localhost:5432/cmsdb"|/postgres@devdb:5432/cmsdb"|' ./config/cms.conf.sample \
+RUN sudo sed -e 's|/cmsuser:your_password_here@localhost:5432/cmsdb"|/postgres@devdb:5432/cmsdb"|' -e 's/127.0.0.1/0.0.0.0/' ./config/cms.conf.sample \
     | sudo tee /usr/local/etc/cms-devdb.conf
+RUN sudo sed -e 's/127.0.0.1/0.0.0.0/' ./config/cms.ranking.conf.sample | sudo tee /usr/local/etc/cms.ranking.conf
```
wil93: This is replacing all occurrences of 127.0.0.1, right? I wonder if this could change **too many** instances, for example we have `prometheus_listen_address` which probably we don't want to change? Not sure if it's really dangerous, but maybe we can make the expression more precise to target exactly the occurrences we need

prandla: Well, if you wanted to test prometheus then presumably you would also want to run it in a different container or just outside docker, in which case you would in fact need this replacement to happen. Besides, the default prometheus port isn't exposed by the docker-compose config anyways. But it would indeed be incorrect if there was some instance of 127.0.0.1 that wasn't a listen address.

prandla: I'm inclined to still leave it as is for now. If we ever add something to the config that would need to stay as exactly 127.0.0.1, we can fix the dockerfile then.

PR #1401, opened 2025-06-09
Better formatting of CPU time when getting TLE

### LINKED ISSUE #1346: Better formatting of CPU time when getting TLE
We should change the time shown in case of TLE to something like `> 5.000 sec` as opposed to the current `5.032 sec`. We've had beginners get confused and think "oh, my solution is so close to passing, i just need to save a few ms", this should make it more clear.

see https://github.com/e-i-o/cms/commit/c2c4c52a26727bc156271a57018d716013f2a55d (though it's a bit more complicated there because we have separate time limits for some languages).

prandla:
closes #1346.

### DIFF
--- a/cms/grading/scoretypes/abc.py
+++ b/cms/grading/scoretypes/abc.py
@@ -36,6 +36,7 @@ from abc import ABCMeta, abstractmethod
 
 from cms import FEEDBACK_LEVEL_RESTRICTED
 from cms.db import SubmissionResult
+from cms.grading.steps import EVALUATION_MESSAGES
 from cms.locale import Translation, DEFAULT_TRANSLATION
 from cms.server.jinja2_toolbox import GLOBAL_ENVIRONMENT
 from jinja2 import Template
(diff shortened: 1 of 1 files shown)


cms/grading/scoretypes/abc.py
```diff
@@ -430,11 +433,17 @@ def compute_score(self, submission_result):
...
+                if evaluations[tc_idx].text == [EVALUATION_MESSAGES.get("timeout").message]:
+                    dataset = evaluations[tc_idx].dataset
+                    time_override = dataset.time_limit
+
                 testcases.append({
                     "idx": tc_idx,
                     "outcome": tc_outcome,
                     "text": evaluations[tc_idx].text,
                     "time": evaluations[tc_idx].execution_time,
+                    "time_override": time_override,
```
veluca93: A bit of a nitpick, but I don't really like "time_override" as a name, mostly because intuitively one would think it would be formatted the same way as "time"

prandla: yeah this name is a bit random. what do you think is a better one? "time_limit_when_exceeded" maybe?

veluca93: "time_is_above" maybe?

prandla: I'll just make it two variables instead, that way it's easier to name them.

PR #1402, opened 2025-06-09
Render announcements, question replies, and messages as markdown

### LINKED ISSUE #1352: Rendering announcements and question replies as markdown
We implemented a feature where admins can send messages as markdown. This allows nicer formatting of code snippets and allows creating links. We used [Python-Markdown](https://python-markdown.github.io/) for it. implementation at: https://github.com/e-i-o/cms/commit/7c704d9322db214d83b4d4771ba1faf04919c9be

Open question: Python-Markdown allows arbitrary HTML. This means admins can inject arbitrary HTML/JS into CWS. Is this a security risk? If so, we should find a different markdown parser that supports sanitization.

### LINKED ISSUE #1362: [IOI fork parity] wording changes to precompiled question answers

prandla:
Closes #1352, closes #1362.

I actually realized that i had forgotten messages to users in my original implementation; this one handles those too.

I switched out the markdown parser to one that allows disabling html tags.

### DIFF
--- a/cms/server/admin/handlers/contestquestion.py
+++ b/cms/server/admin/handlers/contestquestion.py
@@ -68,16 +68,14 @@ class QuestionReplyHandler(BaseHandler):
     """Called when the manager replies to a question made by a user.
 
     """
-    QUICK_ANSWERS = {
-        "yes": "Yes",
-        "no": "No",
-        "invalid": "Invalid Question (not a Yes/No Question)",
-        "nocomment": "No Comment/Please refer to task statement",
-    }
 
     @require_permission(BaseHandler.PERMISSION_MESSAGING)
     def post(self, contest_id, question_id):
-        ref = self.url("contest", contest_id, "questions")
+        userid = self.get_argument("user_id", None)
+        if userid is not None:
+            ref = self.url("contest", contest_id, "user", userid, "edit")
+        else:
+            ref = self.url("contest", contest_id, "questions")
         question = self.safe_get_item(Question, question_id)
         self.contest = self.safe_get_item(Contest, contest_id)
 
@@ -90,12 +88,12 @@ class QuestionReplyHandler(BaseHandler):
         question.reply_text = self.get_argument("reply_question_text", "")
 
         # Ignore invalid answers
-        if reply_subject_code not in QuestionReplyHandler.QUICK_ANSWERS:
+        if reply_subject_code not in Question.QUICK_ANSWERS:
             question.reply_subject = ""
         else:
             # Quick answer given, ignore long answer.
             question.reply_subject = \
-                QuestionReplyHandler.QUICK_ANSWERS[reply_subject_code]
+                Question.QUICK_ANSWERS[reply_subject_code]
             question.reply_text = ""
 
         question.reply_timestamp = make_datetime()
@@ -118,7 +116,11 @@ class QuestionIgnoreHandler(BaseHandler):
     """
     @require_permission(BaseHandler.PERMISSION_MESSAGING)
     def post(self, contest_id, question_id):
-        ref = self.url("contest", contest_id, "questions")
+        userid = self.get_argument("user_id", None)
+        if userid is not None:
+            ref = self.url("contest", contest_id, "user", userid, "edit")
+        else:
+            ref = self.url("contest", contest_id, "questions")
         question = self.safe_get_item(Question, question_id)
         self.contest = self.safe_get_item(Contest, contest_id)
 
@@ -147,7 +149,11 @@ class QuestionClaimHandler(BaseHandler):
 
     @require_permission(BaseHandler.PERMISSION_MESSAGING)
     def post(self, contest_id, question_id):
-        ref = self.url("contest", contest_id, "questions")
+        userid = self.get_argument("user_id", None)
+        if userid is not None:
+            ref = self.url("contest", contest_id, "user", userid, "edit")
+        else:
+            ref = self.url("contest", contest_id, "questions")
         question = self.safe_get_item(Question, question_id)
         self.contest = self.safe_get_item(Contest, contest_id)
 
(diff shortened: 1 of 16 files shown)


cms/server/jinja2_toolbox.py
```diff
@@ -138,6 +140,19 @@ def today(ctx: Context, dt: datetime) -> bool:
...
+    """Renders text as markdown and returns a Markup object
+    corresponding to it.
+
+    text: The markdown text to render.
+
+    return: rendered HTML wrapped in the Markup class.
+    """
+    md = markdown_it.MarkdownIt('js-default')
+    html = md.render(text)
+    return markupsafe.Markup(html)
```
veluca93, line 153: Why do we need this markupsafe line?
AFAIU markdown_it claims to be secure by default, and "render it first but then escape problematic things" is, uh, usually not the best approach

cms/server/admin/templates/announcements.html
```diff
@@ -24,7 +24,8 @@ <h2 id="title_announcements" class="toggling_on">Announcements</h2>
           </div>
           <div class="notification_text">
             <label for="new_text">Text</label>
-            <textarea name="text" id="new_text" style="width: 100%" ></textarea>
+            <textarea name="text" id="new_text" style="width: 100%" ></textarea><br/>
```
veluca93: I feel like it would be nice to have a preview before the message/announcement is sent, but I also see how that would be hard so perhaps we can skip this for now.

cms/server/jinja2_toolbox.py
prandla: This just tells jinja to not escape it further, it doesn't actually do any sanitization.

cms/server/admin/templates/announcements.html
prandla: I agree it would be useful, let me try to write something for it actually.

cms/server/jinja2_toolbox.py
veluca93: Ah, fair enough...

prandla:
gah, i noticed a bug about question-answering (with the precompiled answer dropdown, the js hardcodes that the "other" option is at index 5, which it isn't anymore), and i also noticed that there's a bunch of JS that's duplicated (including duplicating the bug). Do you mind if i refactor it in this same PR? Because i need to refactor that JS a little anyways to add markdown preview.

cms/server/admin/handlers/contestquestion.py
```diff
@@ -68,16 +68,14 @@ class QuestionReplyHandler(BaseHandler):
...
-    }
 
     @require_permission(BaseHandler.PERMISSION_MESSAGING)
     def post(self, contest_id, question_id):
-        ref = self.url("contest", contest_id, "questions")
+        userid = self.get_argument("user_id", None)
+        if userid is not None:
+            ref = self.url("contest", contest_id, "user", userid, "edit")
+        else:
+            ref = self.url("contest", contest_id, "questions")
```
veluca93, line 84: Maybe it makes sense to make a function for this?

cms/server/admin/templates/macro/question.html
veluca93, line 1: Why does this template not render a single question, with the for loop being moved to the caller or another template?

I find the "if user_id == None do something entirely different" API somewhat ugly.

prandla: > Why does this template not render a single question, with the for loop being moved to the caller or another template?

That's what it does, no?

> I find the "if user_id == None do something entirely different" API somewhat ugly.

I agree.

The user_id is used for two things, 1) to not show a link to the user page if we're already on the user page, 2) to redirect to the right place after clicking the claim/ignore/reply buttons.

originally i had two variables for this, "show_user_link" and "redirect_url", but i couldn't construct redirect_url properly (the url() function always makes relative urls, and they are relative to the current page, but they need to be relative to the handler endpoint in order to work as redirects). So i needed to move the url() construction into the handler, and this seemed to be the easiest way to do it.

Maybe we can just redirect to the http referer? I remember that being disabled in some cases due to privacy concerns, but maybe it still works okay for intra-site requests.

veluca93: Ah, I didn't actually read the code and assumed things given the documentation:

> user_id (int|None): if we are rendering a single user's questions, this is the
    user's id. if we are rendering all questions of the contest, then None.
    
Perhaps that should be clarified to avoid the misunderstanding (i.e. "one of the questions in the list of questions of the contest")

I think I am fine with passing a user id here, given how it _actually_ affects things

veluca93 (approved):

cms/server/admin/handlers/contestquestion.py
prandla: I noticed that these classes already had some more duplicated logic, i moved all of it to a new base class.

cms/server/admin/handlers/contestquestion.py
```diff
@@ -64,38 +64,50 @@ def get(self, contest_id):
...
-    QUICK_ANSWERS = {
-        "yes": "Yes",
-        "no": "No",
-        "invalid": "Invalid Question (not a Yes/No Question)",
-        "nocomment": "No Comment/Please refer to task statement",
-    }
+class QuestionActionHandler(BaseHandler):
+    """Base class for handlers for actions on questions."""
+
+    def do(self, question: Question):
```
veluca93: I don't really like the name...

`process_question`?

Also I think this should throw an exception.

prandla: i renamed it and made it an abstract method.

PR #1405, opened 2025-06-10
Optimize some unit tests

prandla:
* Some tests end up having to hash passwords. Add a tweak that allows setting the hashing difficulty to a lower value for the tests.
* file_middleware_test had a really inefficient setUp() method, which matters more now that the FileCacher block size is 1MB.

This should save around 10 seconds in CI (according to codecov test timings).

### DIFF
--- a/cmscommon/crypto.py
+++ b/cmscommon/crypto.py
@@ -48,6 +48,9 @@ __all__ = [
 
 _RANDOM = Random.new()
 
+# bcrypt difficulty parameter. This is here so that it can be set to a lower
+# value when running unit tests. It seems that the lowest accepted value is 4.
+BCRYPT_ROUNDS = 12
 
 def get_random_key() -> bytes:
     """Generate 16 random bytes, safe to be used as AES key.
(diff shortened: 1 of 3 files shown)


cmstestsuite/unit_tests/cmscommon/crypto_test.py
```diff
@@ -167,6 +168,10 @@ def test_default_plaintext(self):
 class TestHashAndValidatePassword(unittest.TestCase):
     """Tests for the functions hash_password and validate_password."""
 
+    def setUp(self):
+        # make the hashing faster.
+        cmscommon.crypto.BCRYPT_ROUNDS = 4
```
veluca93: Shouldn't we perhaps set this in a base class?

prandla: actually i think it might be useful to set this globally for all tests... there might be some other tests doing bcrypt internally which i didn't notice, and it shouldn't hurt anything to use the lower difficulty for all tests, but i don't know where a good place to set that up would be.

veluca93: That's why I was thinking of wrapping the `unittest.TestCase` base class :-)

But this could also go on the first line of some top-level file in cmstestsuite, ig

prandla: do you think the new solution is fine or is this too hacky now? :)

veluca93: Seems fine to me.

PR #1406, opened 2025-06-10
Remove use of patoolib, along with other archive-related improvements

### LINKED ISSUE #1226: DumpExporter fills up RAM
`cmsDumpExporter` will put everything in a tempdir (i.e. usually `/tmp`) before packing it up in an archive at the desired path:
https://github.com/cms-dev/cms/blob/7d962150d9969281148d3eaba1ab35823fdfb1b6/cmscontrib/DumpExporter.py#L189-L190

On some systems, including Arch, `/tmp` is in RAM by default, and limited to half the RAM's size. At least on personal computers, this will usually cause the system to freeze or crash. Of course, CMS writes to `/tmp` on other occasions, but in normal operation, it will cleanup shortly afterwards and never write something as large as the whole database at once.

[It seems that](https://docs.python.org/3/library/tempfile.html#tempfile.gettempdir) by setting one of the `TMPDIR`, `TEMP`, or `TMP` environment variables, one can specify another location, so circumventing the issue is technically easy. However, not everyone will know this and think of it beforehand, and a system freeze is annoying, as is finding out the cause of the problem.

Thus, I would like to suggest improving this somehow.
- The files have to be written somewhere of course, and using `tempfile.mkdtemp` makes sense. Easy and still helpful would be a warning, so at least one will know why the system froze (or be able to hit `STRG+C` if fast enough...). This should include suggesting to set `TMPDIR` to somewhere big enough.
- Maybe it could even explicitly ask along the lines of `Path to temporary directory (default is system temp dir; sufficient free space is needed):`
- Or it could use the directory where it shall export the archive. Drawback: If the exporter crashes, it won't be cleaned up.

This sounds similar to #322, but I think that was more about importing and exporting specific contests and is unrelated.

Happy to hear your thoughts!

PS: This has been in the back of my head since forever and I haven't seen someone else mention it, so who knows if I'm the only one with this problem. :smile:

prandla:
This removes use of patoolib in CWS, instead using Python standard libraries, which are much safer due to not extracting anything to disk.

This also allowed implementing zip bomb protection: any archive with too many files or too big files is rejected before fully decompressing it.

I also rewrote the archive handling of DumpExporter aiming to fix #1226. The dump archive is constructed on the fly, without storing too much things in RAM or in temporary directories. (The entire model is still built in RAM and converted to json in RAM though. Hopefully it doesn't get too big...)

This is WIP because I want to also implement corresponding functionality in DumpUpdater and DumpImporter, which would allow fully removing patoolib.

The obvious drawback of this is that the set of supported archive formats is smaller: only .zip, .tar, .tar.gz, .tar.bz2, and .tar.xz. (And .tar.zst starting from Python 3.14.)

### DIFF
--- a/cms/server/contest/submission/file_retrieval.py
+++ b/cms/server/contest/submission/file_retrieval.py
@@ -33,6 +33,7 @@ format.
 
 """
 
+import io
 import os.path
 import typing
 
@@ -40,7 +41,7 @@ from patoolib.util import PatoolError
 if typing.TYPE_CHECKING:
     from tornado.httputil import HTTPFile
 
-from cmscommon.archive import Archive
+from cmscommon.archive import open_archive
 
 
 # Represents a file received through HTTP from an HTML form.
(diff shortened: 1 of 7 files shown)


prandla:
after mulling over this a bit, i'm not sure how i'd want to implement it in DumpExporter/Importer/Updater in a way that properly addresses its performance concerns. The change in CWS is more important though, so i'll make a separate pull request for that, and we can think about how to handle dumps later.

PR #1410, opened 2025-06-12
Use toml for configuration

prandla:
* changed filename format to cms.toml, cms_sample.toml, cms_ranking.toml, and cms_ranking_sample.toml. i disliked the old .sample setup because then editors don't syntax-highlight them automatically.
* added a warning for when the config contains an unknown value.
* added missing values to cms/conf.py (which i noticed because of that warning :) )
* The config file is a little bit messy currently imho, i think it would be good to use sections properly (convert the `####`-delimited sections into toml tables). currently i needed to move the services table to the very bottom, because once you leave the default table, there's no way to enter it again. But of course this means changing every part of cms that uses the config.
* Would also be nice to have static type-checking on the config values, eventually.


cms/conf.py
```diff
@@ -81,7 +81,7 @@ class AsyncConfig:
 class Config:
     """This class will contain the configuration for CMS. This needs
     to be populated at the initilization stage. This is loaded by
-    default with some sane data. See cms.conf.sample in the config
+    default with some sane data. See cms_sample.toml in the config
```
veluca93: Why not cms.sample.toml? IMO it's a bit more natural

cms/conf.py
```diff
@@ -202,21 +210,18 @@ def _load(self, paths: list[str]):
...
 
         """
         # Load config file.
         try:
-            with open(path, 'rt', encoding='utf-8') as f:
-                data = json.load(f)
+            with open(path, 'rb') as f:
+                data = tomllib.load(f)
         except FileNotFoundError:
             logger.debug("Couldn't find config file %s.", path)
```
veluca93: I would perhaps explicitly point out it was renamed from `cms.conf` :-)

prandla: I changed it, but because this is logging.debug i'm not sure if it's even visible anywhere. At least the terminal output doesn't show it.

prandla: FWIW, this is currently the output when specifying an invalid config file path (after hacking log.py to also show debug messages on stderr).
image

I think this could be improved in general, maybe we should just make a separate issue for it.

veluca93: Agreed, on both points.

Issue #1407, opened 2025-06-12
Allow configuring expiration length of login cookie


prandla:
We discussed this on telegram a bit, putting it here so we don't forget.

Currently the CWS login cookie is a session cookie. Its expiration should instead be configurable by admins, probably in cms.conf.

AWS cookie duration should probably be configurable too, but I don't know if we want it to always have the same length as CWS's one or not (i.e. do we need one option in cms.conf or two).

prandla:
Turns out we already have config options for cookie duration :D (AWS and CWS separately even), but it's only validated server-side.

Then i think it would make sense to make AWS/CWS send them with the same duration.

PR #1411, opened 2025-06-12
Some changes to authentication.

veluca93:
- Do not use session cookies for AWS/CWS, but regular cookies
- Allow authentication in CWS without IP autologin or cookies, but with a `X-CMS-Authorization` header.

### DIFF
--- a/cms/server/admin/authentication.py
+++ b/cms/server/admin/authentication.py
@@ -144,7 +144,8 @@ class AWSAuthMiddleware:
             """
             response = Response(status=status, headers=headers)
             self._cookie.save_cookie(
-                response, AWSAuthMiddleware.COOKIE, httponly=True)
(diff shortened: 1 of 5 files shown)


cms/server/admin/authentication.py
```diff
@@ -24,7 +24,7 @@
 from werkzeug.local import Local, LocalManager
 from werkzeug.wrappers import Request, Response
 
-from cms import config
+from cms import conf, config
```
prandla: was this intentional?

prandla (commented):
i'm not sure how much this kind of header really helps. if the format and thr validity rules are the exact same as the cookie, then how is this any more useful than just sending the `Cookie` header directly?

veluca93:
Closes #1407

PR #1412, opened 2025-06-13
AWS: new code popup

### LINKED ISSUE #1351: AWS submission code view improvements
We mostly rewrote the code view popup in AWS:

Image

The old one had somewhat janky scrolling, dated syntax highlighting (and broken text selection when the syntax highlighting didn't work), and poor behavior when the browser window is small. 

We switched to prism.js for syntax highlighting, and also implemented a "copy" button (selects the entire submission and copies to clipboard) and the ability to exit the modal with Esc.

prandla:
Closes #1351


prandla:
There's one in the linked issue, here's one more:
image

The box shrinks both horizontally and vertically when the file is small enough, but stays near the right side. The max width (1000px) is such that on a usual-sized screen (1920px wide), you can still see the file links of other submissions, which makes it easier to click through them.

PR #1415, opened 2025-06-14
Order tasks and contests by name

pxsit:
Order tasks and contests by name instead of randomly

### DIFF
--- a/cms/server/admin/handlers/base.py
+++ b/cms/server/admin/handlers/base.py
@@ -23,7 +23,9 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-"""Base class for all handlers in AWS, and some utility functions."""
+"""Base class for all handlers in AWS, and some utility functions.
+
+"""
 
 from collections.abc import Callable
 import ipaddress
@@ -52,25 +54,14 @@ from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm import Query, subqueryload
 
 from cms import __version__, config
-from cms.db import (
-    Admin,
-    Contest,
-    Participation,
-    Question,
-    Submission,
-    SubmissionResult,
-    Task,
-    Team,
-    User,
-    UserTest,
-)
+from cms.db import Admin, Contest, Participation, Question, Submission, \
+    SubmissionResult, Task, Team, User, UserTest
 import cms.db
 from cms.grading.scoretypes import get_score_type_class
 from cms.grading.tasktypes import get_task_type_class
 from cms.server import CommonRequestHandler, FileHandlerMixin
 from cmscommon.crypto import hash_password, parse_authentication
 from cmscommon.datetime import make_datetime
-
 if typing.TYPE_CHECKING:
     from cms.server.admin import AdminWebServer
 
@@ -142,7 +133,7 @@ def parse_timedelta_min(value: str) -> timedelta:
 
 def parse_datetime(value: str) -> datetime:
     """Parse and validate a datetime (in pseudo-ISO8601)."""
-    if "." not in value:
+    if '.' not in value:
         value += ".0"
     try:
         return datetime.strptime(value, "%Y-%m-%d %H:%M:%S.%f")
@@ -187,11 +178,9 @@ def require_permission(permission: str = "authenticated", self_allowed: bool = F
        permission.
 
     """
-    if permission not in [
-        BaseHandler.PERMISSION_ALL,
-        BaseHandler.PERMISSION_MESSAGING,
-        BaseHandler.AUTHENTICATED,
-    ]:
+    if permission not in [BaseHandler.PERMISSION_ALL,
+                          BaseHandler.PERMISSION_MESSAGING,
+                          BaseHandler.AUTHENTICATED]:
         raise ValueError("Invalid permission level %s." % permission)
 
     _P = typing.ParamSpec("_P")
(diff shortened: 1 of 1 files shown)


wil93 (approved):

prandla:
i'd rather the contest list be sorted in reverse chronological order of start_time, so that the 5 contests that show up on the sidebar are more likely to be relevant.

...but maybe that should be handled in the sidebar rendering code instead.

veluca93:
IMO the ordering should be consistent, so I am in favour of not doing special logic in the sidebar rendering.

cms/server/admin/handlers/base.py
```diff
@@ -69,6 +70,13 @@
 logger = logging.getLogger(__name__)
 
 
+def natural_sort_key(text):
+    #Generate a natural sorting key for alphanumeric strings.
```
veluca93: nitpick: `# Generate`

cms/server/admin/handlers/base.py
```diff
@@ -345,9 +353,15 @@ def render_params(self) -> dict:
                 .filter(Question.ignored.is_(False))\
                 .count()
         # TODO: not all pages require all these data.
-        params["contest_list"] = self.sql_session.query(Contest).all()
-        params["task_list"] = self.sql_session.query(Task).all()
-        params["user_list"] = self.sql_session.query(User).all()
+        params["contest_list"] = self.sql_session.query(Contest).order_by(Contest.name).all()
```
veluca93, line 349: What about `desc(Contest.start_time)` as @prandla suggested?

cms/server/admin/handlers/base.py
```diff
@@ -69,6 +70,13 @@
 logger = logging.getLogger(__name__)
 
 
+def natural_sort_key(text):
+    #Generate a natural sorting key for alphanumeric strings.
+    def convert(part):
+        return int(part) if part.isdigit() else part.lower()
+    return [convert(c) for c in re.split("([0-9]+)", str(text))]
```
veluca93: isn't all of this effectively equivalent to sorting by `text.lower()`?

IMO sorting by that is more than good enough.

cms/server/admin/handlers/base.py
pxsit: I think sorting them by name is better, but your call.

cms/server/admin/handlers/base.py
pxsit: I just pasted it in from one of my templates, should function the same tho.

cms/server/admin/handlers/base.py
wil93: I think if start_time is not visible in that table it would not be intuitive to sort by that.

My opinion: let's go with name for now, and later we could rewrite this UI to make it easy to sort on demand (e.g. you click on the name column to sort, click again to reverse sort, click on start_time column to sort by start time, etc...)

cms/server/admin/handlers/base.py
```diff
@@ -69,6 +70,13 @@
 logger = logging.getLogger(__name__)
 
 
+def natural_sort_key(text):
```
wil93: I think it was better to sort more simply, we can always improve it later (and if we decide to actually use natural sorting I would prefer to do it through a library that handles things like accent normalization etc)

wil93 (changes requested):

cms/server/admin/handlers/base.py
```diff
@@ -345,9 +353,15 @@ def render_params(self) -> dict:
                 .filter(Question.ignored.is_(False))\
                 .count()
         # TODO: not all pages require all these data.
-        params["contest_list"] = self.sql_session.query(Contest).all()
-        params["task_list"] = self.sql_session.query(Task).all()
-        params["user_list"] = self.sql_session.query(User).all()
```
wil93, line 350: Since we're doing this, let's also sort the users by username

cms/server/admin/handlers/base.py
```diff
@@ -68,7 +69,14 @@
...
 
-
+"""
+#TODO improve and implement this function
+def natural_sort_key(text):
+    #Generate a natural sorting key for alphanumeric strings.
+    def convert(part):
+        return int(part) if part.isdigit() else part.lower()
+    return [convert(c) for c in re.split("([0-9]+)", str(text))]
+"""
```
wil93: I think the TODO is enough. Let's remove this commented code, we can add it later (and it's easy to retrieve by tracing back to this PR following the git history).

wil93 (changes requested):

cms/server/admin/handlers/base.py
```diff
@@ -31,6 +31,7 @@
 import ipaddress
 import json
 import logging
+import re
```
wil93: We can remove this

cms/server/admin/handlers/base.py
```diff
@@ -68,7 +69,6 @@
 
 logger = logging.getLogger(__name__)
 
-
```
wil93: And restore this empty line. Actually, are you using VS code? It should automatically re-format on save, though you probably need to install some extension (it should automatically recommend to install Ruff)

wil93 (changes requested):

cms/server/admin/handlers/base.py
pxsit: Yes, I'm using Ruff and VS code.

wil93 (approved):
Thank you!

PR #1416, opened 2025-06-15
Added a remove function to teams

pxsit:
Added a remove function to teams and also made a check when adding a team to a user if that team actually exists or not

### DIFF
--- a/cms/server/admin/handlers/__init__.py
+++ b/cms/server/admin/handlers/__init__.py
@@ -90,15 +90,18 @@ from .task import \
     AttachmentHandler, \
     TaskListHandler, \
     RemoveTaskHandler
-from .user import \
-    AddUserHandler, \
-    UserHandler, \
-    UserListHandler, \
-    RemoveUserHandler, \
-    AddParticipationHandler, \
-    EditParticipationHandler, \
-    AddTeamHandler, \
-    TeamHandler
+from .user import (
+    AddUserHandler,
+    UserHandler,
+    UserListHandler,
+    RemoveUserHandler,
+    AddParticipationHandler,
+    EditParticipationHandler,
+    AddTeamHandler,
+    TeamHandler,
+    TeamListHandler,
+    RemoveTeamHandler,
+)
 from .usertest import \
     UserTestHandler, \
     UserTestFileHandler
@@ -115,55 +118,38 @@ HANDLERS = [
     (r"/notifications", NotificationsHandler),
     (r"/file/([a-f0-9]+)/([a-zA-Z0-9_.-]+)", FileFromDigestHandler),
     (r"/render_markdown", MarkdownRenderHandler),
-
     # Contest
-
     (r"/contests", ContestListHandler),
     (r"/contests/([0-9]+)/remove", RemoveContestHandler),
     (r"/contests/add", AddContestHandler),
     (r"/contest/([0-9]+)", ContestHandler),
     (r"/contest/([0-9]+)/overview", OverviewHandler),
     (r"/contest/([0-9]+)/resourceslist", ResourcesListHandler),
-
     # Contest's users
-
     (r"/contest/([0-9]+)/users", ContestUsersHandler),
     (r"/contest/([0-9]+)/users/add", AddContestUserHandler),
     (r"/contest/([0-9]+)/user/([0-9]+)/remove", RemoveParticipationHandler),
     (r"/contest/([0-9]+)/user/([0-9]+)/edit", ParticipationHandler),
     (r"/contest/([0-9]+)/user/([0-9]+)/message", MessageHandler),
-
     # Contest's tasks
-
     (r"/contest/([0-9]+)/tasks", ContestTasksHandler),
     (r"/contest/([0-9]+)/tasks/add", AddContestTaskHandler),
-
     # Contest's submissions / user tests
-
     (r"/contest/([0-9]+)/submissions", ContestSubmissionsHandler),
     (r"/contest/([0-9]+)/user_tests", ContestUserTestsHandler),
-
     # Contest's announcements
-
-    (r"/contest/([0-9]+)/announcements",
-     SimpleContestHandler("announcements.html")),
+    (r"/contest/([0-9]+)/announcements", SimpleContestHandler("announcements.html")),
     (r"/contest/([0-9]+)/announcements/add", AddAnnouncementHandler),
     (r"/contest/([0-9]+)/announcement/([0-9]+)", AnnouncementHandler),
-
     # Contest's questions
-
     (r"/contest/([0-9]+)/questions", QuestionsHandler),
     (r"/contest/([0-9]+)/question/([0-9]+)/reply", QuestionReplyHandler),
     (r"/contest/([0-9]+)/question/([0-9]+)/ignore", QuestionIgnoreHandler),
     (r"/contest/([0-9]+)/question/([0-9]+)/claim", QuestionClaimHandler),
-
     # Contest's ranking
-
     (r"/contest/([0-9]+)/ranking", RankingHandler),
     (r"/contest/([0-9]+)/ranking/([a-z]+)", RankingHandler),
-
     # Tasks
-
     (r"/tasks", TaskListHandler),
     (r"/tasks/([0-9]+)/remove", RemoveTaskHandler),
     (r"/tasks/add", AddTaskHandler),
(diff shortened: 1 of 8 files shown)


cms/server/admin/handlers/contestuser.py
```diff
@@ -235,14 +235,23 @@ def post(self, contest_id, user_id):
             self.get_bool(attrs, "unrestricted")
 
             # Update the participation.
-            participation.set_attrs(attrs)
-
-            # Update the team
+            participation.set_attrs(attrs)  # Update the team
```
veluca93: I think this comment got misplaced

cms/server/admin/templates/team.html
```diff
@@ -18,10 +18,20 @@ <h2 id="title_general_info" class="toggling_on">General Information</h2>
         <td>Name</td>
         <td><input type="text" name="name" value="{{ team.name }}"/></td>
       </tr>
-    </table>
-    <input type="submit" value="Update" />
+    </table>    <input type="submit" value="Update" />
```
veluca93: formatting is a bit messed up

veluca93 (approved):

cms/server/admin/templates/team.html
```diff
@@ -18,10 +18,21 @@ <h2 id="title_general_info" class="toggling_on">General Information</h2>
         <td>Name</td>
         <td><input type="text" name="name" value="{{ team.name }}"/></td>
       </tr>
-    </table>
+    </table>
```
veluca93: still some extra space

pxsit: Should be updated by now

cms/server/admin/handlers/user.py
```diff
@@ -145,6 +167,54 @@ def delete(self, user_id):
...
+        self.r_params["participation_count"] = participation_query.count()
+        self.render("team_remove.html", **self.r_params)
+
+    @require_permission(BaseHandler.PERMISSION_ALL)
+    def delete(self, team_id):
+        team = self.safe_get_item(Team, team_id)
+
+        try:
+            # First, we need to remove the team association from all participations
+            # rather than deleting the participations themselves
```
prandla: can't you do this automatically in the database by changing the participation->team foreign key to "on delete set null"?

pxsit:
One second, doing codeforce atm and removing it returns an error

cms/server/admin/handlers/__init__.py
wil93, line 1: Is it possible to make the necessary changes to this file (i.e. two lines of code) without the formatting updates? By default VS code should format only the changes, not the whole file.

cms/server/admin/handlers/contestuser.py
```diff
@@ -239,10 +239,21 @@ def post(self, contest_id, user_id):
 
             # Update the team
             self.get_string(attrs, "team")
-            team: Team | None = (
-                self.sql_session.query(Team).filter(Team.code == attrs["team"]).first()
-            )
-            participation.team = team
+            team_code = attrs.get("team", "")
+            if isinstance(team_code, str):
+                team_code = team_code.strip()
```
wil93: Is this necessary? (i.e. do we actually get back from the form this field with extra spaces in it?) and is it correct? (i.e. what if for some reason we wanted to allow spaces in the team name)

cms/server/admin/handlers/__init__.py
```diff
@@ -115,55 +117,38 @@
     (r"/notifications", NotificationsHandler),
     (r"/file/([a-f0-9]+)/([a-zA-Z0-9_.-]+)", FileFromDigestHandler),
     (r"/render_markdown", MarkdownRenderHandler),
-
```
wil93: Can we add back these empty lines?

cms/server/admin/handlers/contestuser.py
```diff
@@ -239,10 +239,17 @@ def post(self, contest_id, user_id):
 
             # Update the team
             self.get_string(attrs, "team")
-            team: Team | None = (
-                self.sql_session.query(Team).filter(Team.code == attrs["team"]).first()
-            )
-            participation.team = team
+            team_code = attrs.get("team", "")
```
wil93: The `get_string` call above seems to have an `empty=""` parameter, does that mean that we already default to an empty string there? https://github.com/cms-dev/cms/blob/f277dcc945589a3a86c99b60c1df26fa1f3cd84d/cms/server/admin/handlers/base.py#L375C5-L375C56

If yes, then we probably don't need the `.get()` here and we can do `attrs["team"]` like before

cms/server/admin/handlers/user.py
```diff
@@ -145,6 +167,53 @@ def delete(self, user_id):
...
+            # remove team association
+            participations = (
+                self.sql_session.query(Participation)
+                .filter(Participation.team == team)
+                .all()
+            )
+
+            for participation in participations:
+                participation.team = None
+                participation.team_id = None
```
wil93: Do we actually need to do this? Doesn't the `ondelete="SET NULL"` take care of this at the SQL layer?

pxsit: I tried that but it returns me an error, I don't remember much but let me try again

cms/server/admin/handlers/user.py
wil93: (Just realized that this point was already raised here).

I agree we should do it via the database also for consistency reasons. For example, it could happen that we start unlinking the participations, and when we are halfway done there is an error somewhere, and we end up with the team not deleted, half of the participations linked to the team, and the other half unlinked.

prandla: > I agree we should do it via the database also for consistency reasons. For example, it could happen that we start unlinking the participations, and when we are halfway done there is an error somewhere, and we end up with the team not deleted, half of the participations linked to the team, and the other half unlinked.

that should already be handled by database transactions, no? i just think set null is cleaner in this case

cms/server/admin/templates/team.html
```diff
@@ -12,16 +12,22 @@ <h2 id="title_general_info" class="toggling_on">General Information</h2>
...
       <tr>
         <td>Name</td>
-        <td><input type="text" name="name" value="{{ team.name }}"/></td>
+        <td><input type="text" name="name" value="{{ team.name }}" /></td>
       </tr>
     </table>
     <input type="submit" value="Update" />
     <input type="reset" value="Reset" />
   </form>
+  <form action="{{ url("teams") }}" method="POST" style="display:inline;">
```
wil93: I tested this locally and noticed that the "Remove" button sits slightly lower than the "Update" and "Reset" buttons. I suspect it's because the `display:inline` should be added to both forms (like it's done [here](https://github.com/cms-dev/cms/blob/f277dcc945589a3a86c99b60c1df26fa1f3cd84d/cms/server/admin/templates/user.html#L67-L138))

We should probably rework this anyway (and likely avoid ever using `style="..."`) but for now let's keep them consistent.

wil93:
We just added a script to perform SQL updates (which will be part of the upgrade instructions from 1.5 to 1.6).

Can you please add the following updater to it?

```
-- https://github.com/cms-dev/cms/pull/1416
ALTER TABLE ONLY public.participations DROP CONSTRAINT participations_team_id_fkey;
ALTER TABLE ONLY public.participations ADD CONSTRAINT participations_team_id_fkey FOREIGN KEY (team_id) REFERENCES public.teams(id) ON UPDATE CASCADE ON DELETE SET NULL;
```

cmscontrib/updaters/update_from_1.5.sql
```diff
@@ -14,10 +14,15 @@ ALTER TABLE public.contests ALTER COLUMN allow_unofficial_submission_before_anal
...
+ALTER TABLE ONLY public.participations DROP CONSTRAINT participations_team_id_fkey;
+ALTER TABLE ONLY public.participations ADD CONSTRAINT participations_team_id_fkey FOREIGN KEY (team_id) REFERENCES public.teams(id) ON UPDATE CASCADE ON DELETE SET NULL;
+
 -- https://github.com/cms-dev/cms/pull/1419
 ALTER TABLE submissions ADD COLUMN opaque_id BIGINT;
 UPDATE submissions SET opaque_id = id WHERE opaque_id IS NULL;
 ALTER TABLE submissions ADD CONSTRAINT participation_opaque_unique UNIQUE (participation_id, opaque_id);
 ALTER TABLE submissions ALTER COLUMN opaque_id SET NOT NULL;
 
+
```
wil93: nit: remove empty space here?

cms/server/admin/templates/team.html
wil93: Did you check that this change fixes the issue? (bonus points if you have a screenshot)

dev-requirements.txt
```diff
@@ -1,6 +1,6 @@
 # Only for testing
 beautifulsoup4>=4.8,<4.14
-coverage>=4.5,<7.9
+coverage>=4.5,<7.10
```
wil93: Let's avoid unrelated changes in the same PR, I would rather prefer to have a separate PR just for the dependency updates

cms/server/admin/templates/team.html
pxsit: image

Fixed It

PR #1419, opened 2025-06-18
Add a "opaque_id" column to submissions.

veluca93:
This gives a stable ID that can be used to refer to a specific submission.


wil93:
What is an opaque_id? Is it to avoid disclosing the submission.id to users?

wil93:
What is an opaque_id?

veluca93:
> What is an opaque_id really? And why we need it? It would be good to update the PR's description.

I believe the PR description already explains:

> This gives a stable ID that can be used to refer to a specific submission.

wil93:
@veluca93 not really, if you can add Martin's sentence `database IDs assigned serially would leak information on the number of submissions by other contestants` to the PR's description it would be much better.

cms/server/contest/handlers/contest.py
```diff
@@ -250,7 +250,7 @@ def get_task(self, task_name: str) -> Task | None:
             .filter(Task.name == task_name) \
             .one_or_none()
 
-    def get_submission(self, task: Task, submission_num: str) -> Submission | None:
```
wil93, line 253: Quick question: why can't we keep using `submission_num`? Could the API return this instead of opaque_id?

veluca93: `submission_num` is not stable to a bunch of things that could happen, such as time going backwards, submissions  being deleted, etc.

veluca93:
> i'm not sure i like this approach much, because of the possibility of collisions. even though it would take around 2^31 submissions in the DB before collisions become likely enough to be a problem, it still seems a bit unclean to me. (also, if there is collision-avoidance in the generate_opaque_id function, then we should also implement similar logic in the dumpupdater, to at least ensure the IDs don't collide within a dump.)
> 
> if we want to make collisions so unlikely as to make them really negligible, i think it'd be better to use UUIDs. alternatively we can use an encrypted form of the database ID (encrypted with e.g. the secret_key in cms.conf).
> 
> to be clear i'm not really opposed to the current implementation, it's probably easier than my proposed alternatives. (well, uuids actually shouldn't be more difficult.)

IMO if IDs collide in a dump you can just retry, that's unlikely enough.

During normal operations collisions are indeed avoided.

@gollux mentioned that UUID had some issues in postgres, but I don't remember exactly what.

I am opposed to using a function of the submission ID, because it introduces the possibility of mistakes leaking information.

cmscontrib/AddSubmission.py
```diff
@@ -22,6 +22,7 @@
 
 import argparse
 import logging
+import random
```
gollux: This looks unused.

cmscontrib/updaters/update_45.py
```diff
@@ -0,0 +1,45 @@
...
+    def __init__(self, data):
+        assert data["_version"] == 44
+        self.objs = data
+
+    def run(self):
+        for k, v in self.objs.items():
+            if k.startswith("_"):
+                continue
+            if v["_class"] == "Submission":
+                v["opaque_id"] = random.randint(0, 2**63-1)
```
gollux: This does not guarantee uniqueness.

veluca93: As far as I can tell the updater scripts don't have database access, so I am not sure how to guarantee uniqueness.

IMO it is fine to have a low probability of the importer failing, and just have people retry. Given that contests rarely have more than a few thousand submissions, the probability of collisions should be low enough.

I could guarantee no collisions within the dump itself, but I am not sure it's worth it.

cms/db/submission.py
```diff
@@ -177,6 +184,21 @@ def tokened(self) -> bool:
         """
         return self.token is not None
 
+    @classmethod
+    def generate_opaque_id(cls, session):
+        randint_upper_bound = 2**63-1
+
+        opaque_id = random.randint(0, randint_upper_bound)
+
+        while (session
```
gollux, line 199: I wonder if this is really correct wrt. concurrent transactions. If two transactions choose the same `opaque_id` that wasn't used yet, one of them violates the uniqueness constraint upon commit, causing the transaction to fail.

It is probably harmless because the probability of this happening is roughly (number of concurrent submissions)^2 / size of ID space. Still, it deserves a comment with this justification.

gollux:
My original proposal was to have the opaque IDs scoped to the given participation, while you have them globally unique.

I still think that a more restricted scope (at least a contest if not a participation) is better. One reason is that it reduces the probability of collisions, the other that it allows preserving the IDs in contest dumps without fear of collisions when importing. Also, when a dump is upgraded from an older format, it is sufficient to avoid collisions in newly assigned IDs.

gollux:
> Just to cover all bases: how bad is it really to leak information on the total number of submissions?

That's not a healthy approach to security :) Contestants are supposed to be separated, so all leaks are bad to some degree.

Of course there are some leaks that are almost unavoidable. One such example is evaluation delays. We can think about them and argue that they give no real advantage to contestants.

But when the number of leaks grows, you must argue that even a combination of all the leaks gives no advantage, which is fiendishly hard. So whenever a leak is easy to stop, we want to do it, so that the few remaining unavoidable leaks will be manageable.

wil93:
OK so you think it gives some kind of advantage to be able to see that information? That's fair.

As for this point made by @veluca93:

> I am opposed to using a function of the submission ID, because it introduces the possibility of mistakes leaking information.

What kind of mistakes do you mean? If I understood correctly, the "function of the submission ID" approach would re-use the secret key configured in cms.toml and to some reversible hash, for which there is probably some standard function out there (likely already among our dependencies).

gollux:
> OK so you think it gives some kind of advantage to be able to see that information? That's fair.

I am not sure that it gives. But I am sure that contestants are much more ingenious than me, so I prefer minimizing leaks to quantifying advantages obtainable from the leaks :)

cms/db/submission.py
veluca93: Added a comment.

cmscontrib/updaters/update_45.py
veluca93: Fixed.

cmscontrib/updaters/update_from_1.5.sql
```diff
@@ -14,4 +14,9 @@ ALTER TABLE public.contests ALTER COLUMN allow_unofficial_submission_before_anal
 -- https://github.com/cms-dev/cms/pull/1393
 ALTER TABLE public.submission_results ADD COLUMN scored_at timestamp without time zone;
 
+-- https://github.com/cms-dev/cms/pull/1419
+ALTER TABLE submissions ADD COLUMN opaque_id BIGINT;
+UPDATE submissions SET opaque_id = id WHERE opaque_id IS NULL;
+ALTER TABLE submissions ADD CONSTRAINT participation_opaque_unique UNIQUE (participation_id, opaque_id);
```
wil93, line 20: I checked the diff and I think we need this at the end:

```
ALTER TABLE submissions ALTER COLUMN opaque_id SET NOT NULL;
```

cmscontrib/updaters/update_45.py
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env python3
+
+# Contest Management System - http://cms-dev.github.io/
+# Copyright © 2014 Fabian Gundlach <320pointsguy@gmail.com>
```
gollux: Really? ;)

cms/db/submission.py
```diff
@@ -177,6 +187,25 @@ def tokened(self) -> bool:
...
 
+    @classmethod
+    def generate_opaque_id(cls, session, participation_id):
+        randint_upper_bound = 2**63-1
+
+        opaque_id = random.randint(0, randint_upper_bound)
+
+        # Note that in theory this may cause the transaction to fail by
+        # generating a non-actually-unique ID. This is however extremely
+        # unlikely (prob. ~num_parallel_submissions_per_contestant^2/2^64).
```
gollux: I know that the estimate is from me, but it's wrong. It should be `/2^63`. I didn't realize that the IDs are signed.

cmscontrib/updaters/update_45.py
veluca93: Heh, indeed not...

cms/db/submission.py
veluca93: Didn't think of that either...

PR #1420, opened 2025-06-19
Add updater to adapt CMS's schema from v1.5

wil93:


cmscontrib/updaters/update_from_1.5.sql
```diff
@@ -0,0 +1,17 @@
+BEGIN;
```
gollux, line 1: Is it permitted to put schema changes within a transaction? I think that `DROP` _anything_ is not and I am not sure about other changes.

cmscontrib/updaters/update_from_1.5.sql
```diff
@@ -0,0 +1,17 @@
...
+
+-- https://github.com/cms-dev/cms/pull/1378
+ALTER TYPE public.feedback_level ADD VALUE 'oi_restricted';
```
gollux, line 4: Do we want to use the `public.` schema prefix everywhere? It's the default and we don't use any other schemata.

gollux (commented):
Thanks for starting it!

cmscontrib/updaters/update_from_1.5.sql
wil93: I tried the script and it seemed to work. It seems that at least the `ALTER` statements work.

cmscontrib/updaters/update_from_1.5.sql
wil93: I did it like this mostly because I used the diff between the old schema and the new schema (dumped with `pg_dump --schema-only`) and this was included there. I guess it doesn't hurt to be more explicit.

cmscontrib/updaters/update_from_1.5.sql
gollux: It's possible that `ALTER`s work, but I am not sure that it's guaranteed that they are actually a part of the transaction. I would prefer not using transactions in updaters at all.

cmscontrib/updaters/update_from_1.5.sql
gollux: Well, it adds noise, which is always a burden for the reader. But it's probably an acceptable level of noise :)

PR #1421, opened 2025-06-19
Remove CODEOWNERS.

veluca93:
As we discussed, the consensus between active devs is that CODEOWNERS is not the best way for individual devs to be notified about specific PRs.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,0 @@
-# For things like schema changes, where we need to make sure that an updater is provided
-/cms/db     @wil93


wil93:
We discussed it but I don't think that we reached an agreement on it.

I don't see any harm in keeping this, so I opposed this change.

wil93:
I took a stab at implementing the Github action diff thing 😄  PTAL #1422

veluca93:
@wil93 friendly ping on @gollux's question

veluca93:
> As this setting does not influence anyone else's experience in contributing to CMS other than my own

I believe we said in the chat that we believe this is not true, as it creates an impression on contributors that you are the owner of the database-related parts of the codebase (that's literally what this file is for), and I don't believe there is agreement with the other people with write access that this is the case.

Consequently, I also disagree on this being a minor change.

> I don't see this as a strong enough reason to undo this now. There were other similar "meta" settings, one random example that comes to mind is the .vscode/ folder, that I introduced without requesting approval: I don't think we should remove that folder now just because it was not reviewed when it was added. 

The key difference is that nobody else with write access disagrees with the changes to the `.vscode` folder, while I believe from the discussion the agreement we reached with everyone else was that this is not a good change.

stefano-maggiolo:
William, I think the consensus for this PR is clear and it doesn't need long discussions.

A second pair of eyes looking at any commit is a standard practice. It is also reasonable that the reactions of not doing it can change depending on the nature of the change.

Let's move on from this topic towards something more productive, and try to use code reviews for all changes from now on. If anything, I don't think anyone will forget about updaters from now on!

stefano-maggiolo (approved):

PR #1422, opened 2025-06-19
Add Github action to automatically diff new schema against old schema with updater applied

wil93:


wil93:
It doesn't seem to work very well because pg_dump's output is not very stable :( 

image

Probably we should use something like https://github.com/stripe/pg-schema-diff or https://github.com/joncrlsn/pgdiff.

prandla:
order of columns is a genuine property of the schema. i'm not sure if it matters when using sqlalchemy but it does matter when doing e.g. `select *` from a table in the psql console. unfortunately it seems that reordering columns is not something that postgresql explicitly supports. so to get the order to match up, we would need to either always add new fields at the end of a table, or do some pretty horrible hacks in the migration script (dropping and recreating columns, and then fixing up all foreign keys) to get the columns in the right order.

gollux:
Maintaining consistent order of columns is too hard and removing `ENUM` values is plain impossible. So if we want an automated check of schema migration scripts, it must take this things into account.

PR #1423, opened 2025-06-23
Add documentation for the automated interaction API.

veluca93:
Also fix potential confusion due to JSON not roundtripping 64-bit integers correctly.

### DIFF
--- a/docs/API.rst
+++ b/docs/API.rst
@@ -1,0 +1,90 @@
+API for ContestWebServer
+************************
+
+Error handling
+==============
+
+Generally, API calls will return a non-200 status code on invalid usage. In
+that case, the body of the response *may* be a JSON object with an "error"
+field, containing additional information on the error.
+
+Authentication
+==============
+
+All the log-in methods and restrictions that apply to normal ContestWebServer
+usage also apply for API usage. In particular, this implies that IP autologin,
+if configured, can be used for API calls.
+
+Otherwise, making a `POST` request to the `/api/login` endpoint, passing
+`username` and `password` as form data (as is done for regular login), will
+return a JSON object with the following structure:
+
+.. sourcecode:: json
+
+  {"login_data": string}
+
+The content of the returned login data string should be passed to API calls
+that require authentication in a `X-CMS-Authorization` header.
+
+All other API methods described here require authentication.
+
+Task list
+=========
+
+An authenticated `GET` request to `/api/task_list`, done while the contest is
+running, will return the following object:
+
+.. sourcecode:: json
+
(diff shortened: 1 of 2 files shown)


docs/API.rst
```diff
@@ -0,0 +1,83 @@
...
+
+An authenticated `POST` request to `/api/{taskname}/submit` will send a
+submission. Files should be provided (according to the submission format) as
+form data, with field names matching the submission format.
+
+The request will return an object with the ID of the new submission:
+
+.. sourcecode:: json
+
+  {"id": number}
```
gollux: We should make it clear what is the range of the numbers.

Ah, I just realized a nasty trap we are falling into: JSON is generally not transparent wrt. 64-bit numbers, because many libraries interpret them as floats. Let us make the submission IDs in the API strings, even though their contents will be numeric for the time being.

docs/API.rst
```diff
@@ -0,0 +1,83 @@
...
+if configured, can be used for API calls.
+
+Otherwise, making a `POST` request to the `/api/login` endpoint, passing
+`username` and `password` as form data (as is done for regular login), will
+return a JSON object with the following structure:
+
+.. sourcecode:: json
+
+  {"login_data": string}
+
```
gollux, line 25: We should specify what happens on errors (preferably once for the whole API).

gollux (commented):
I agree except for a couple of nitpicks.

docs/API.rst
veluca93: Ugh, indeed. Fixed.

docs/API.rst
veluca93: Done.

gollux (changes requested):
Fine except that I don't like making API changes in the same commit with documentation changes. Could you please split it to make the history easier to read?

veluca93:
> Fine except that I don't like making API changes in the same commit with documentation changes. Could you please split it to make the history easier to read?

Done.

docs/API.rst
```diff
@@ -0,0 +1,90 @@
...
+===================
+
+An authenticated `GET` request to `/tasks/{taskname}/submissions/{id}` will
+retrieve information on the submission with the given `id`. In particular,
+its field `public_score` will contain the score of this submission, and
+`task_public_score` will contain the current score for the task.
+
+Additional details on the submission's results can be retrieved by making an
+authenticated `GET` request to `/tasks/{taskname}/submissions/{id}/details`.
+The endpoint will return an HTML snipped matching what is seen by contestants.
```
prandla: typo: snipped -> snippet

gollux (approved):

PR #1426, opened 2025-06-26
Installation

gollux:


pyproject.toml
```diff
@@ -1,3 +1,73 @@
...
+[project.optional-dependencies]
+devel = [
+    # Only for testing
+    "beautifulsoup4>=4.8,<4.14",
+    "coverage>=4.5,<7.9",
+    "pytest",
+    "pytest-cov",
+
+    # Only for building documentation
+    "Sphinx>=1.8,<1.9",
```
prandla: our documentation can't actually be built with this sphinx version, and we can't specify the proper version (>=8.3 or something i think) because that conflicts with our old jinja2 version. so it'd probably be better to remove it from here for now.

install.py
```diff
@@ -0,0 +1,223 @@
...
+    progress("Installing Python dependencies")
+    subprocess.run(
+        [str(target_path / 'bin/pip'), 'install', '-r', 'constraints.txt'],
+        check=True)
+
+
+def install_systemd(args: Namespace) -> None:
+    progress("Installing systemd services")
+    source_path = Path('config/systemd')
+    dest_path = Path('~/.config/systemd/user').expanduser()
```
prandla, line 161: IMO it would make more sense for these to be systemd system services with User=cmsuser and Group=cmsuser. AFAIK that's the "usual" way to package these kinds of services. when they are user services, they can only be started when the user has logged in, which might complicate server setup. but of course, installing the service files system-wide requires root permissions...

cmsranking/Config.py
```diff
@@ -60,18 +60,24 @@ def __init__(self):
...
-        self.lib_dir = os.path.join("/", "var", "local", "lib", "cms", "ranking")
-        self.conf_paths = [
-            os.path.join("/", "usr", "local", "etc", "cms_ranking.toml"),
-            os.path.join("/", "etc", "cms_ranking.toml"),
-        ]
+
+        # Try to find CMS installation root from the venv in which we run
+        self.base_dir = sys.prefix
+        if self.base_dir == '/usr':
+            # This looks like a system-wide installation, so try backward-compatible paths
```
prandla: We (estonian informatics olympiad) currently use a venv while still keeping config in /usr/local, so this detection would fail for us

pyproject.toml
gollux: OK, I will comment it out.

install.py
gollux: > but of course, installing the service files system-wide requires root permissions...

That's one of my reasons for using user services. The other is that it's easy to run multiple instances of CMS on the same machine this way.

> Iwhen they are user services, they can only be started when the user has logged in, which might complicate server setup.

This is solved by `loginctl enable-linger <user>`. I should remember to mention it in the docs.

cmsranking/Config.py
gollux: I expect that everybody installs CMS in their own unique way :) Which is why I started writing my installation script.

I wonder how much backward compatibility we need. Do we want to ask everybody to migrate to the new setup? Perhaps provide a migration script? Or show how to add explicit paths to the old directories to `cms.conf`?

prandla: eh, people need to migrate from cms.conf to cms.toml manually anyways, i don't think relocating the directories is too much of a problem. we should mention it (the ability to set explicit paths) in the docs though.

gollux:
I updated the documentation and most bits of the container machinery.

Dockerfile
```diff
@@ -1,59 +1,74 @@
...
+WORKDIR /home/cmsuser/src
 
-RUN sudo pip3 install --break-system-packages -r requirements.txt
-RUN sudo pip3 install --break-system-packages -r dev-requirements.txt
+RUN ./install.py venv
 
-COPY --chown=cmsuser:cmsuser . /home/cmsuser/cms
+RUN \
+    echo >>~/.bash_profile 'source ~/.bashrc' && \
+    echo >>~/.bashrc 'source ~/cms/bin/activate'
```
prandla: you can also achieve this with `ENV VIRTUAL_ENV=/home/cmsuser/cms`, a little cleaner imo
edit: wait no, need to update PATH too. still a bit better than mucking with bashrc i think

docs/Docker image.rst
```diff
@@ -97,7 +97,7 @@ similar to:
 The command will build a fresh CMS image when necessary, and drop you into a
 bash prompt where the repository is mounted on ``~/cms`` for ease of
```
prandla: this is ~/src now

Dockerfile
prandla: i just noticed that CI is currently failing because it's not finding any of the cms commands, so it seems that putting it in bashrc is in fact not enough. so apart from feeling cleaner, doing it with ENV also works better :)

i derped a bit earlier; PATH is the more critical variable here, VIRTUAL_ENV is just to let some other tools (other than python/pip) know that a venv is active, it's not strictly necessary to set it but it might be nice.

pyproject.toml
```diff
@@ -1,3 +1,75 @@
+[project]
+dynamic = ["version"]
```
prandla:
```suggestion
dynamic = ["version", "entry-points", "scripts"]
```
needed to make pip install the console entry_points in setup.py.

docker-compose.test.yml
prandla, line 1: you missed one of the CMS_CONFIG paths in this file.

docker-compose.dev.yml
```diff
@@ -11,10 +11,10 @@ services:
     depends_on:
       - "devdb"
     environment:
-      CMS_CONFIG: /usr/local/etc/cms-devdb.toml
+      CMS_CONFIG: /home/cmsuser/cms/etc/cms-devdb.toml
     volumes:
       - "./.dev/home:/home/cmsuser"
```
prandla: this mount conflicts with /home/cmsuser/cms.

docker-compose.test.yml
gollux: Thanks, fixing.

pyproject.toml
gollux: Thanks, fixed.

docs/Docker image.rst
gollux: Fixed, thanks.

Dockerfile
gollux: Sorry, CI still needs a couple of bug fixes.

`ENV` is problematic, because we need to _add_ to `PATH`, not to replace it.

prandla: yes, you can do `ENV PATH="/home/cmsuser/bin:$PATH"`.

gollux: This does something slightly different than what I wanted (which was to extend the `PATH` as provided by the OS). But I see that the default Ubuntu image forces a `PATH` anyway. Weird, but we can live with that easily.

gollux:
All tests pass. I will give it more testing tomorrow.

AUTHORS.txt
wil93, line 1: Nit: Luca is now mentioned twice :) probably good to deduplicate.

docker-compose.dev.yml
```diff
@@ -1,6 +1,6 @@
 services:
   devdb:
-    image: postgres:15
+    image: docker.io/postgres:15
```
wil93, line 3: Is this a different Image or the same? If it's different then what is the difference?

gollux: It's a more systematic name of the same image.

AUTHORS.txt
gollux: Will do.

docs/Installation.rst
```diff
@@ -49,180 +61,125 @@ All dependencies can be installed automatically on most Linux distributions.
...
+Alternatively, if you like living at the bleeding edge, check out the CMS
+`Git repository <https://github.com/cms-dev/cms>`_ instead.
 
-        deactivate
+The preferred method of installation is using :samp:`./install.py --dir={target} cms`,
+which does the following:
 
-Method 2: Using the Docker image
---------------------------------
+* Creates an *target directory* of the given name. It contains a Python
```
prandla: typo: an -> a

docs/Installation.rst
```diff
@@ -49,180 +61,125 @@ All dependencies can be installed automatically on most Linux distributions.
...
---------------------------------------
+* Checks that Isolate is available.
 
-There are good reasons to install CMS and its Python dependencies via pip
-instead of your package manager (e.g. apt-get). For example: two different Linux
-distro (or two different versions of the same distro) may offer two different
-versions of ``python-sqlalchemy``. When using pip, you can choose to install a
-*specific version* of ``sqlalchemy`` that is known to work correctly with CMS.
+* Installs the sample configuration files to :samp:`{target}/etc/cms.toml`
+  and :samp:`TARGET/etc/cms_ranking.toml`.
```
prandla: this should use either `{target}` or `TARGET` in both cases.

prandla (commented):
do we have anything that keeps constraints.txt up to date with the requirement specs in pyproject.toml? currently you still have sphinx and all its deps in constraints.txt even though we agreed to remove it from pyproject.toml.

docs/Installation.rst
gollux: Fixed.

docs/Installation.rst
gollux: Ah, one survived. Thanks.

prandla (approved):

gollux:
@prandla: Thanks for your review!

Before I remove the draft status, I will clean up commit history.

prandla:
i just remembered one more thing: we can put the setuptools<81 constraint in pyproject.toml too (in the build-system section). it's not too important though; i was planning to fix the thing with pkg_resources soon anyways.

cms/conf.py
```diff
@@ -174,41 +173,50 @@ def __init__(self):
...
-        self.run_dir = os.path.join("/", "var", "local", "run", "cms")
-        paths = [
-            os.path.join("/", "usr", "local", "etc", "cms.toml"),
-            os.path.join("/", "etc", "cms.toml"),
-        ]
+        # Try to find CMS installation root from the venv in which we run
+        self.base_dir = sys.prefix
+        if self.base_dir == '/usr':
+            # This looks like a system-wide installation, so try backward-compatible paths
+            # FIXME: Do we want to support this?
```
veluca93: I have a mild preference for not supporting this, and making things simpler.

cmsranking/Config.py
```diff
@@ -60,18 +60,24 @@ def __init__(self):
...
+            # FIXME: Do we want to support this?
```
veluca93: Ditto :-)

cmsranking/Config.py
```diff
@@ -132,7 +138,9 @@ def _load_many(self, conf_paths: list[str]) -> bool:
                                 conf_path, errno.errorcode[error.errno],
                                 os.strerror(error.errno))
                 return False
-        logger.warning("No config file found, using hardcoded defaults.")
+        tried = " ".join(conf_paths)
+        logging.warning(f"No configuration file found, tried: {tried}. "
+                        "Falling back to default values.")
```
veluca93: At some point we discussed this being a hard failure instead of a warning.

IMO it's a good idea as the default values are almost never what you want (either for CMS or for RWS), but it might be worth doing separately.

docs/Installation.rst
```diff
@@ -49,180 +61,125 @@ All dependencies can be installed automatically on most Linux distributions.
...
+        python python-pip postgresql-libs libcups libyaml
 
     # Install the following from AUR.
-    # https://aur.archlinux.org/packages/libcgroup/
     # https://aur.archlinux.org/packages/cppreference/
 
-    # Only if you are going to use pip/venv to install python dependencies
-    sudo pacman -S --needed postgresql-libs libcups libyaml python-pip
+    # Install Isolate manually, it's not packaged for Arch yet:
+    # https://github.com/cms/isolate/
```
veluca93: https://aur.archlinux.org/packages/isolate ?

veluca93 (approved):

cms/conf.py
gollux: OK. I removed support for running outside a venv. I simplified reading of configuration, made a missing configuration file always fatal and removed the "Using configuration file..." message upon startup.

cmsranking/Config.py
gollux: Ditto :-)

cmsranking/Config.py
gollux: Agreed.

docs/Installation.rst
gollux: Oops, I missed it. Added.

gollux:
> Regarding my earlier comment about setuptools<81: we still have users of pkg_resources in our dependency tree (jinja2 at the very least, seems that werkzeug can use it too). So i think we should still move the version pin into pyproject.toml. Then it should be possible to remove the reference to setuptools from install.py altogether; pip should be able to figure out that it's needed and install the correct version itself.

I tried that, but I could not get it working.

When the venv is created, the system version of setuptools is copied there. This makes pip satisfied, even if this version does not satisfy the version requirements in `pyproject.toml`.

For the time being, I maintain setuptools version requirements in both `install.sh` and `pyproject.toml`.

prandla:
actually, a thought just occurred to me: what happens if you put the version restriction in both build-system.requires and project.dependencies? because we care about pkg_resources at runtime, so it'd make sense to specify it along with other runtime dependencies.

gollux:
I'm not sure if it's a good idea to install the packages with a version of setuptools different from the one in runtime.

gollux:
I cleaned up branch history. Unless somebody protests, I am going to merge it soon.

prandla:
> I'm not sure if it's a good idea to install the packages with a version of setuptools different from the one in runtime.

this is what is already happening though, no?

if you notice, the pip install output contains "installing build dependencies...". as i understand it, pip is installing them into a separate venv which is only used for building the wheel. nothing guarantees that the version of setuptools here is the same as the version that's in the venv that will run the package. (though, if both of them are the version bundled with pip, then they would end up being equal after all.)

i think it might actually be worth it to pin the exact version of setuptools, in both project.dependencies and build-system, if we want to ensure that an up-to-date one is always used.

cmsranking/Config.py
```diff
@@ -108,56 +107,39 @@ def load(self, config_override_fobj=None):
...
+        # Load config file.
         try:
-            data = tomllib.load(conf_fobj)
-        except ValueError:
-            logger.critical("Config file is invalid TOML.")
+            with open(path, 'rb') as f:
+                data = tomllib.load(f)
+        except FileNotFoundError:
+            logger.debug("Couldn't find config file %s (maybe you need to "
+                         "convert it from cms_ranking.conf to cms_ranking.toml?).", path)
```
prandla: very minor nit: the old name was `cms.ranking.conf`, not `cms_ranking.conf`

pyproject.toml
```diff
@@ -1,3 +1,75 @@
...
+    "Development Status :: 5 - Production/Stable",
+    "Natural Language :: English",
+    "Operating System :: POSIX :: Linux",
+    "Programming Language :: Python :: 3.11",
+]
+
+dependencies = [
+    "tornado==4.5.3",         # http://www.tornadoweb.org/en/stable/releases.html
+    "psycopg2==2.9.10",       # http://initd.org/psycopg/articles/tag/release/
+    "sqlalchemy>=1.3,<1.4",   # http://docs.sqlalchemy.org/en/latest/changelog/index.html
```
wil93, line 20: I'm thinking we could pin all dependencies to an exact version, using `==` everywhere. What do you think?

gollux: I don't think we should. The dependencies in `pyproject.toml` should be bounds on what is believed to work, while `constraints.txt` should tell what we really tested and want to install CMS with.

Also, I am not sure how useful is the stream of PRs churned out by dependabot.

cmsranking/Config.py
gollux: Indeed. Fixing.

gollux:
@prandla: Now I agree that listing setuptools as both build and runtime dependency is right. Applied.

gollux:
I rebased on the top of the current main (which involved resolving a couple of conflict and I'm going to merge.

Issue #1427, opened 2025-06-28
pkg_resources are obsolete


gollux:
The `pkg_resources` module is declared obsolete in recent versions of `setuptools` and it is slated for removal. See https://setuptools.pypa.io/en/latest/pkg_resources.html for more details.

CMS uses `pkg_resources` at multiple places. We should replace it by using `importlib` appropriately.

PR #1428, opened 2025-06-30
Replace the obsolete pkg_resource module and replace it with importlib

pxsit:
Fix : https://github.com/cms-dev/cms/issues/1427


cms/plugin.py
```diff
@@ -49,14 +49,17 @@ def plugin_list(entry_point_group: str) -> list[type]:
 
     """
     classes = []
-    for entry_point in pkg_resources.iter_entry_points(entry_point_group):
+    for entry_point in entry_points(group=entry_point_group):
         try:
             classes.append(entry_point.load())
-        except (pkg_resources.UnknownExtra, ImportError):
+        except ImportError:
```
prandla: This should catch more than just ImportError; e.g. if someone typos the class name in their setup.py then an AttributeError will be thrown instead. I think it'd be fine to use `except Exception` here.

cmsranking/Config.py
```diff
@@ -59,7 +59,12 @@ def __init__(self):
         # Buffers
         self.buffer_size = 100  # Needs to be strictly positive.
 
-        self.web_dir = pkg_resources.resource_filename("cmsranking", "static")
+        try:
+            self.web_dir = str(resources.files("cmsranking") / "static")
+        except AttributeError:
+            # Fallback incase of errors
+            with resources.path("cmsranking", "static") as static_path:
```
prandla: this context manager will delete the temporary directory two lines later. we need it to be longer-lived; or refactor RankingWebServer to use importlib.resources properly and not require creating a temporary directory at all.

cms/locale/locale.py
```diff
@@ -282,13 +282,24 @@ def get_translations() -> dict[str, Translation]:
...
+                mofile_path = lang_dir / "LC_MESSAGES" / "cms.mo"
+                if mofile_path.is_file():
+                    try:
+                        with mofile_path.open("rb") as f:
+                            t = Translation(lang_code, f)
+                            logger.info("Found translation %s", t.identifier)
+                            result[t.identifier] = t
+                    except Exception as e:
+                        logger.warning(
+                            "Failed to load translation for %s: %s", lang_code, e
```
prandla: you can use `logger.warning("...", exc_info=True)` instead of manually formatting the exception.

cms/locale/locale.py
```diff
@@ -41,7 +41,7 @@
 import babel.numbers
 import babel.support
 import babel.units
-import pkg_resources
+from importlib import resources
```
prandla: i'd prefer leaving this import as just `import importlib.resources`.

cms/locale/locale.py
```diff
@@ -282,13 +282,24 @@ def get_translations() -> dict[str, Translation]:
...
-                logger.info("Found translation %s", t.identifier)
-                result[t.identifier] = t
+    try:
+        locale_pkg = resources.files("cms.locale")
+        for lang_dir in locale_pkg.iterdir():
+            if lang_dir.is_dir():
+                lang_code = lang_dir.name
+                mofile_path = lang_dir / "LC_MESSAGES" / "cms.mo"
+                if mofile_path.is_file():
+                    try:
```
prandla: if you already have a try-except here, then i think it'd make sense to remove the `if` check and just log the warning if opening it fails. (also, you can then move the mofile_path into the try-except. from reading importlib source code, it seems that it can sometimes throw when the path doesn't exist.)

prandla (changes requested):

cms/locale/locale.py
```diff
@@ -282,13 +282,25 @@ def get_translations() -> dict[str, Translation]:
...
+                        logger.info("Found translation %s", t.identifier)
+                        result[t.identifier] = t
+                except Exception:
+                    logger.warning(
+                        "Failed to load translation for %s",
+                        lang_code,
+                        exc_info=True,
+                    )
+    except Exception as e:
+        logger.warning("Failed to scan locale directory: %s", e)
```
prandla: same comment about exc_info=True.

cmsranking/Config.py
```diff
@@ -161,3 +163,64 @@ def _load_one(self, conf_fobj: typing.BinaryIO) -> bool:
                 return False
             setattr(self, key, value)
         return True
+
+    def _extract_static_files(self) -> str:
```
prandla: this feels like way too much code for this task. the simpler solution is to just keep the context manager from importlib.resources.path around longer.

cmsranking/Config.py
```diff
@@ -23,7 +23,7 @@
 import tomllib
 import typing
 
-import pkg_resources
+from importlib import resources
```
prandla: same as earlier, i'd rather `import importlib.resources` here.

cmsranking/Config.py
```diff
@@ -59,7 +59,9 @@ def __init__(self):
         # Buffers
         self.buffer_size = 100  # Needs to be strictly positive.
 
-        self.web_dir = pkg_resources.resource_filename("cmsranking", "static")
+        # Keep the static files context manager alive for the application lifetime
+        self._static_files_context = resources.path("cmsranking", "static")
+        self.web_dir = str(self._static_files_context.__enter__())
```
prandla, line 65: almost correct: we still need to call `__exit__` on the context manager at some point. I think an atexit handler might be the simplest way here.

cmsranking/Config.py
```diff
@@ -161,3 +163,5 @@ def _load_one(self, conf_fobj: typing.BinaryIO) -> bool:
                 return False
             setattr(self, key, value)
         return True
+
+    # ...existing code...
```
prandla: what is this comment doing here?

prandla (changes requested):

cmsranking/Config.py
pxsit: Is it ok yet? Github notifications just notified me (why is it so delayed).

PR #1430, opened 2025-06-30
sanitize compiler output more

### LINKED ISSUE #888: Wrong upload bug
New CMS (cf4cdf0039ba45d21058790cf19cf8516a55adb3 commit) failed to handle the wrong java submission which led to the EvaluationService crash due to the null insertion into the DB. 
[Skladisce.txt](https://github.com/cms-dev/cms/files/1804248/Skladisce.txt) This was the .java file (I renamed it to upload it here). We haven't made a further investigation whether a contestant submited this file or it was corrupted by the CMS.

prandla:
notably, this removes null bytes from compiler output, should they end up there. this is necessary because postgres doesn't allow storing null bytes in text columns, so ES would previously error when trying to store the compilation output in the DB. (additionally, it's a bit nicer to the contestants, as control characters aren't rendered in an useful way by browsers anyways.)

Fixes #888 (the other half of that issue, null bytes in checker/manager output, was already handled in #1396, seemingly without even noticing #888.)


gollux:
Is there any reason for not replacing `0x0b` (VT) and `0x0c` (FF) control characters? Otherwise OK.

veluca93 (approved):

PR #1436, opened 2025-07-02
Support BatchAndOutput task type in the importers

birka0:
Based on: #1330 

`tps`: Since TPS does not handle testcase metadata very well (e.g., codenames are generated arbitrarily and developers might unintentionally change them), explicitly listing output-only testcases in a config file is not a good idea.
Instead, we can optionally mark subtasks as collections of output-only testcases.
Note that this approach aligns with IOI rules, where output-only is considered a subtask property rather than a testcase-specific one.

`italy_yaml`: Simply copied the logic used for handling `public_testcases`.

### DIFF
--- a/cmscontrib/loaders/tps.py
+++ b/cmscontrib/loaders/tps.py
@@ -197,13 +197,21 @@ class TpsTaskLoader(TaskLoader):
                 for filename in os.listdir(testcases_dir)
                 if filename[-3:] == '.in'])
         if data["task_type"] == 'BatchAndOutput':
-            output_only_testcases = {}
+            output_only_testcases = set()
+            output_optional_testcases = set()
             for cur_subtask_data in subtask_data.values():
                 is_output_only = cur_subtask_data.get('output_only', False)
+                is_output_optional = cur_subtask_data.get('output_optional', False)
                 if is_output_only:
                     codenames = cur_subtask_data.get('testcases', list())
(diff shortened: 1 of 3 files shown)


cmscontrib/loaders/tps.py
```diff
@@ -181,12 +196,26 @@ def get_task(self, get_statement=True):
...
+        # Setting the submission format
         if data["task_type"] == 'OutputOnly':
             args["submission_format"] = list()
             for codename in testcase_codenames:
                 args["submission_format"].append("output_%s.txt" % codename)
         elif data["task_type"] == 'Notice':
             args["submission_format"] = list()
+        elif data["task_type"] == "BatchAndOutput":
+            args["submission_format"] = ["%s.%%l" % name]
+            for codename in data["output_only_testcases"]:
```
veluca93: In principle, BatchAndOutput allows one to configure testcases for which the output *can* be provided, but doesn't *have* to.
This code wouldn't support that.

I imagine this is fine, but just wanted to point it out (same for italy_yaml)

veluca93 (approved):

cmscontrib/loaders/tps.py
birka0: Good catch.. I overlooked that feature. I think we should support it - PTAL at the new commit.

prandla (approved):

PR #1438, opened 2025-07-03
Docker fixes

prandla:
* dropdb would always print `dropdb: error: database removal failed: ERROR:  database "cmsdbfortesting" does not exist` at the start of testing. Silenced it using the `--if-exists` flag.
* Removed `docker/` and `!docker/_cms-*` from dockerignore. It doesn't strictly hurt anything to include these scripts in the docker image, and it fixes the issues the wildcard was causing with .dev/postgres-data.
* The codecov mount was in the wrong place now that the docker-compose files were moved to a subdirectory.


gollux (approved):
Thanks for fixing it!

PR #1447, opened 2025-07-06
Make API login succeed if the user would already be logged in.

### LINKED ISSUE #1445: IP-based logins via API
I am trying to make ioisubmit work with the new API. I got stuck at authentication :)

I am using an account with IP-based authentication enabled. When I try using the submit endpoint, the request is authenticated successfully, but then it's rejected because of missing XSRF protection. When I try to log in first to get an authentication token, it fails too, because I haven't supplied the user name (ioisubmit has no knowledge of users).

I can make it work by sending `X-CMS-Authorization` with a dummy value, but I don't think it's the right solution :)

I think that the right fix is making the login endpoint accept IP-based auth without a login name.

veluca93:
Fixes #1445

### DIFF
--- a/cms/server/contest/handlers/api.py
+++ b/cms/server/contest/handlers/api.py
@@ -45,7 +45,22 @@ class ApiLoginHandler(ContestHandler):
     """
     @multi_contest
     def post(self):
+        current_user = self.get_current_user()
+
         username = self.get_argument("username", "")
+
+        if current_user is not None:
+            if username != "" and current_user.user.username != username:
+                self.json(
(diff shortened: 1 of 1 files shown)


cms/server/contest/handlers/api.py
```diff
@@ -45,7 +45,23 @@ class ApiLoginHandler(ContestHandler):
     """
     @multi_contest
     def post(self):
+
```
gollux: Nitpicking: An extra empty line.

gollux (approved):
Aside from too many empty lines, I approve.

PR #1448, opened 2025-07-07
Fix API errors if the contest is in a wrong phase

gollux:
Previously, the CWS just returned HTML.


veluca93 (approved):

prandla:
Shouldn't we also be using something other than `@tornado_web.authenticated` for these handlers then? I think that just redirects to the login url currently...

PR #1449, opened 2025-07-07
CWS API: Report login errors properly

gollux:
Previously, HTML with the login page was returned.

### DIFF
--- a/cms/server/contest/handlers/api.py
+++ b/cms/server/contest/handlers/api.py
@@ -20,13 +20,11 @@
 
 """
 
+from collections.abc import Callable
+import functools
 import ipaddress
 import logging
-
-try:
-    import tornado4.web as tornado_web
-except ImportError:
-    import tornado.web as tornado_web
+import typing
 
 from cms.db.submission import Submission
 from cms.server import multi_contest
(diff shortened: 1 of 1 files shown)


cms/server/contest/handlers/api.py
```diff
@@ -20,8 +20,11 @@
...
 """
 
+from collections.abc import Callable
+import functools
 import ipaddress
 import logging
+import typing
 
 try:
     import tornado4.web as tornado_web
```
prandla: this import is unused now

gollux: It's not clear what import you are referring to :) Do you mean `tornado_web`?

prandla: yes.

cms/server/contest/handlers/api.py
```diff
@@ -49,6 +52,27 @@ def __init__(self, *args, **kwargs):
...
+    func: Callable[typing.Concatenate[_Self, _P], _R],
+) -> Callable[typing.Concatenate[_Self, _P], _R | None]:
+    """A decorator filtering out unauthenticated requests.
+
+    """
+
+    @functools.wraps(func)
+    def wrapped(self: _Self, *args: _P.args, **kwargs: _P.kwargs):
+        if not self.current_user:
+            self.json({"error": "An authenticated user required"}, 403)
```
veluca93: nit: is required

cms/server/contest/handlers/api.py
gollux: OK, will remove.

cms/server/contest/handlers/api.py
gollux: OK :)

PR #1454, opened 2025-07-07
CWS: Impersonation of users by administrators

gollux:
This is a proof of concept of an extension of API authentication, which makes it possible to impersonate an arbitrary user if one knows a secret admin token (set in the configuration file). Impersonation will also make it possible to ignore certain restrictions (contest phase, time between submits).

I would like to use this for offline submissions made by the `ioisubmit` tool.

At this moment, I would like to know if you consider this approach reasonable before I finish all the details. (Please look only at the whole diff, individual commits are currently a tangled web of messy fixups.)

### DIFF
--- a/cms/server/contest/authentication.py
+++ b/cms/server/contest/authentication.py
@@ -71,6 +71,7 @@ def validate_login(
     username: str,
     password: str,
     ip_address: AnyIPAddress,
+    admin_token: str = ""
 ) -> tuple[Participation | None, bytes | None]:
     """Authenticate a user logging in, with username and password.
 
@@ -91,6 +92,7 @@ def validate_login(
     username: the username the user provided.
     password: the password the user provided.
     ip_address: the IP address the request came from.
+    admin_token: administrator's token used to impersonate a user
 
     return: if the user couldn't
         be authenticated then return None, otherwise return the
@@ -103,7 +105,7 @@ def validate_login(
                     "%r, on contest %s, at %s: " + msg, ip_address,
                     username, contest.name, timestamp, *args)
 
-    if not contest.allow_password_authentication:
+    if not contest.allow_password_authentication and admin_token == "":
         log_failed_attempt("password authentication not allowed")
         return None, None
 
@@ -119,6 +121,20 @@ def validate_login(
     if participation is None:
         log_failed_attempt("user not registered to contest")
         return None, None
+
+    if admin_token != "":
+        if (config.contest_admin_token is not None
+            and admin_token != config.contest_admin_token):
+            log_failed_attempt("invalid admin token")
+            return None, None
+
+        logger.info("Successful impersonated login from IP address %s, as user %r, on "
+                    "contest %s, at %s", ip_address, username, contest.name,
+                    timestamp)
+
+        return (participation,
+                json.dumps([username, "", make_timestamp(timestamp), True])
+                    .encode("utf-8"))
 
     correct_password = get_password(participation)
 
@@ -151,7 +167,7 @@ def validate_login(
     # If hashing is used, the cookie stores the hashed password so that
     # the expensive bcrypt call doesn't need to be done at every request.
     return (participation,
-            json.dumps([username, correct_password, make_timestamp(timestamp)])
+            json.dumps([username, correct_password, make_timestamp(timestamp), False])
                 .encode("utf-8"))
 
 
@@ -166,7 +182,7 @@ def authenticate_request(
     cookie: bytes | None,
     authorization_header: bytes | None,
     ip_address: AnyIPAddress,
-) -> tuple[Participation | None, bytes | None]:
+) -> tuple[Participation | None, bytes | None, bool]:
     """Authenticate a user returning to the site, with a cookie.
 
     Given the information the user's browser provided (the cookie) and
@@ -200,16 +216,17 @@ def authenticate_request(
     timestamp: the date and the time of the request.
     cookie: the cookie the user's browser provided in the
         request (if any).
+    authorization_header: the value of X-CMS-Authorization header (if any).
     ip_address: the IP address the request
         came from.
 
-    return: if the user
-        couldn't be authenticated then return None, otherwise return
-        the participation that they wanted to authenticate as; if a
-        cookie has to be set return it as well, otherwise return None.
+    return: a tuple consisting of participation (None if authentication failed),
+        a cookie that has to be set (or None), and a boolean flag indicating
+        whether the admin token was used to impersonate a user.
 
     """
     participation: Participation | None = None
+    impersonated = False
 
     if contest.ip_autologin:
         try:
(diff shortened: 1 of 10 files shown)


prandla:
> I like the idea, but I wonder whether it wouldn't make more sense to use an admin username + password instead of a config-global secret.

currently, admin login data is only used on AWS. technically this means that leaking admin credentials is not a problem if AWS itself is blocked off from contestants/the public internet (via IP restrictions or additional authentication in a reverse proxy). With this change, admin credentials become useful with just CWS access.

I guess we could just tell people to to guard their credentials better, but this is something we should consider.

gollux:
I think that high-entropy tokens are better for API authentication than (often low-entropy) passwords. In particular, passwords require strong hashing, which is slow.

Later, I would like to switch the API to tokens stored in the database. The login endpoint would return a token for a user identified by username + password. Other endpoints would be authenticated by a token (obtained either from the login endpoint or by another channel like looking it up in the AWS). The same system of tokens could be used for admin tokens in CWS and a future API of the AWS.

For now, I would prefer a simple approach with admin tokens in the configuration file and not so intrusive changes in the CWS.

veluca93:
> I think that high-entropy tokens are better for API authentication than (often low-entropy) passwords. In particular, passwords require strong hashing, which is slow.
> 
> Later, I would like to switch the API to tokens stored in the database. The login endpoint would return a token for a user identified by username + password. Other endpoints would be authenticated by a token (obtained either from the login endpoint or by another channel like looking it up in the AWS). The same system of tokens could be used for admin tokens in CWS and a future API of the AWS.
> 
> For now, I would prefer a simple approach with admin tokens in the configuration file and not so intrusive changes in the CWS.

Ok.

cms/server/contest/authentication.py
```diff
@@ -200,16 +216,17 @@ def authenticate_request(
...
     ip_address: the IP address the request
         came from.
 
-    return: if the user
-        couldn't be authenticated then return None, otherwise return
-        the participation that they wanted to authenticate as; if a
-        cookie has to be set return it as well, otherwise return None.
+    return: a tuple consisting of participation (None if authentication failed),
+        a cookie that has to be set (or None), and a boolean flag indicating
+        that the admin token was used to impersonate a user.
```
veluca93: nit: "whether"

cms/server/contest/authentication.py
```diff
@@ -323,26 +343,31 @@ def _authenticate_request_from_cookie_or_authorization_header(
...
         last_update = make_datetime(cookie[2])
+        impersonated: bool = cookie[3]
     except Exception as e:
         # Cookies are stored securely and thus cannot be tampered with:
         # this is either a programming or a configuration error.
         logger.warning("Invalid cookie (%s): %s", e, cookie)
-        return None, None
+        return None, None, False
+
+    # Reject if password authentication is disabled and it's not an impersonation cookie.
```
veluca93: or authentication header, right?

cms/server/contest/submission/workflow.py
```diff
@@ -105,6 +107,8 @@ def accept_submission(
     official: whether the submission was sent in during a regular
         contest phase (and should be counted towards the score/rank) or
         during the analysis mode.
+    force_max_number: skip checks for the maximum number of submissions
+    force_min_interval: skip checks for the minimum interval between submissions
```
veluca93: The name seems a bit off for me, maybe `override_` would be better than `force_` here?

cms/server/contest/phase_management.py
```diff
@@ -220,9 +220,14 @@ def decorator(
     ) -> Callable[typing.Concatenate[_Self, _P], _R | None]:
         @wraps(func)
         def wrapped(self: _Self, *args: _P.args, **kwargs: _P.kwargs):
-            if self.r_params["actual_phase"] not in actual_phases and \
-                    (self.current_user is None or
-                     not self.current_user.unrestricted):
+            unrestricted = self.current_user is not None and self.current_user.unrestricted
+            if self.impersonated_by_admin:
+                try:
+                    unrestricted = self.get_boolean_argument("force_phase", unrestricted)
```
veluca93: naming: `ignore_phase`? `force_phase` sounds like it is trying to replace the phase with a different one.

docs/API.rst
```diff
@@ -21,7 +21,7 @@ return a JSON object with the following structure:
 
 .. sourcecode:: json
 
-  {"login_data": string}
+  {"login_data": "string"}
```
prandla, line 24: I don't like using double-quotes here, that would imply that this is the literal string "string". maybe `/* string */` would be okay? (pygments' json highlighter should support comments.)

cms/server/contest/handlers/api.py
```diff
@@ -26,6 +26,11 @@
 import logging
 import typing
 
+try:
+    import tornado4.web as tornado_web
```
prandla: this import looks unused

docs/API.rst
```diff
@@ -88,3 +88,35 @@ its field `public_score` will contain the score of this submission, and
 Additional details on the submission's results can be retrieved by making an
 authenticated `GET` request to `/tasks/{taskname}/submissions/{id}/details`.
 The endpoint will return an HTML snippet matching what is seen by contestants.
+
+Impersonation of users
+======================
+
+Administrators may impersonate a user and perform requests on their behalf.
+
+This is accomplished by using the authentication endpoint without `password`
```
prandla, line 97: we should be using ``` `` ``` everywhere in this file instead of `` ` `` imho

prandla (approved):

cms/server/contest/authentication.py
gollux: OK, fixing.

cms/server/contest/authentication.py
gollux: Fixing.

cms/server/contest/submission/workflow.py
gollux: I agree.

cms/server/contest/phase_management.py
gollux: I liked that all override options have a common prefix. I made it `override_phase_check`.

cms/server/contest/handlers/api.py
gollux: Ah, too much moving of things between files back and forth ... fixed.

docs/API.rst
gollux: Probably yes, but let's keep it a separate PR.

docs/API.rst
gollux: I don't think that it's better with a comment - it perhaps makes better sense, but it's harder to read as the whole line is stretched too much. Let us try to look for a better solution in a separate PR. I am keeping this commit because it improved the situation slightly (no more weird red boxes around letters).

Issue #1242, opened 2023-10-08
Avoid bind-mounting temporary dirs within sandbox


gollux:
I'm not sure if this is a bug report or a feature request :-)

Currently, CMS puts files in the sandbox by creating a temporary directory, initializing a sandbox, and bind-mounting the directory inside the sandbox as `/tmp`. This is not the approach recommended by Isolate's documentation and it leads to multiple problems with file owners and permissions, which might have security implications.

Isolate is started by `cmsuser` (the user running the worker services), but the process inside the sandbox runs on its own UID (each sandbox has its own). The CMS's temporary directory is owned by `cmsuser` and writable by everybody. It allows the sandbox user write in this directory. However, if the sandbox user creates a sub-directory not writable by everybody, CMS is then unable to remove its contents, leaving the temporary directory in `/tmp` forever. Also, the temporary directory is writable by all other users of the system.

The approach recommended by Isolate's documentation is to use the `/box` directory inside the sandbox. When the sandbox is created, this directory is owned by the caller.  When `isolate --run` is called, the owner of all files inside `/box` is changed to the sandbox user, and once the sandboxed process finishes, the owner is changed back. This avoids all of the permission problems mentioned above.

I would very much recommend changing the way Isolate uses the sandbox to the recommended one. See also #1005.

Also, when we are at it, we should replace setting the file size limit by proper filesystem quotas (#916).

prandla:
I tried implementing this in our fork [here](https://github.com/e-i-o/cms/commit/33b299467408883b7cfca834939fbf947f12d2b7). However, I think this is not exactly ideal: as the box directory is now owned by the isolate user, they can always create new files in it or change the permissions on other files to override the `allow_writing_only` mechanism (currently, for batch tasks, the submission is only allowed to write to the stdout and stderr files). For now i just made all the `allow_writing_*` functions empty stubs. I think this means proper filesystem quotas are in fact necessary to prevent resource exhaustion. Also, this is more of a problem with my current implementation, but `cleanup` fails when the submission chmod's some of its files such that `shutil.copytree()` is unable to read them, I guess we would need to chmod them back beforehand.

gollux:
Thanks for your try!

I don't think that `allow_writing_only` makes any sense. Even in the current master, there exist places where the solution can write to. And running solutions without a proper filesystem quota was never secure.

[...]

PR #1456, opened 2025-07-08
"Rerun and archive sandbox" button in AWS

### LINKED ISSUE #1355: "Rerun and archive sandbox" button in AWS
Our fork has a feature where a specific sandbox can be reran and saved as a zip file, for easier access.

Image

Here the first testcase was reran and archived, the "1" links to a zip file that contains the entire sandbox contents.

This allows having keep_sandboxes = false in the cms config, while still allowing debugging individual submissions. it's also much easier to access the sandbox directly from AWS, instead of having to ssh into a worker and go spelunking around the filesystem. (in our case, the workers are also in docker, which further complicates finding the right sandbox...)

prandla:
Closes #1355.

This still needs a dump updater and SQL updater (because i changed the "sandbox paths" column from a :-separated string to an array). Other parts should be ready for review though.

I changed the UI a bit from the old implementation:
image
sandbox links are clickable when archived. There's also a similar button for the compile sandbox.

There's also most of the logic to archive user test sandboxes, but there's no RPC to invalidate them, so it can't actually be reached currently...

### DIFF
--- a/cms/grading/Sandbox.py
+++ b/cms/grading/Sandbox.py
@@ -542,21 +542,21 @@ class SandboxBase(metaclass=ABCMeta):
         """
         logger.info("Archiving sandbox in %s.", self.get_root_path())
 
-        # Archive the working directory
-        sandbox_archive_filename = "sandbox.zip"
-        sandbox_archive = self.relative_path(sandbox_archive_filename)
-        content_path = self.get_root_path()
-        with zipfile.ZipFile(sandbox_archive, "w") as zip_file:
-            for root, dirs, files in os.walk(content_path):
-                if sandbox_archive_filename in files:
-                    files.remove(sandbox_archive_filename)
-                zip_file.write(root, os.path.relpath(root, content_path))
-                for f in files:
-                    f_path = os.path.join(root, f)
-                    zip_file.write(f_path, os.path.relpath(f_path, content_path))
+        with tempfile.TemporaryFile(dir=self.temp_dir) as sandbox_archive:
+            # Archive the working directory
+            content_path = self.get_root_path()
+            with zipfile.ZipFile(sandbox_archive, "w",
+                                 compression=zipfile.ZIP_DEFLATED,
+                                 compresslevel=9) as zip_file:
+                for root, dirs, files in os.walk(content_path):
+                    zip_file.write(root, os.path.relpath(root, content_path))
+                    for f in files:
+                        f_path = os.path.join(root, f)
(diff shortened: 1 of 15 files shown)


cms/grading/Sandbox.py
```diff
@@ -532,10 +533,31 @@ def cleanup(self, delete: bool = False):
 
         delete: if True, also delete get_root_path() and everything it
             contains.
-
         """
         pass
 
+    def archive(self):
```
veluca93: Couldn't this cause potential issues with malicious solutions? I.e. solutions that create a symlink to places `cmsuser` can read but the solution cannot.

Since the archive is not given to the contestant, I imagine it's not a security issue, but causing a DoS with this is probably easy...

prandla: You're right. `os.walk` doesn't follow symlinks, but `ZipFile.write` does, and there's no option to turn it off, short of manually constructing `ZipInfo` objects, which involves manually setting the (undocumented???) external file attributes field correctly to capture the unix permissions. (and also need to figure out how to encode the symlink itself, which is even less documented)

I love python sometimes...

maybe we should use tarfile here instead, that one is at least capable of reproducing an unix filesystem properly. but i don't know how well windows supports those.

gollux: I think we should use tarfile. If grading fails due to a file with strange permissions produced by the solution, we want to see it in the archived sandbox. On Windows, both Total Commander and 7-Zip handle tar files pretty well.

prandla: Changed to .tar.gz now.

gollux:
Beware that some files in the sandbox could be unreadable for the user running CMS.

prandla:
i think it's only possible for compilation (for evaluation, allow_writing_only should prevent it i think?), but it is indeed a possibility.

i think the proper way to handle this is to implement #1242, and then implement an archiver that chmods files as it goes to make them readable. (this then requires some trickery to get the metadata in the archive right.) we would also need similar chmodding before deleting the sandbox (or just letting isolate deal with it).

for now, i guess i can add an extra check and make sure that the case where creating the archive fails is at most a warning.

gollux:
> For now, i guess i can add an extra check and make sure that the case where creating the archive fails is at most a warning.

I agree.

Otherwise, I think that `allow_writing_only` is a terrible hack and I am looking forward for the moment when we remove it. But we really should wait with #1242 for after the IOI :)

gollux (approved):

gollux:
> @gollux does this last patch look okay? :)

Yes.

PR #1457, opened 2025-07-08
don't install testlib.h

### LINKED ISSUE #1379: Installation of testlib.h
Do we really want to install polygon.h with CMS? It seems to be very specific to a single task preparation system. Installation of this header should be managed by Polygon, not by CMS.

In particular, do we want polygon.h to be accessible to solutions?

prandla:
closes #1379.

Admittedly I think the original problem (testlib.h being accessible to solutions) was already fixed by the new installation method (now that it's forced to install in a venv, the header is in venv/include, which shouldn't be in the isolate sandbox), but doing it like this feels cleaner anyways.

### DIFF
--- a/cmscontrib/loaders/polygon.py
+++ b/cmscontrib/loaders/polygon.py
@@ -22,8 +22,12 @@
 
 import logging
 import os
+import shutil
 import subprocess
 import xml.etree.ElementTree as ET
+import importlib.resources
+import importlib.util
+import tempfile
 from datetime import datetime, timedelta
 
 from cms import config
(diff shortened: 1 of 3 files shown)


cmscontrib/loaders/polygon.py
```diff
@@ -181,20 +184,26 @@ def get_task(self, get_statement=True):
...
+                    # package with our patched version. Since the package
+                    # includes a testlib.h too, the easiest way to achieve this
+                    # is to copy the checker source to a temporary directory.
+                    testlib_res = importlib.resources.files("cmscontrib.loaders").joinpath("polygon/testlib.h")
+                    with testlib_res.open('rb') as src, open(tempdir + "/testlib.h", "wb") as dst:
+                        shutil.copyfileobj(src, dst)
+                    new_checker_src = tempdir + '/check.cpp'
+                    output_path = tempdir + '/check'
+                    shutil.copyfile(checker_src, new_checker_src)
+                    code = subprocess.call(["g++", "-x", "c++", "-O2",
```
veluca93, line 198: How do people usually include testlib.h? AFAICT this won't work if `#include <testlib.h>`, but the previous version would have.

prandla: i just checked and polygon itself doesn't support that, so i think it's fine to not support it here.

PR #1461, opened 2025-07-10
Documentation copy-editing + some fixes

prandla:
* use ``` `` ``` in places where bare `` ` `` were used previously (in most cases i think ``` `` ``` was what was intended originally anyways)
* fix missing references and other sphinx errors
* change json code-blocks to use comments to indicate the types of things (open to suggestions on how to format this better)
* updated some docs that referred to unsupported installation methods

todo:
* i think i saw a dead link on one page.
* i'd like to use `{install_dir}` instead of `{target}` to refer to the installation directory. (i already did it in the places i touched, but should change it everywhere for consistency)

### DIFF
--- a/cms/db/__init__.py
+++ b/cms/db/__init__.py
@@ -81,7 +81,7 @@ __all__ = [
 
 # Instantiate or import these objects.
 
-version = 45
+version = 46
 
 engine = create_engine(config.database, echo=config.database_debug,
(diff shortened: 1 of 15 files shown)


pxsit:
Task type line 201 distutils link is dead
Replace with https://docs.python.org/3.11/distutils/setupscript.html

gollux:
OK, let's call it _installation directory._

PR #1462, opened 2025-07-12
AWS: Implement diffing of solutions

### LINKED ISSUE #1408: Allow diffing solutions in AWS
This is something I want to implement soon. I'm not sure about how to make a good UI for selecting the two submissions to diff though. It ought to be possible to select two submissions from different pages of the paginated submissions list, I'm not sure how to implement that...

I'm going to say this is blocked on #1351 because once we have prism.js, we can get [really nice syntax highlighting on the diffs](https://prismjs.com/plugins/diff-highlight/). (I have the code for 1351 written already, just need to make a PR...)

prandla:
Closes #1408.

Screenshots:
image

The "diff submissions" details element is closed by default. When it's closed, the radio buttons on each row are hidden:
image

(I used an html `<details>` element instead of our custom "toggling" class, because i needed an event listener for when it gets opened/closed and this was the easiest way to implement that. It doesn't have an open/close animation though.)

Using the radio buttons just fills in the submission ID inputs, you can also enter submission IDs manually. The diff appears in the same popup as submission code, after clicking "Diff".

image

It's possible to compare almost any two solutions: if they have different languages, there is simply no syntax highlighting. If they are for different tasks, it's possible to compare them if both tasks expect a single source file.  I'm not sure how useful this latter part is, but we have seen people submit code for the wrong task occasionally.

The implementation uses python's builtin `difflib.unified_diff` to compare files (and uses prism.js to syntax-highlight the diff). difflib also contains an HTML diff, but I don't like its output styling, it isn't easy to syntax-highlight, and it always outputs a two-column diff (i prefer unified). It is better at showing changes within lines though, and it shows line numbers nicely.

Overall I think this is the best approach (for rendering the diff itself) without spending ages implementing a custom solution.

I'm not sure the submission selection UI is the best possible though, I'm open to suggestions for improvements.


veluca93 (approved):
Seems reasonable enough to me. Main question is whether comparing two large outputonly outputs risks being very slow...

pxsit (approved):
Looks Good!

Issue #1372, opened 2025-06-06
[IOI fork parity] cli script changes


prandla:
* https://github.com/ioi-2024/cms/commit/9819be6ae2a1b44e87e16983ed23b2f213cdc168 "add option '--name' to ImportDataset"
* https://github.com/ioi-2024/cms/commit/c4bbefcde368470c08c8ac411285fa0ab4258f67 "Add -d/--default flag in cmsAddStatement to indicate is the statement is the default one"
* https://github.com/ioi/cms/commit/66bd28170a69d931927f5c0ef47ca905aacd2685 "cmsImportTask: Add -d/--dataset flag to indicate which dataset to upsert"

and some changes to cmsAddSubmission related to the ioisubmit tool:
* https://github.com/ioi-2024/cms/commit/acb638203dc87b5b80ec614971907fc3d6a1d09e "add the --lang parameter"
* https://github.com/ioi-2024/cms/commit/5979ed89b72da6d0c342bbaf58580195f4365361 "temporary hack for connecting to the ES"
* https://github.com/ioi-2024/cms/commit/ac1e42f4f85eae0942e0d0d11ee9c4a1d766d515 "check for submission limits"

the last two of which need to be implemented in a less hacky way.

wil93:
My two cents:

- Adding lang parameter: we could disambiguate with an extra prompt (similar to what we do when calling `cmsContestWebServer` without specifying a contest ID when there are multiple contests). However I think it's fine to have the lang parameter too (it could be useful when calling AddSubmission as part of a script or a loop).
- Hack for ES: as I said before, we could replace the RPC system and simplify our architecture by using an existing job queue mechanism.
- Check submission limits: I think AddSubmission is a strictly admin-facing script that is **not** intended to be run by users. Why are we checking the submission limits? If we want to have a user-facing script, then it would anyway be better to do it as an interface to CWS (the only user-visible part of CMS) via REST API or something like that, and in that case the limits would be checked implicitly.

gollux:
> Using the new CWS API would IMO be better, if there aren't strong reasons why it wouldn't work - wdyt?

It could be better if we figure out how to do the submission on behalf of a given contestant.

veluca93:
If nothing else, we can log in as them at the end of the contest, right?

gollux:
We would have to authenticate as the particular contestant (which probably means to keep a list of passwords and turn off IP-based restrictions) and somehow flag the submissions as official even though they are made after the contest ends.

[...]

PR #1464, opened 2025-07-14
cmsAddSubmission: Add the --lang parameter

gollux:
When multiple versions of C++ are enabled in a contest, it is impossible to guess the language by filename extension.

It is not checked that the language is really enabled in the contest. We consider this a feature (and document it as such), because it allows the administrator to test languages before they are made available to the contestants.

See issue #1372.

### DIFF
--- a/cms/grading/languagemanager.py
+++ b/cms/grading/languagemanager.py
@@ -50,7 +50,7 @@ def get_language(name: str) -> Language:
     return _BY_NAME[name]
 
 
-def filename_to_language(filename: str) -> Language | None:
+def filename_to_language(filename: str, available_languages: list[Language] | None=None) -> Language | None:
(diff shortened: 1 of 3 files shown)


gollux:
Sorry, this turned out to be a bigger problem: `cmsAddSubmission` without `--lang` defaults to the _first_ language on the list which matches the filename extension, regardless of what is enabled in the contest. The test suite did not originally catch this, because it used C which has only one version enabled.

All callers of  `cms.grading.languagemanager.filename_to_language` seem to be affected, in particular `cmsMake` and `cmstestsuite/web/CWSRequests.py`.

I will return to this PR later and fix it properly.

veluca93:
FYI, I am happy to drop cmsMake, there are better alternatives nowadays.

veluca93 (approved):

PR #1471, opened 2025-07-22
Use isolate's cg-oom-killed to detect "memory limit exceeded" conditions

### LINKED ISSUE #912: Use isolate's cg-oom-killed
It was introduced here: https://github.com/ioi/isolate/commit/cd542d637aec5f4d707e8d8e26b3cf465637a487

prandla:
Closes #912.

I deleted the extra text from killed-by-signal's help text because it's not relevant there anymore (I hope). I didn't re-add it to the new error's help text because in my experience, it's not true: in all cases I have tried, the oom killer has been precise enough that the reported memory use is exactly the task's memory limit (up to the rounding precision, at least). The behaviour that was described in that deleted paragraph might have been possible with non-cgroups isolate? I'm not sure.

Also, PHP has a `memory_limit` setting that limits the amount of memory the interpreter even tries to allocate; this limit is 128M by default. Exceeding this causes PHP to exit with code 255 (and print a message on stdout). This broke the new tests (and in fact, this could have been caught earlier: all the other oom tests were killed by signal, whereas php was nonzero exit code, but this fact was not checked by the test). Actually, this was a bigger problem, in that PHP enforced a 128MB memory limit, ignoring whatever the task's memory limit was. This would be possible to fix in php.ini on worker machines, but the contest administrators would have to be aware of it. I fixed it by changing the limit to -1 (meaning "infinite") using a command-line argument.

### DIFF
--- a/cms/grading/Sandbox.py
+++ b/cms/grading/Sandbox.py
@@ -188,6 +188,7 @@ class SandboxBase(metaclass=ABCMeta):
     EXIT_SIGNAL = 'signal'
     EXIT_TIMEOUT = 'timeout'
     EXIT_TIMEOUT_WALL = 'wall timeout'
+    EXIT_MEM_LIMIT = 'memory limit exceeded'
     EXIT_NONZERO_RETURN = 'nonzero return'
 
     def __init__(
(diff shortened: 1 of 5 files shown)


gollux (approved):
I agree, but please wait with merging until the end of this IOI.

pxsit (approved):

PR #1476, opened 2025-07-23
Add a Score counter to task overview and sidebar

pxsit:
Add a Score counter to task overview
image


Muaath5:
> I have some comments about the code, but i'm wondering if this is the UI we want in the first place... I saw some browser extension that showed the task scores on the sidebar next to the task links, maybe that would be better.
> 
> Does anyone else have any thoughts?

I don't see any issues if both of these ways where implemented, but if it was in the task overview it would be better in case of onsite competitions and you don't want others to look into your score :)

pxsit:
> > I have some comments about the code, but i'm wondering if this is the UI we want in the first place... I saw some browser extension that showed the task scores on the sidebar next to the task links, maybe that would be better.
> > Does anyone else have any thoughts?
> 
> I don't see any issues if both of these ways where implemented, but if it was in the task overview it would be better in case of onsite competitions and you don't want others to look into your score :)

Added an option to toggle this in AWS (:

cms/server/contest/handlers/main.py
```diff
@@ -76,8 +79,60 @@ class MainHandler(ContestHandler):
...
     def get(self):
+        self.r_params = self.render_params()
         self.render("overview.html", **self.r_params)
 
+    def render_params(self):
+        ret = super().render_params()
+
+        if self.current_user is not None:
+            # This massive joined load gets all the information which we will need
+            participation = self.sql_session.query(Participation)\
```
prandla: please reformat this with ruff

cms/server/contest/handlers/main.py
```diff
@@ -76,8 +79,60 @@ class MainHandler(ContestHandler):
...
+                .filter(Participation.id == self.current_user.id)\
+                .options(
+                    joinedload('user'),
```
prandla: you should use the column objects here instead of strings, like `Participation.user`

cms/server/contest/handlers/main.py
```diff
@@ -76,8 +79,60 @@ class MainHandler(ContestHandler):
...
+                task_scores = {}
+                for task in self.contest.tasks:
+                    score_type = task.active_dataset.score_type_object
+                    max_public_score = round(
+                        score_type.max_public_score, task.score_precision
+                    )
+                    public_score, _ = task_score(
+                        participation, task, public=True, rounded=True
+                    )
+                    public_score = round(public_score, task.score_precision)
```
prandla: isn't this unnecessary, since you already pass round=True to task_score?

cms/server/contest/static/cws_style.css
```diff
@@ -559,27 +559,33 @@ td.token_rules p:last-child {
     color: #AAA;
 }
 
-.submission_list td.public_score.score_0 {
+.submission_list td.public_score.score_0,
+.main_task_list td.public_score.score_0 {
```
prandla, line 578: i think you can just remove the first class from the condition here, instead of duplicating the conditions

cmscontrib/updaters/update_from_1.5.sql
```diff
@@ -42,4 +42,7 @@ ALTER TABLE user_test_results ADD COLUMN evaluation_sandbox_digests VARCHAR[];
 UPDATE user_test_results SET evaluation_sandbox_paths = string_to_array(evaluation_sandbox, ':');
 ALTER TABLE user_test_results DROP COLUMN evaluation_sandbox;
 
+-- https://github.com/cms-dev/cms/pull/1476
+ALTER TABLE contests ADD COLUMN show_task_scores_in_overview boolean NOT NULL DEFAULT true;
```
prandla, line 46: needs a `ALTER TABLE contests ALTER COLUMN ... DROP DEFAULT` after this (to make the sql schema match what sqlalchemy generates exactly)

cms/server/contest/templates/overview.html
```diff
@@ -208,6 +211,9 @@ <h2>{% trans %}Task overview{% endtrans %}</h2>
 {% set extensions = "[%s]"|format(contest.languages|map("to_language")|map(attribute="source_extension")|unique|join("|")) %}
 {% for t_iter in contest.tasks %}
         <tr>
+{% if contest.show_task_scores_in_overview and task_scores is defined %}
+            <td class="public_score {% if task_scores[t_iter.id][0] <= 0 %}score_0{% elif task_scores[t_iter.id][0] >= task_scores[t_iter.id][1] %}score_100{% else %}score_0_100{% endif %}">{{ task_scores[t_iter.id][2] }}</td>
```
prandla: you should use the get_score_class jinja function here

prandla (changes requested):

cms/server/contest/templates/overview.html
```diff
@@ -208,6 +211,9 @@ <h2>{% trans %}Task overview{% endtrans %}</h2>
 {% set extensions = "[%s]"|format(contest.languages|map("to_language")|map(attribute="source_extension")|unique|join("|")) %}
 {% for t_iter in contest.tasks %}
         <tr>
+{% if contest.show_task_scores_in_overview and task_scores is defined %}
+            <td class="public_score {{ get_score_class(task_scores[t_iter.id][0], task_scores[t_iter.id][1]) }}">{{ task_scores[t_iter.id][2] }}</td>
```
prandla, line 220: Please test your changes locally before pushing next time. This crashes, `get_score_class` (the python function, not the javascript function with the same name...) takes 3 arguments, the 3rd one is the score precision.

cms/server/contest/static/cws_style.css
```diff
@@ -559,27 +559,27 @@ td.token_rules p:last-child {
     color: #AAA;
 }
 
-.submission_list td.public_score.score_0 {
+.main_task_list td.public_score.score_0 {
```
prandla, line 578: You broke the coloring in the submission list of a single task with this. I meant you should change the entire line to just `td.public_score.score_0 {`.

prandla (changes requested):

cms/server/contest/handlers/main.py
```diff
@@ -381,7 +438,7 @@ def get(self):
         language_docs = []
         if config.docs_path is not None:
             for language in languages:
-                ext = language.source_extensions[0][1:]  # remove dot
+                ext = language.source_extensions[0][1:] # remove dot
```
prandla: sorry, i didn't notice this earlier: please undo this formatting change. 1) it's unrelated to this PR (in fact, it causes a conflict with another pending PR) and 2) it's not what `ruff format` does.

prandla (approved):

cms/server/contest/handlers/main.py
```diff
@@ -74,8 +78,62 @@ class MainHandler(ContestHandler):
...
+                .options(
+                    joinedload(Participation.user),
+                    joinedload(Participation.contest),
+                    joinedload(Participation.submissions).joinedload(Submission.token),
+                    joinedload(Participation.submissions).joinedload(
+                        Submission.results
+                    ),
+                )
+                .first()
+            )
```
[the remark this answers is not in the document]
pxsit: This would worsen the performance as lazy-loading all the data will be way worse than fetching the single user row twice.

PR #1477, opened 2025-07-23
Use github actions cache to cache docker layers

prandla:
This could save up to 2 minutes on each actions run (unless there's a new Ubuntu image, or we update our requirements).

I don't expect this to work first try :), but this stuff is hard to test locally.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,9 +15,17 @@ jobs:
         run: |
           mount | grep cgroup
 
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
       - name: Build docker image
-        run: |
-          docker compose -p cms -f docker/docker-compose.test.yml build testcms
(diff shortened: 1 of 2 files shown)


prandla:
....oh, it fails because apparently building like this somehow automatically excludes .git, which makes the functional test runner not find the root of the repo, which makes it try to start the services with the wrong path. (I don't even know why this path doesn't work; all the services should be available in $PATH.)

In any case, this is probably fixable (and imo should be fixed, including .git in docker just to find the repo root is complete overkill), but it demonstrates the time saved on a hot rebuild: it takes 1m 46s just to load the cached layers when cms code changes. This is versus the ~2m 30s that building the image from scratch takes. And the clean build is now over 2x slower than previously (5m 11s here).

It seems that somehow, copying around the layers inside the actions runner is so slow that it almost completely nullifies the point of having a cache at all. So I'm considering this a failed experiment. I'll try to look into other ways to optimize the runtime.

PR #1481, opened 2025-07-27
New config parser

prandla:
I wrote a custom parser that turns the output from tomllib into a strongly-typed dataclass-based config object. (This is basically the same as what pydantic does, but it's much smaller. We really don't need to bring in such a big dependency for this, as was pointed out by Martin.)

I also reorganized the config file format into sections, which are used in both CMS code and in the config file. This means the PR is quite huge though... The main things to review are the .sample.toml files, cms/conf.py, and cmscommon/conf_parser.py.

I also:
* added all undocumented options into cms.sample.toml and cms_ranking.sample.toml
* removed other_services from the config and replaced core_services with just services
* removed the deprecated is_proxy_used option. People need to port their configs anyways, so they can replace the deprecated option with the new one.
* removed the undocumented and unused timeout parameter from the RWS config.
* removed the ability to set web_dir in RWS config. I can't see any reason why anyone would need to change that.

This is an even more substantial change than the json->toml one. I'd like to get this in before we release 1.6, to avoid people having to port their configs twice (or well, the ones who were using the git version need to do it twice anyways, but that's unavoidable).

Maybe we should also provide a script that auto-converts json to the new config format.


cms/conf.py
```diff
@@ -21,236 +21,201 @@
...
+class AWSConfig:
+    listen_address: str = "127.0.0.1"
+    listen_port: int = 8889
+    cookie_duration: int = 10 * 60 * 60  # 10 hours
+    num_proxies_used: int = 0
+
+
+@dataclass()
+class ProxyServiceConfig:
+    # TODO do we want this default baked in?
```
veluca93: I'd say we don't. It can be in the sample config...

cms/conf.py
```diff
@@ -21,236 +21,201 @@
...
+@dataclass()
+class PrometheusConfig:
+    listen_address: str = "127.0.0.1"
+    listen_port: int = 8811
+
+
+@dataclass()
+class TelegramBotConfig:
+    bot_token: str | None = None
+    chat_id: str | None = None
```
veluca93: These really ought to be None at the same time, could you make this not None and the parent field a TelegramBotConfig | None?

config/cms.sample.toml
```diff
@@ -1,208 +1,223 @@
...
-# Sandbox #
-###########
 
-# Do not allow contestants' solutions to write files bigger
-# than this size (expressed in KB; defaults to 1 GB).
+[sandbox]
+# Which sandbox implementation to use. Currently only isolate is
+# supported.
+sandbox_implementation = "isolate"
+# Whether to use cgroups in isolate.
```
veluca93: Not even sure why we have this flag, is "false" something we want to support at all?

prandla: Last time I checked, isolate's manpage and source code both still mentioned a non-cgroups mode. But @gollux probably knows better...

gollux: Non-cgroup mode is still supported by isolate, but I don't it's needed/useful in CMS. Let's kill the option.

PR #1485, opened 2025-07-28
Remove Pascal from default language

pxsit:
Pascal is not really popular and barely any people use it anymore


veluca93 (commented):
I would maybe go even a bit further and make C++ the only default language.

Issue #1155, opened 2020-05-01
Per-task allowed languages


CristianCantoro:
**Is your feature request related to a problem?**
I would like the possibility to limit allowed languages (a subset of the languages allowed at contest level) per task.

We use CMS to assign a set of problems (tasks) to students and for most of them we are happy with solutions written either using C and C++,  and C++ is recommended. However. for some tasks (approximation problems), that require a custom grader, we have only developed a C++ grader, thus we would like to be able to restrict the allowed languages for that task to just C++.
  
**Describe the solution you'd like**

At the moment, in the _Contest configuration_ page, you can specify which languages are allowed.

I would like to have a similar checkbox, and an option in the task configuration files, to specify a subset of the allowed languages that can be used for a specific task.

**Describe alternatives you've considered**
An alternative would limit the allowed language to C+++ for the whole contest, but we would like to avoid that.

**Additional context**

Screenshot of the checkboxes marking the allowed languages in the _Contest configuration_ page:

![Screenshot of the checkboxes marking the allowed languages in the Contest configuration page](https://i.imgur.com/4RYSsUd.png),

PR #1486, opened 2025-07-28
Allow Per-task allowed languages

pxsit:
Closes https://github.com/cms-dev/cms/issues/1155

### DIFF
--- a/cms/db/task.py
+++ b/cms/db/task.py
@@ -417,8 +417,8 @@ class Dataset(Base):
         nullable=True)
     memory_limit: int | None = Column(
         BigInteger,
-        CheckConstraint("memory_limit > 0"),
-        CheckConstraint("MOD(memory_limit, 1048576) = 0"),
+        CheckConstraint("memory_limit > 0", name='datasets_memory_limit_check'),
+        CheckConstraint("MOD(memory_limit, 1048576) = 0", name='datasets_memory_limit_check1'),
         nullable=True)
 
     # Name of the TaskType child class suited for the task.
--- a/cmsranking/static/UserDetail.js
+++ b/cmsranking/static/UserDetail.js
@@ -153,7 +153,12 @@ var UserDetail = new function () {
         if (self.data_fetched == 2) {
             self.f_name_label.text(self.user["f_name"]);
             self.l_name_label.text(self.user["l_name"]);
-            self.face_image.attr("src", Config.get_face_url(self.user_id));
+
+            // Create a new image element and replace the old one
+            var new_face_image = $('<img id="UserDetail_face" alt="Face"/>');
+            new_face_image.attr("src", Config.get_face_url(self.user_id));
+            self.face_image.replaceWith(new_face_image);
+            self.face_image = new_face_image;
 
             if (self.user["team"]) {
                 self.team_label.text(DataStore.teams[self.user["team"]]["name"]);
--- a/cmstestsuite/unit_tests/schema_diff_test.py
+++ b/cmstestsuite/unit_tests/schema_diff_test.py
@@ -1,0 +1,168 @@
+import os
+import unittest
+import subprocess
+import difflib
+
+from cms.conf import config
+from cms.db.drop import drop_db
+from cms.db.init import init_db
+from cms.db.session import custom_psycopg2_connection
+
+"""
+Compare the DB schema obtained from upgrading an older version's database using
+an SQL updater, with the schema of a fresh install. These should be as close as
+possible, but there are a few quirks which means it's not possible for the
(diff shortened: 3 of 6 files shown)


cms/db/__init__.py
```diff
@@ -81,7 +81,7 @@
 
 # Instantiate or import these objects.
 
-version = 46
+version = 47
```
prandla: If you only add a column that has a default value, you don't need to bump the schema version. (And this also means you don't need the python updater.)

cms/server/contest/templates/task_submissions.html
```diff
@@ -6,6 +6,23 @@
...
+    '{{ lang }}': {
+{% for extension in (lang|to_language).source_extensions %}
+        '{{ extension }}': true,
+{% endfor %}
+    },
+{% endfor %}
+};
+
+// Override the global LANGUAGES with task-specific ones for this page
+LANGUAGES = TASK_LANGUAGES;
```
prandla: I'd rather you didn't change LANGUAGES here, I think it'd be better if the use sites were updated to use TASK_LANGUAGES instead.

cms/server/contest/templates/test_interface.html
```diff
@@ -2,6 +2,23 @@
 
 {% set page = "test_interface" %}
 
+{% block js_init %}
```
prandla, line 5: This block is duplicated between task_submissions and test_interface. It should probably be refactored out, but then again, there is a lot of duplication between submissions and user tests anyways, so maybe it's fine to leave it for now...

cmscontrib/updaters/update_47.sql
prandla, line 1: This sql should be put in update_from_1.5.sql instead.

cmscontrib/updaters/update_47.sql
```diff
@@ -0,0 +1,5 @@
+begin;
+
+alter table tasks add allowed_languages varchar[] not null default '{}';
```
prandla: The `default '{}'` is unnecessary here. SQLAlchemy doesn't generate default values in its table definitions.

(actually, it is necessary in this alter table command, because otherwise postgres doesn't know what to fill the existing rows with. so really, this line should be kept as is, and you should add a `alter table tasks alter column allowed_languages drop default;`.)

cms/server/contest/templates/task_submissions.html
```diff
@@ -262,7 +279,8 @@ <h2 style="margin-bottom: 10px">{% trans %}Submit a solution{% endtrans %}</h2>
                 <div class="control-group">
                     <div class="controls">
                         <select name="language">
-  {% for lang in contest.languages %}
+  {% set task_languages = task.allowed_languages if task.allowed_languages else contest.languages %}
```
prandla: I think it'd be cleaner to expose get_task_allowed_languages to the jinja templates and use it here.

Actually, the simplest way to do that might be to make get_allowed_languages a method of Task.

cms/db/task.py
```diff
@@ -116,6 +117,13 @@ class Task(Base):
         nullable=False,
         default=[])
 
+    # The list of names of programming languages allowed for this task.
+    # If empty, all contest languages are allowed.
```
prandla: I'm not sure I like the semantics of this, I think it'd make more sense to make this nullable and have null represent "all languages". But if anyone has strong opinions towards keeping it like this then i guess it's fine.

prandla (changes requested):
I see you also changed the UI of the "allowed languages" input in AWS. I think that should be moved into a separate PR, where we can make it consistent between the contest-wide and per-task language selection.

Also, the "compilation commands" list in the task description page should show only the enabled languages. Also, the "contest overview" table should show the correct file extensions in the "files" column.

prandla:
> I see you also changed the UI of the "allowed languages" input in AWS. I think that should be moved into a separate PR, where we can make it consistent between the contest-wide and per-task language selection.
> 
> Also, the "compilation commands" list in the task description page should show only the enabled languages. Also, the "contest overview" table should show the correct file extensions in the "files" column.

You still haven't addressed these comments.

cms/server/admin/handlers/contest.py
```diff
@@ -93,6 +93,28 @@ def post(self, contest_id: str):
 
             attrs["languages"] = self.get_arguments("languages")
 
+            # Auto-enable newly added languages in all tasks
```
prandla: I'm not sure we want this behavior. If a task has language restrictions, it's probably because there are only graders for specific languages; these graders aren't going to magically appear when adding a new language to the contest. (or do you have any use cases in mind where this behaviour is more useful?)

PR #1489, opened 2025-07-30
Update pytest config

prandla:
* Move settings into pyproject.toml (one less file in the repo root :)
* Set testpaths to avoid the "please monkey-patch earlier" warning from gevent (caused by pytest importing unnecessary files, including one that imported requests somewhere)

ps. feel free to ignore this until end of ioi if it makes your life easier in any way


wil93:
Can you push again to trigger the tests?

@pxsit I think it's also fine to create a separate PR to remove that file 🙂

wil93:
Actually I just bypassed the check, given that the tests were anyway already passing.

Issue #1205, opened 2022-05-11
Option to stop evaluation of a subtask on first 0 score. 


RezwanArefin01:
### The Problem
It is often a problem that CMS runs all files for each submission. In most OI styles problems, there is usually a subtask where very worse complexity solution can pass. Many participants usually code that first and submit. But since this is run on all the files, the later subtasks take up huge unnecessary judging time that it often jams up the server (low spec servers), especially in the beginning of the contest, sometimes even crash it (from experience)!   

### The Solution
For minimum in subtask scoring, stop evaluating more files once at least one of the file gets score 0. (This may also be related to feedback level full / restricted). 
It may not be the default, but having an option for this will be very helpful! 
Although it may be difficult to implement, since in CMS evaluation is done in parallel, in contrary to other online judges that stop on first wrong answer. 

### Alternative
Current way to overcome this problem is to just have small number of input files. Which can be a bad thing for some problems.

Issue #344, opened 2014-08-09
Evaluate unneeded testcases with low priority


artikz:
This one is connected with #342 and #258. In some contests it's not possible to use #258 due to traditions or some special rules (i.e. all testcases should always be evaluated). But we still can benefit from #258 if #342 is implemented: instead of skipping testcases like in #258 we just lower their priorities in the queue, so this testcases will be evaluated after everything other is done.

Issue #258, opened 2014-02-22
Stop worker after wrong answers


giomasce:
See http://www.freelists.org/post/contestms/Stopping-workers-after-wrong-answers. Basically the proposal is to make the workers at least partially score-aware, so that they can stop an evaluation after some testcases have failed, in order to save on resources.

This is difficult to do with our design. This issue is considered in very low priority.

PR #1496, opened 2025-08-05
Added a function to skip entire subtask when one testcase failed

pxsit:
Example
image


prandla:
See #258, #344, #1205 for some prior discussions.

Sorry, I don't think I like this implementation at all. First off, the logic for when to skip testcases belongs in the ScoreTypes themselves, not in EvaluationService (see Luca's comment in #258). This should allow removing most of the code in subtask_skipper, which I don't think really needs to be a separate module at all (parts should be in the ScoreTypes (or more likely just in ScoreTypeGroup), parts should be in EvaluationService, probably as just functions). Also, I don't like the UI, I think it would be simpler to show the skipped testcases the same way hidden testcases are shown with restricted feedback.

Also, as a general note, please try to avoid creating functions with 12 (!?) levels of indentation. And what's up with all the getattr()?

veluca93:
While I haven't read the code at all, one thing to keep in mind is that one
test case may belong to multiple subtasks.

So, we should be careful to not skip evaluating a testcase if another
testcase in the same subtask fails, as that testcase might also be present
in another subtask without the failing one.

PR #1502, opened 2025-08-06
Added Caching to workflows

pxsit:
Implement Docker layer caching
Instead of rebuilding the entire Docker image from scratch on every commit, the optimized workflow uses actions/cache to save and reuse the layers of your Docker image. On a new commit, Docker will only have to rebuild the layers that have actually changed, while reusing all the previous layers from the cache.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,8 +1,9 @@
 name: ci
 
 on:
-  - push
-  - pull_request
+  push:
+    branches: [ main, develop ]
+  pull_request:
 
 jobs:
   test:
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -1,8 +1,9 @@
 name: ci
 
 on:
-  - push
-  - pull_request
+  push:
+    branches: [ main, develop ]
```
gollux, line 5: Do we have a branch called `develop`?

.github/workflows/main.yml
```diff
@@ -13,16 +14,30 @@ jobs:
...
+      - name: Checkout code
+        uses: actions/checkout@v4
 
-      - name: Check available cgroups
-        run: |
-          mount | grep cgroup
+      - name: Cache Docker layers
+        uses: actions/cache@v4
+        with:
+          path: /tmp/.docker-cache
```
gollux, line 23: Is there any particular reason for naming the directory with a dot? Just `/tmp/docker-cache` would seem more straightforward.

PR #1509, opened 2025-08-12
Point to Weblate for translations

wil93:


prandla (approved):

pxsit:
Isn't there supposed to be a translation for the new Memory Limit Exceed?

prandla:
> Isn't there supposed to be a translation for the new Memory Limit Exceed?

That's just because we haven't updated cms.pot in the repo in a while.

I'm a bit more concerned that all the languages show up as "100% translated" even though they're obviously missing strings. Maybe this is because we haven't manually ran the update_catalog command?

PR #1512, opened 2025-08-16
Make cms.pot string order consistent, update .pot and .po files

prandla:
By default, `./setup.py extract_messages` will generate the .pot file in basically a random order (depending on the order of os.listdir, which returns the files in whichever order they are actually stored on disk). This means running extract_messages effectively shuffles the .pot file, generating an unnecessarily large diff and introducing merge conflicts.

To fix this, I added `sort_by_file: 1` to the extract_messages settings in setup.cfg, which sorts the strings in the .pot file by the filepath they are used in.

I then also ran extract_messages. Of course, because the order changed again, this commit once again does some arbitrary shuffling of the strings, but hopefully this is the last time it happens :)

I also ran update_catalog for all languages; hopefully this makes new strings show up on Weblate properly.

### DIFF
--- a/cms/locale/ar/LC_MESSAGES/cms.po
+++ b/cms/locale/ar/LC_MESSAGES/cms.po
@@ -1,33 +1,36 @@
+
 msgid ""
 msgstr ""
-"Project-Id-Version: VERSION\n"
-"POT-Creation-Date: 2025-06-06 09:35+0000\n"
+"Project-Id-Version:  VERSION\n"
+"Report-Msgid-Bugs-To: EMAIL@ADDRESS\n"
+"POT-Creation-Date: 2025-08-16 14:52+0300\n"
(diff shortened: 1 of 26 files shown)


veluca93 (approved):

prandla:
looks like a majority of the diffs are because our babel setup is configured to word-wrap at 79 chars (whereas the previous .po files from onesky were unwrapped). reading Weblate docs, it seems they have options for a width of 77 chars, or only wrapping at \n, or no wrapping at all. I think it'd make sense to configure our babel to emit the same format that weblate would use, as otherwise weblate might try to reformat the .po files (again causing unnecessarily large diffs).

So, I think we should probably just disable line wrapping altogether. It's not that useful for manual editing either; all text editors have word-wrap built in these days. @veluca93 what do you think?

veluca93:
I am fine with that.

PR #1530, opened 2025-08-26
Use Isolate's /box directory and filesystem quotas

### LINKED ISSUE #309: Sandbox does not set file size limit
The sandbox does not set a limit on the size of the files created by the solution. This can lead to various problems.
- Causing an out-of-memory situation for the worker, if the sandboxes are kept in a big enough tmpfs (which happens in a lot of distributions that mount a tmpfs in /tmp).

```
#include <unistd.h>
#include <sys/stat.h>
#include <fcntl.h>

#define BS 8192

int main() {
    int fd = open("output.txt", O_CREAT | O_WRONLY | O_TRUNC,
                                S_IRUSR | S_IWUSR | S_IRGRP | S_IROTH);
    char buf[BS] = {};
    while(1)
        write(fd, buf, BS);
}
```
- Allowing the user to somewhat bypass memory limits by creating a big file, then mmapping parts of it and using it as extra memory.
- Leaving a very big output.txt file, causing the worker to fail if there is no space left on the device, or to eat up a lot of memory (~4 times the size of the file). Using fallocate allows a program to allocate a lot of space in a very short time. For example, the following program

```
#include <unistd.h>
#include <sys/stat.h>
#include <fcntl.h>

int main() {
    int fd = open("output.txt", O_CREAT | O_WRONLY | O_TRUNC,
                                S_IRUSR | S_IWUSR | S_IRGRP | S_IROTH);
    fallocate(fd, 0, 0, 10*1024*1024*1024l);
}
```

takes only two seconds to complete on my computer and creates a 10GB file.

### LINKED ISSUE #916: Use isolate's quota option instead of fsize
See #913 for the discussion.

### LINKED ISSUE #1005: Use isolate's temporary directory rules
As previously discusse with @stefano-maggiolo in PR #992, I taught isolate to create temporary directories on request, and /tmp by default. It would be nice to use it in CMS.

### LINKED ISSUE #1242: Avoid bind-mounting temporary dirs within sandbox

prandla:
* Use /box instead of /tmp as the sandbox's home dir.
* Remove allow_writing_only; it's impossible to enforce with the new design.
* Remove "secure commands" mechanism, run all commands in sandbox. (I think it was mostly useful earlier to run unzip, by default with allow_writing_only it couldn't unzip properly. But none of the languages even run unzip anymore.)
* Simplify cleanup logic a bit, instead of keeping the sandbox dir, we archive the sandbox in case of failure. Also remove the keep_sandbox flag. (TODO: write the tool that unzips an archived sandbox for debugging. It can take the archive hash as an argument and load it from the file storage automatically.)
* Remove Job.keep_sandbox field, from what I can tell this was already unused.
* Remove sandbox.max_file_size option: It breaks dotnet, and it doesn't enforce any cap on the total size of /box.
* And finally, make CMS set isolate's --quota.

Creating automated tests to verify the quota behavior was nontrivial: setting up a quota-enabled filesystem is actually somewhat tricky (or at least poorly documented). Initially I tested locally with a tmpfs as the box_root; this required hacking isolate to use quotactl_fd instead of quotactl. (the former is a much nicer interface, but it's undocumented. I need to make a bug report to man-pages to get it documented :) )

~~Eventually I figured out that with an ext4 filesystem on a loop device, it's also possible to use the old quotactl, and thus the current release of isolate, when running tests.~~ We apparently need tmpfs quotas on github actions. New Isolate release uses quotactl_fd exclusively.

Also, since the quota setting is completely global, I had to set it to a somewhat high value (or well, I don't actually know how big is required, but since it's also used during compilation i set it to 64MB). And then I had to make the loop filesystem quite large also...

I also wrote some documentation about quotas in the sample config file (should this be moved to the documentation instead?). The conditions for how to configure the filesystem for quota support were found by pure trial and error :), but during it I discovered some combinations where isolate runs successfully with the --quota setting, but the quota is silently not enforced. This seems a bit scary, perhaps we should explicitly warn against this...

Closes #1242, closes #1005, closes #916, closes #309, maybe a few more :)


prandla:
well great. github actions is running kernel 6.11.0-1018-azure, which simply does not have quota support compiled in (neither built-in nor as a kernel module). (this wasn't easy to even verify because they also disabled /proc/config.gz.)

So, time to go spelunking for a kernel that supports it i guess???

PR #1534, opened 2025-08-27
Remove token for codecov

wil93:
In v5 it should not be required anymore for public repositories

See https://github.com/codecov/codecov-action?tab=readme-ov-file#v5-release


prandla:
okay, it's now at least uploading coverage for the pull request successfully, so i can only hope that it'll also work on the main branch...

then again, it was also doing that earlier (i.e. failing on main but successful in PRs), so i don't know if this actually fixes anything.

Issue #1480, opened 2025-07-26
Languages should be more configurable


prandla:
Inspired by me trying to get the test suite to pass on my arch linux, for which the following changes were necessary:
* I use rustup, so /usr/bin/rustc isn't useful on its own. I installed a separate toolchain via `RUSTUP_HOME=/opt/rustup rustup toolchain install stable`. This means I needed to change the rustc path to `/opt/rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/rustc` (and add `/opt` to the sandbox).
* On arch, /usr/bin/pypy is a symlink into /opt as well, so this also requires the `/opt` bind mount.
* Arch's GHC doesn't support static linking, so I needed to replace the `-static` flag to it with `-dynamic`.

I think fixing this satisfactorily is two-fold:
* first, languages ought to have more control over the parameters the sandbox. This would allow moving https://github.com/cms-dev/cms/blob/9375d44d8a10dd7f0b61aef56ddadc1f22965946/cms/grading/Sandbox.py#L975-L982 into the language-specific code. Apart from the mapped directories, languages could also configure things like the environment variables and open file limits (which are needed for dotnet).
* second, system administrators ought to have control over the parameters of a language. This would allow changing the executable path of compilers/interpreters if necessary, adding extra bind mounts, or perhaps allow arbitrary tweaking of compiler flags.

Regarding the first point, we could add methods like "configure_{compilation/evaluation}_sandbox" to each language, which take as argument the Sandbox which is about to be used. This would be the most flexible from our POV, but would effectively make the Sandbox class a part of the "public" API that's necessary for adding new languages. Though I'm not sure if we really want to make any stability guarantees about these extensions, so maybe that's not a big issue. The alternative is to have slightly more restricted configuration, e.g. via returning a dict which can contain some specific sandbox parameters.

For the second point, we could make each language have some set of "options" that are configured in cms.toml. I think by default this could include `compiler_path` or `interpreter_path`, `compilation_extra_dirs` (for additional directories to bind-mount for compilation), `evaluation_extra_dirs`, and maybe `compiler_extra_flags` or `interpreter_extra_flags`. Each language could also add its own options (e.g. static vs dynamic linking in haskell). In the config file, this might look like:
```toml
[languages."Rust"]
compiler_path = "/opt/rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/rustc"

[languages."Python 3 / PyPy"]
compilation_extra_dirs = ["/opt"]
[...]

Issue #1243, opened 2023-10-08
Add support for C# compiled by .NET SDK


gollux:
I tried adding support for C# compiled by .NET SDK 7.0. I have a very preliminary version running in the Czech national olympiad, but before I submit a pull request, I would like to discuss some problems I stumbled upon and agree on how they should be solved properly.

(1) Permission problems. The dotnet compiler creates a whole forest of intermediate directories, often with restrictive permissions. This hits the issue #1242. I worked around this by creating a temporary directory within `/box` and symlinking it as `obj` in the current directory.

(2) User lookup problem. The compiler expects that the current UID has an entry in `/etc/passwd`. With a typical installation of Isolate (as installed by CMS installation scripts), the sandbox users don't have their entries. As CMS bind-mounts `/etc` in compilation sandboxes, this can be worked around by adding all sandbox users to `/etc/passwd` of the host system. I wonder if this has a nicer solution...

(3) To use `dotnet build`, one must create a `*.cproj` project file first. I figured out how to do it, but not how to integrate it nicely with the CMS language modules: they allow generating compilation commands, but I would rather need to call a Python function which would generate the file and put it in the sandbox directory. I worked around it by calling a helper script, but then we have to install the helper script to a location which is available inside the sandbox (e.g., `/usr/local/bin`), which is different from where we install the CMS itself (in my case, it's a virtual environment).

(4) I need to modify some parameters of the sandbox: increase the maximum number of open files, pass some extra environment variables, possibly bind-mount more directories. I did it by adding these options to all languages, but it would be really nice to have them per-language. I handled environment variables in the helper script for the time being.

I would appreciate any ideas on how to solve these problems in an upstreamable way.

andreyv:
> (2) User lookup problem. The compiler expects that the current UID has an entry in `/etc/passwd`. [...] I wonder if this has a nicer solution

Does the compiler really read `/etc/passwd` directly, or does it use `getpwuid()`?

In the latter case it should be possible to write a glibc NSS plugin for isolate that would resolve isolate uids dynamically. systemd's `DynamicUser=` feature is resolved this way via `nss-systemd(8)`.

Alternatively, it may be possible to define the user records in `/etc/userdb` or `/run/userdb`. Then they would get resolved through `systemd-userdbd(8)`. But the sandboxed program needs to have the ability to call host D-Bus for this.

gollux:
> Does the compiler really read `/etc/passwd` directly, or does it use `getpwuid()`?

[...]

PR #1545, opened 2025-09-03
Allow setting sandbox parameters per-language, make sandbox more restrictive

prandla:
* Allow each language to modify the parameters of the sandbox for both compilation and evaluation.
* Don't put /etc in the sandbox by default, only include the parts of /etc that are necessary for specific languages. (except for pascal... isolate doesn't support bind-mounting individual files, and pascal uses /etc/fpc.cfg instead of a subdirectory, so we still need to mount all of /etc for it. Note that the kernel does support bind-mounting files - should this feature be added to isolate?)
* Don't use preserve_env=True for compilation sandboxes, instead set PATH (which seems to be the only environment variable that was actually necessary) to a reasonable value manually.

This implements one part of #1480 and is necessary for supporting dotnet (#1243).

I call configure_compilation_sandbox and configure_evaluation_sandbox as late as possible, to allow languages to override the general parameters if needed. None of them need that currently, but it might be useful in the future.

Note that TwoSteps doesn't actually call configure_evaluation_sandbox, but then again it never even called get_evaluation_commands in the first place...

### DIFF
--- a/cms/grading/Sandbox.py
+++ b/cms/grading/Sandbox.py
@@ -287,23 +287,22 @@ class Sandbox:
         # between sandboxes.
         self.dirs.append((None, "/dev/shm", "tmp"))
 
-        # Set common environment variables.
+        # Set common configuration that is relevant for multiple
+        # languages.
+
+        self.set_env["PATH"] = "/usr/local/bin:/usr/bin:/bin"
+
         # Specifically needed by Python, that searches the home for
         # packages.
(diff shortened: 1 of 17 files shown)


prandla:
Luca thought that exposing different things to different languages is probably not a good idea.

I suppose that makes sense, at least wrt add_mapped_directory. I managed to get dotnet working without this (though it's even uglier now), so I guess we can close this for now.

PR #1583, opened 2025-11-03
Remove PrintingService

### LINKED ISSUE #1544: Printing feature has bitrotted
printing doesn't work currently. We are using PyPDF2's PdfFileReader, PdfFileMerger, and getNumPages, which were deprecated in pypdf2 2.0 and removed in 3.0, and we are currently pinning to version 3.0.0.

It's not hard to fix (just renaming to non-deprecated variants seems to make it work), but it's been broken for who knows how long, so I wonder how many people even use this feature.

(Also, a2ps (the tool we use to convert text files to pdf) doesn't support unicode at all, and I can't find a replacement that does and is packaged in debian.)

Virv12:
Closes #1544.
As already discussed the PrintingService is unused and even broken.
This PR removes all (I hope) the relative code, this makes installation easier since it requires fewer dependencies.
I just left the translations marking them as deprecated (should we completely remove them?).

### DIFF
--- a/cms/locale/bg/LC_MESSAGES/cms.po
+++ b/cms/locale/bg/LC_MESSAGES/cms.po
@@ -27,7 +27,7 @@ msgid "Partially correct"
 msgstr "Частично вярно"
 
 msgid "#"
-msgstr ""
+msgstr "#"
 
 msgid "Outcome"
 msgstr "Резултат"
@@ -42,7 +42,7 @@ msgid "Memory used"
 msgstr "Използвана памет"
 
 msgid "Score details temporarily unavailable."
-msgstr ""
+msgstr "Някои подробности са временно недостъпни."
 
 #, python-format
 msgid "Subtask %(index)s"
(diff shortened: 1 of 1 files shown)


veluca93:
On translations, I think we can get rid of them.

prandla:
I thought we wanted to do a poll of existing users to see if they still use some of these "deprecated" features? printing was only broken in v1.5.0, it seems likely that some people might not have upgraded to it yet. and it's not hard to keep it working if someone does use it. (this feature is much less complicated than tokens for example...)

regarding the pr: i'll need to pull translation updates from weblate first to avoid merge conflicts.

i think cmscommon/tex.py is also useless now.

prandla:
also, i think your updater is buggy: you're mutating self.objs while iterating over it.

veluca93:
> I thought we wanted to do a poll of existing users to see if they still use some of these "deprecated" features? printing was only broken in v1.5.0, it seems likely that some people might not have upgraded to it yet. and it's not hard to keep it working if someone does use it. (this feature is much less complicated than tokens for example...)


I don't think we need a poll for PrintingService.

veluca93:
We have had the issue about printing being broken for two months. On that issue, the only messages present have been from people saying that we should remove it. I think that was long enough time for someone to come forward if need be (in fact, it was probably more, since the feature was broken before).

Given the cirumstances, I will merge this PR once it is technically ready. If someone complains once the feature is removed, we can restore it easily since we have access to git history.

docs/Installation.rst
prandla: texlive and a2ps should be removed from here too

prandla: texlive and a2ps should be removed from here too

PR #1633, opened 2026-02-12
Better handling of compilation memory limit exceeded errors

prandla:
* check for the sandbox "memory limit exceeded" result in `compilation_step`.
* change the sandbox so that memory limit is diagnosed more often: specifically, gcc has a driver process which catches the child's death and prints a nice error message. with the previous code this would result in a regular "nonzero return" verdict; now it's memory-limit (though the nice message on stderr is still visible).
* add a new user-visible compilation outcome for it.
* added a test for it. this test tries to use 1.1GB of memory for me, and fails in CMS in around 0.3 seconds, so it should quite reliably trigger the memory limit.
   * currently AWS actually doesn't distinguish the different kinds of compilation failure, so these tests all just assert that some compilation failure happened. this could be made more specific in the future.

### DIFF
--- a/cms/grading/Sandbox.py
+++ b/cms/grading/Sandbox.py
@@ -415,11 +415,15 @@ class Sandbox:
                 return self.EXIT_TIMEOUT_WALL
             else:
                 return self.EXIT_TIMEOUT
+        elif 'cg-oom-killed' in self.meta:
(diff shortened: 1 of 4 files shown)


veluca93 (approved):
LGTM, but I would want to double check with @gollux :-)

gollux (approved):

PR #1667, opened 2026-04-05
Add a new "interactive" task type.

veluca93:


cms/grading/tasktypes/Interactive.py
```diff
@@ -0,0 +1,306 @@
...
+    def _uses_stub(self) -> bool:
+        return self.compilation_type == self.COMPILATION_STUB
+
+    def compile(self, job, file_cacher):
+        """See TaskType.compile."""
+        if not check_executables_number(job, 0):
+            return
+
+        language = get_language(job.language)
+        if language is None:
```
prandla: this check is not present in any other tasktypes and it's not specific to this type; should be moved to some common place if it is something we want (but i can make a followup issue for that i guess)

cms/grading/tasktypes/Interactive.py
```diff
@@ -0,0 +1,306 @@
...
+        if not check_manager_present(job, self.CONTROLLER_FILENAME):
+            return
+        controller_digest = job.managers[self.CONTROLLER_FILENAME].digest
+
+        language = get_language(job.language)
+        controller_command = ["./%s" % self.CONTROLLER_FILENAME]
+        solution_commands = language.get_evaluation_commands(executable_filename)
+        # language.get_evaluation_commands returns a list of lists (multiple
+        # commands or alternatives). We take the first one.
+        solution_command = solution_commands[0]
```
prandla: This does not match the behavior of Communication:
```
# Assumes that the actual execution of the user solution is the
# last command in commands, and that the previous are "setup"
# that don't need tight control.
```
this behavior is not used by any builtin languages (except for java when USE_JAR = False, but that one is hardcoded to True), but it's necessary for the dotnet support that I want to upstream. (for dotnet, a separate unzip step before execution seems necessary; i couldn't find any cleaner solution.)

(well, the "don't need tight control" part of that comment is not necessary and probably not a good idea. but still, the other commands do need to be executed in the sandbox.)

(also, it'd probably be good to document this convention somewhere other than a comment inside Communication.py...)

cms/grading/tasktypes/Interactive.py
```diff
@@ -0,0 +1,306 @@
...
+                stdout=subprocess.PIPE,
+                stderr=subprocess.STDOUT,
+            )
+            stdout, _ = p.communicate(timeout=self.controller_wall_limit * 2)
+
+            KEEPER_ERROR_MESSAGE = N_("Internal error in interactive keeper")
+
+            if p.returncode != 0:
+                logger.error("Keeper failed with return code %d", p.returncode)
+                logger.error("Keeper output: %s", stdout.decode("utf-8"))
```
prandla: probably not very important but you could do `stdout.decode(errors='replace')`, seeing as this is already a failure path

cms/grading/tasktypes/Interactive.py
```diff
@@ -0,0 +1,306 @@
...
+                job.success = False
+                job.text = [KEEPER_ERROR_MESSAGE]
+                return
+
+            stdout_str = stdout.decode("utf-8")
```
prandla: and this .decode could be moved into the following try block

cms/grading/tasktypes/Interactive.py
```diff
@@ -0,0 +1,306 @@
...
+
+            p = subprocess.Popen(
+                [sys.executable, keeper_path, json.dumps(keeper_config)],
+                stdin=subprocess.PIPE,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.STDOUT,
+            )
+            stdout, _ = p.communicate(timeout=self.controller_wall_limit * 2)
+
+            KEEPER_ERROR_MESSAGE = N_("Internal error in interactive keeper")
```
prandla: this does not need to be marked as translatable; we set job.success to False when showing this so it'll never show up in CWS

(in fact I don't know how it shows up in AWS either...)

cms/grading/tasktypes/interactive_keeper.py
prandla, line 1: why is this file needed? why can't this logic live in the Interactive class?

i assume you do have a good reason for this, but it should be explained somewhere.

cms/grading/tasktypes/interactive_keeper.py
```diff
@@ -0,0 +1,262 @@
...
+    next_process_index = 0
+    solution_sandboxes = []
+    solution_procs = []
+
+    while True:
+        line = controller_proc.stdout.readline()
+        if not line:
+            break
+        line = line.decode("utf-8").strip()
+        if line == "START_SOLUTION":
```
prandla, line 130: any other lines in the controller stdout seem to be ignored; should they be explicitly rejected?

cms/grading/tasktypes/interactive_keeper.py
```diff
@@ -0,0 +1,262 @@
...
+    box_success_user = all(r[0] for r in user_results)
+    evaluation_success_user = all(r[1] for r in user_results)
+
+    valid_stats = [r[2] for r in user_results if r[2] is not None]
+
+    if valid_stats:
+        stats_user = reduce(merge_execution_stats, valid_stats)
+        if sum_memory_usage:
+            stats_user["execution_memory"] = sum(
+                s["execution_memory"] for s in valid_stats
```
prandla: merge_execution_stats takes care not to choke on None values here (some individual stats can be None even if the stats_dict itself is not None). i'd say it's probably better to use merge_execution_stats directly for this. the only problem is that it only has one boolean argument for sum_memory_usage/sum_wall_time. if you really need sum_memory_usage to be separate from sum_wall_time, you could add a new kwarg to merge_execution_stats for it.

cms/grading/Sandbox.py
```diff
@@ -655,13 +656,15 @@ def execute_without_std(
             return the Popen object from subprocess.
 
         """
-        popen = self._popen(
-            command,
-            stdin=subprocess.PIPE,
-            stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE,
-            close_fds=True,
+        stdin = self.stdin_file if isinstance(self.stdin_file, int) else subprocess.PIPE
```
prandla, line 660: if these fields can now be ints, their type hint should be updated

cms/grading/Sandbox.py
```diff
@@ -900,6 +904,8 @@ def build_box_options(self) -> list[str]:
             res += ["--wall-time=%g" % self.wallclock_timeout]
         if self.extra_timeout is not None:
             res += ["--extra-time=%g" % self.extra_timeout]
+        if not self.close_fds:
+            res += ["--inherit-fds", "--open-files=8192"]
```
prandla: why the limit of 8192 specifically? isolate allows setting `--open-files=0` for no limit

prandla (changes requested):

cms/grading/tasktypes/Interactive.py
veluca93: I don't think this should be here :-)

cms/grading/tasktypes/Interactive.py
veluca93: ugh, I can't say I like this convention :-) but we should not needlesly diverge from it.

cms/grading/tasktypes/Interactive.py
veluca93: make sense, not super sure how it shows up in AWS either...

cms/grading/tasktypes/interactive_keeper.py
veluca93: Added a comment.

cms/grading/tasktypes/interactive_keeper.py
veluca93: I removed the two separate knobs and made this into a single `concurrent` boolean.

cms/grading/tasktypes/interactive_keeper.py
```diff
@@ -0,0 +1,274 @@
...
+
+
+def get_controller_text(sandbox):
+    score = None
+    text = []
+    admin_text = None
+    with sandbox.get_file_text("stderr.txt") as f:
+        for line in f.readlines():
+            line = line.strip()
+            if line.startswith("SCORE: "):
```
prandla, line 53: i think there probably should be some more error-checking here aswell. (there should be errors for duplicated lines, and for unrecognized lines)

cms/grading/tasktypes/interactive_keeper.py
```diff
@@ -0,0 +1,274 @@
...
+                except ValueError:
+                    pass
+            elif line.startswith("USER_MESSAGE: "):
+                text = [line[len("USER_MESSAGE: ") :].strip()]
+            elif line.startswith("ADMIN_MESSAGE: "):
+                admin_text = line[len("ADMIN_MESSAGE: ") :].strip()
+    return score, text, admin_text
+
+
+class CopyAbsPathFileCacher:
```
prandla: this feels really ugly....

i think you could just not give the Sandbox any FileCacher. then you'd have to use create_file_from_string or plain create_file instead of create_file_from_storage, but it would still feel a lot less hacky to me.

cms/grading/tasktypes/Interactive.py
```diff
@@ -0,0 +1,292 @@
...
+                job.text = [KEEPER_ERROR_MESSAGE]
+                return
+
+            try:
+                stdout_str = stdout.decode("utf-8")
+                result = json.loads(stdout_str)
+                logger.info("Parsed keeper result: %s", result)
+            except (ValueError, json.JSONDecodeError) as e:
+                logger.error(
+                    "Failed to parse keeper output: %s. Output: %r", e, stdout_str
```
prandla: sorry, i didn't notice before that stdout_str is used here too - this needs to be stdout.decode(errors='replace') aswell

cms/grading/tasktypes/Interactive.py
```diff
@@ -0,0 +1,292 @@
...
+    create_sandbox,
+    delete_sandbox,
+    is_manager_for_compilation,
+)
+
+logger = logging.getLogger(__name__)
+
+
+# Dummy function to mark translatable string.
+def N_(message):
```
prandla: this is unused now

cms/grading/ParameterTypes.py
```diff
@@ -134,6 +134,28 @@ def parse_string(self, value):
         return value
 
 
+class ParameterTypeBool(ParameterType):
+    """Type for an integer parameter."""
+
+    TEMPLATE = GLOBAL_ENVIRONMENT.from_string("""
+<select name="{{ prefix ~ parameter.short_name }}">
```
prandla: a dropdown with the options "true" and "false"? have you ever heard of a checkbox? :)

also, the docstring just above seems to be a copy-paste error, should say "boolean parameter". (unless you're subtly alluding to the fact that bool is a subclass of int :p)

prandla (commented):
i think the implementation in general is fine now. interactive_keeper.py doesn't really spark joy for me, but i don't know how to improve it much. (aside from the one comment i left in it)

cms/grading/tasktypes/interactive_keeper.py
veluca93: Good catch :-)

cms/grading/ParameterTypes.py
veluca93: I fixed the docstring.

However, it's at least non-trivial to use a checkbox, since for whatever reason a checkbox does *not* send a value if the checkbox is unchecked, and the existing infra expects a value.

prandla: yeah, that's just how html checkboxes work, it is a bit annoying.

i'd still prefer if this was fixed. a checkbox is the semantically correct input for a boolean. you don't need to change existing infra much: just override parse_handler instead of parse_string. (i think you still need to provide a dummy parse_string to make it a non-abstract class though)

veluca93: That was easier than I expected :-)

cms/grading/tasktypes/interactive_keeper.py
veluca93: Got rid of this.

prandla (approved):

cms/grading/tasktypes/Interactive.py
```diff
@@ -0,0 +1,290 @@
...
+        _CONTROLLER_TIME_LIMIT,
+        _CONTROLLER_WALL_LIMIT,
+    ]
+
+    def __init__(self, parameters):
+        super().__init__(parameters)
+        self.process_limit = self.parameters[0]
+        self.compilation_type = self.parameters[1]
+        self.concurrent = self.parameters[2]
+        # Note: isolate wants the memory limit in *bytes*!
```
gollux: I doubt that. `--cg-mem` is given in KiB.

cms/grading/tasktypes/Interactive.py
```diff
@@ -0,0 +1,290 @@
...
+                )
+                job.success = False
+                job.text = [KEEPER_ERROR_MESSAGE]
+                return
+
+            try:
+                stdout_str = stdout.decode("utf-8")
+                result = json.loads(stdout_str)
+                logger.info("Parsed keeper result: %s", result)
+            except (ValueError, json.JSONDecodeError) as e:
```
gollux: Is this necessary? `JSONDecodeError` should be a subclass of `ValueError`.

cms/grading/tasktypes/interactive_keeper.py
```diff
@@ -0,0 +1,270 @@
...
+from cms.grading.steps.evaluation import (
+    evaluation_step_before_run,
+    evaluation_step_after_run,
+    human_evaluation_message,
+)
+from cms.grading.steps.stats import merge_execution_stats
+from cms.grading.steps import trusted_step
+
+# Note: we keep a separate interactive keeper (running in a separate
+# process) to avoid opening many file descriptors in the main worker.
```
gollux, line 37: Having to `exec` the keeper (and having it load lots of CMS libraries again) is not nice.  But I admit that it's much safer than all alternative solutions I've been able to come up with :) So probably OK.

cms/grading/tasktypes/Interactive.py
```diff
@@ -0,0 +1,290 @@
...
+                "delete_sandbox": not (job.keep_sandbox or job.archive_sandbox),
+            }
+
+            keeper_path = os.path.join(
+                os.path.dirname(__file__), "interactive_keeper.py"
+            )
+
+            p = subprocess.Popen(
+                [sys.executable, keeper_path, json.dumps(keeper_config)],
+                stdin=subprocess.PIPE,
```
gollux: Is stdin used for anything?

cms/grading/tasktypes/interactive_keeper.py
```diff
@@ -0,0 +1,270 @@
...
+        os.set_inheritable(c_to_u_w, True)
+        os.set_inheritable(u_to_c_r, True)
+        os.set_inheritable(u_to_c_w, True)
+        pipes.append({"c_to_u": (c_to_u_r, c_to_u_w), "u_to_c": (u_to_c_r, u_to_c_w)})
+
+    controller_sandbox = Sandbox(0, shard, name="controller", temp_dir=temp_dir)
+    for path in controller_files:
+        with controller_sandbox.create_file(path, executable=True) as f:
+            with open(os.path.join(temp_dir, path), "rb") as g:
+                f.write(g.read())
```
gollux: Do we want to load the whole (potentially huge) file in memory at once?

cms/grading/tasktypes/Interactive.py
```diff
@@ -0,0 +1,290 @@
...
+        {
+            COMPILATION_ALONE: "Submissions are self-sufficient",
+            COMPILATION_STUB: "Submissions are compiled with a stub",
+        },
+    )
+
+    _PROCESS_LIMIT = ParameterTypeInt(
+        "Process limit",
+        "process_limit",
+        "Maximum number of solution instances the controller can spawn",
```
gollux: We should explain that we consume resources proportionally to this limit.

gollux (changes requested):
I have several nitpicking comments, but overall I like it. However, there should be some documentation on how to use the task type.

cms/grading/tasktypes/Interactive.py
veluca93: Ah I was referring to the `Sandbox` class, not isolate directly. Updated the comment.

cms/grading/tasktypes/Interactive.py
veluca93: Doesn't look like it.

cms/grading/tasktypes/interactive_keeper.py
veluca93: Ideally `interactive_keeper` does not depend from a significant number of CMS libraries... but that seems tricky to do, so this is probably the best we can do for now.

cms/grading/tasktypes/interactive_keeper.py
veluca93: Replaced with `shutil.copyfileobj`

docs/Task types.rst
```diff
@@ -103,6 +103,43 @@ Communication supports user tests. In addition to the input file, contestant mus
...
+- ``USER_MESSAGE: <message>``: a message for the contestant.
+- ``ADMIN_MESSAGE: <message>``: an optional message for the administrators.
+
+Special messages like ``translate:success``, ``translate:wrong`` and ``translate:partial`` are supported for the user message.
+
+Interactive has six parameters:
+
+- whether the contestant's source is compiled on its own (``alone``) or with an admin-provided stub called :file:`stub.{ext}` (``stub``);
+- the maximum number of solution instances the controller can spawn (``process_limit``);
+- whether the spawned solutions are assumed to run concurrently (``concurrent``);
```
gollux, line 133: Explain this, please.

docs/Task types.rst
```diff
@@ -103,6 +103,43 @@ Communication supports user tests. In addition to the input file, contestant mus
...
+- ``SCORE: <score>``: the outcome of the testcase, as a floating point number between 0.0 and 1.0.
+- ``USER_MESSAGE: <message>``: a message for the contestant.
+- ``ADMIN_MESSAGE: <message>``: an optional message for the administrators.
+
+Special messages like ``translate:success``, ``translate:wrong`` and ``translate:partial`` are supported for the user message.
+
+Interactive has six parameters:
+
+- whether the contestant's source is compiled on its own (``alone``) or with an admin-provided stub called :file:`stub.{ext}` (``stub``);
+- the maximum number of solution instances the controller can spawn (``process_limit``);
```
gollux, line 132: Please mention that this consumes resources.

gollux (approved):

docs/Task types.rst
prandla: I'd suggest something like
"This controls how the total wall clock time and total memory usage of the submission is computed. If concurrent, then wall clock times are max'd and memory usages added; if not concurrent, then wall clock times are added and memory usages max'd. This only affects the display of statistics, not the actual verdict (for the verdict, each spawned solution gets the full time/memory limit). Also note that CPU times are always added."

docs/Task types.rst
```diff
@@ -103,6 +103,43 @@ Communication supports user tests. In addition to the input file, contestant mus
...
+- whether the spawned solutions are assumed to run concurrently (``concurrent``);
+- the maximum memory (in MB) that the controller can use;
+- the maximum CPU time (in seconds) that the controller can use;
+- the maximum wall time (in seconds) that the controller can use.
+
+Admins can provide a manager called :file:`stub.{ext}` for each allowed language. When using a stub, the controller usually sends the file descriptors to the solution as command-line arguments, which the stub then uses to set up communication.
```
prandla: what does this line mean? from what i can tell, the solution isn't given any command-line args, the fifos are only connected to its stdin/stdout.

prandla:
Also, I wonder if we should allow setting the solutions' wall time limit separately from the cpu time limit. If one has a task with multiple concurrent solutions of which only one is actively computing something at any given time, then it seems quite likely that the default wall limit of 2*cpu_limit+1 is not enough.

docs/Task types.rst
veluca93: I don't think that line made sense... Removed it.

veluca93:
> Also, I wonder if we should allow setting the solutions' wall time limit separately from the cpu time limit. If one has a task with multiple concurrent solutions of which only one is actively computing something at any given time, then it seems quite likely that the default wall limit of 2*cpu_limit+1 is not enough.

Isn't that also the case for other task types, i.e. communication?

prandla:
> Isn't that also the case for other task types, i.e. communication?

I suppose so, but then this could be another way for Interactive to improve upon Communication :)

veluca93:
> Also, I wonder if we should allow setting the solutions' wall time limit separately from the cpu time limit. If one has a task with multiple concurrent solutions of which only one is actively computing something at any given time, then it seems quite likely that the default wall limit of 2*cpu_limit+1 is not enough.

Changed to use the same limit as the controller wall time limit!

cms/grading/tasktypes/interactive_keeper.py
```diff
@@ -0,0 +1,272 @@
...
+                        shutil.copyfileobj(g, f)
+
+            # Matches Communication's handling of multi-command executions.
+            if len(solution_commands) > 1:
+                trusted_step(sandbox_sol, solution_commands[:-1])
+            sol_proc = evaluation_step_before_run(
+                sandbox_sol,
+                solution_commands[-1],
+                time_limit=solution_time_limit,
+                wall_limit=controller_wall_limit,
```
prandla, line 153: i'd suggest adding a comment here (or somewhere nearby) explaining why we set the wall limit to this.

something like "the wall-clock limit mostly exists to eventually kill stuck solutions. there's no point setting it higher than the controller's limit (as then the controller would die before any of the solutions), and there's little to be gained from setting it any lower. to reduce the amount of unnecessary configurability, set it to the same limit as the controller."

docs/Task types.rst
```diff
@@ -103,6 +103,47 @@ Communication supports user tests. In addition to the input file, contestant mus
...
+- whether the contestant's source is compiled on its own (``alone``) or with an admin-provided stub called :file:`stub.{ext}` (``stub``);
+- the maximum number of solution instances the controller can spawn (``process_limit``);
+- whether the spawned solutions are assumed to run concurrently (``concurrent``);
+- the maximum memory (in MB) that the controller can use;
+- the maximum CPU time (in seconds) that the controller can use;
+- the maximum wall time (in seconds) that the controller can use.
+
+The ``concurrent`` parameter controls how the total wall clock time and total memory usage of the submission is computed. If concurrent, then wall clock times are max'd and memory usages added; if not concurrent, then wall clock times are added and memory usages max'd. This only affects the display of statistics, not the actual verdict (for the verdict, each spawned solution gets the full time/memory limit). Also note that CPU times are always added.
+
+Note that in the current implementation each *potential* process spawned by the controller consumes two file descriptors. Moreover, resource usage is proportional to the number of actually-spawned solutions.
```
prandla: i'd rather phrase this as more explicitly about `process_limit`.

"Note that in the current implementation, all file descriptors for communication are preallocated before running the controller, which means that the controller always uses (at least) `2*process_limit` file descriptors. Thus it is not a good idea to set `process_limit` higher than actually necessary."

docs/Task types.rst
```diff
@@ -103,6 +103,47 @@ Communication supports user tests. In addition to the input file, contestant mus
...
+
+Admins can provide a manager called :file:`stub.{ext}` for each allowed language.
```
prandla: this paragraph feels a little awkward being only one sentence like this; perhaps also add something like "The stub serves the same purpose as in the Communication task type."

PR #1677, opened 2026-04-10
Allow configuring subtasks to always show testcases.

veluca93:
Also defines a new (optional) format for parameters for ScoreTypeGroup.

### DIFF
--- a/cms/grading/scoretypes/abc.py
+++ b/cms/grading/scoretypes/abc.py
@@ -32,6 +32,7 @@ task.
 
 import logging
 import re
+from typing import TypedDict, NotRequired
 from abc import ABCMeta, abstractmethod
 
 from cms import FEEDBACK_LEVEL_RESTRICTED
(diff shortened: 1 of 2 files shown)


cms/grading/scoretypes/abc.py
```diff
@@ -192,26 +193,38 @@ class ScoreTypeAlone(ScoreType):
     pass
 
 
+class ScoreTypeGroupParametersDict(TypedDict):
+    max_score: float
+    testcases: int | str | list[str]
+    threshold: NotRequired[float]
```
prandla, line 199: i'm a little unhappy about this scoretype-specific field being here, but well, the typing of the parameters is a huge mess anyways and this is hardly the worst part of it.

i don't think it's possible to fix it without refactoring how scoretypes work in general. python doesn't have associated types, so i don't think there is even a good way to express "this scoretype uses this subtype of the parameters dict for its methods". so we can just continue ignoring the type errors.

cms/grading/scoretypes/abc.py
```diff
@@ -333,6 +346,24 @@ class ScoreTypeGroup(ScoreTypeAlone):
 </div>
 {% endfor %}"""
 
+    def get_max_score(self, group_parameter: ScoreTypeGroupParameters) -> float:
+        if isinstance(group_parameter, tuple) or isinstance(group_parameter, list):
```
prandla, line 350: can these ever legitemately be tuples? these parameters should be loaded from json, which doesn't return tuples. the type hint says tuple only because there is no way to type-hint a list whose elements have different types 🙃 (actually, that should probably be added to the comment near the definition of ScoreTypeGroupParameters)

if you assumed doing an isinstance(tuple) here would get the type checker to behave on the next line, then unfortunately that doesn't appear to be the case, as pyright at least assumes that the parameter is some kind of magical intersection of list and ScoreTypeGroupParametersDict.

prandla (approved):
Some philosophical comments regarding type hints.

The implementation itself looks fine.

prandla:
I also briefly considered whether we should drop support for the list-based format entirely (at least within cms.grading.scoretypes) and convert the lists to dicts in the importers / a db updater, but it'd probably end up being more effort in total. that would probably also be more disruptive to forks with custom scoretypes.

cms/grading/scoretypes/abc.py
veluca93: It helps with *my* typechecker, at least (presumably `ty`), so I'd leave it here :-)

cms/grading/scoretypes/abc.py
veluca93: Yeah, I am not hugely happy with it myself,, but I figured it was better than other options....

veluca93:
> I also briefly considered whether we should drop support for the list-based format entirely (at least within cms.grading.scoretypes) and convert the lists to dicts in the importers / a db updater, but it'd probably end up being more effort in total. that would probably also be more disruptive to forks with custom scoretypes.

Agreed, I really don't like the "list of arbitrary parameters" approach, but I figure maintaining backwards compatibility is important here.

PR #1685, opened 2026-04-22
Bugfix

### LINKED ISSUE #1684: redirect after login has wrong url
* make a submission
* wait until you get logged out (or probably logout from another tab)
* refresh, log back in
* get redirected to `http://localhost:8888/tasks/batch/submissions%3Fsubmission_id%3DJsdiRWBytP6rTwiI56HPfsQTvhhGUmrxupCA7nWyXk4.` and get a 404 (note the extraneous urlencoding)

pxsit:
Improves the handling of the `next` page redirection parameter in both the admin and contest handlers.
Fix https://github.com/cms-dev/cms/issues/1684


cms/server/admin/handlers/main.py
```diff
@@ -48,10 +49,22 @@ def post(self):
...
             error_args["next"] = next_page
-            if next_page != "/":
-                next_page = self.url(*next_page.strip("/").split("/"))
+            split = urlsplit(next_page)
+            path = split.path or "/"
+            if split.scheme or split.netloc or not path.startswith("/"):
+                next_page = self.url()
+            elif path != "/":
+                path_segments = path.strip("/").split("/")
+                if any(segment in ("", ".", "..") for segment in path_segments):
```
prandla: it looks like you're trying to prevent path traversal attacks here, but i don't think they are relevant here. if an attacker can cause you to navigate to an arbitrary login url, they can already cause you to navigate to any url directly. (and since the redirect after the login is always a GET request, this should be harmless anyways.)

usually i'd say this is fine as a defense-in-depth thing, but some parts of cms do use relative paths so i'm afraid this might break something else.

prandla (changes requested):
there's a bunch of code that's duplicated between the CWS and AWS parts now. i'd suggest moving them into a helper function in cms/server/util.py.

prandla (approved):

PR #1690, opened 2026-05-01
Round score at subtask level

akmohtashami:
This change updates scoring to match the updated IOI rule: each subtask score is rounded to the task’s score_precision, and the final task score is computed as the sum of those rounded subtask scores (with a final rounding pass to avoid float artifacts).

It also allows us to avoid passing unrounded score around. I am keeping some rounding in ranking in case of float issues.


akmohtashami:
FYI, @veluca93

veluca93 (approved):

Issue #1483, opened 2025-07-27
Mapping extensions to languages is broken


gollux:
In PR #1464, I discovered that `cmsAddSubmission` without `--lang` defaults to the _first_ language on the list which matches the filename extension, regardless of what is enabled in the contest. The test suite did not originally catch this, because it used C which has only one version enabled.

All callers of `cms.grading.languagemanager.filename_to_language` seem to be affected, in particular `cmsMake` and `cmstestsuite/web/CWSRequests.py`.

`cmsAddSubmission` is already fixed by PR #1464, but the other places need attention, too.

@veluca93 mentioned that `cmsMake` can be probably removed.

PR #1691, opened 2026-05-02
Add a script to check solutions.

veluca93:


cmscontrib/SolutionChecker.py
```diff
@@ -0,0 +1,178 @@
...
+        "--checks-json", "-c", required=True, help="Path to solution_checks.json"
+    )
+    parser.add_argument(
+        "--url",
+        "-u",
+        required=True,
+        help="CMS contest URL (e.g. http://localhost:8888/contest)",
+    )
+    parser.add_argument("--task", "-t", required=True, help="Task name")
+    parser.add_argument("--username", "-U", required=True, help="CMS username")
```
Virv12: Why is the username required?

Virv12 (approved):

prandla:
Does this not run into #1483 ?

on a somewhat related note, filename_to_language is the only part of cms that is actually imported. this script might be more useful if this dependency were removed; then using it would not require a (local) cms installation at all.

veluca93:
> Does this not run into #1483 ?
> 
> on a somewhat related note, filename_to_language is the only part of cms that is actually imported. this script might be more useful if this dependency were removed; then using it would not require a (local) cms installation at all.

Probably yes.

veluca93:
> > Does this not run into #1483 ?
> > on a somewhat related note, filename_to_language is the only part of cms that is actually imported. this script might be more useful if this dependency were removed; then using it would not require a (local) cms installation at all.
> 
> Probably yes.

Now it does not anymore, and checks for slow testcases. PTAL.

cmscontrib/SolutionChecker.py
```diff
@@ -0,0 +1,247 @@
...
+            failed = True
+            error = "Evaluation failed."
+
+        if not failed:
+            logger.info("%20s: check successful", sol_name)
+        else:
+            has_failures = True
+            logger.error("%20s: %s", sol_name, error)
+
+        if checker.has_slow_testcases(args.task, sub_id, time_limit):
```
Virv12: We should check for warnings only if there are no errors

cmscontrib/SolutionChecker.py
```diff
@@ -0,0 +1,247 @@
...
+    def poll_status(self, task_name: str, filename: str, submission_id: str) -> Dict[str, Any]:
+        status_url = f"{self.base_url}/tasks/{task_name}/submissions/{submission_id}"
+        while True:
+            response = self.session.get(status_url, headers=self.auth_header)
+            response.raise_for_status()
+            data = response.json()
+            # status 5 is SCORED, 2 is COMPILATION_FAILED
+            if data.get("status") in [2, 5]:
+                return data
+            logger.info("Waiting for evaluation of %s...", filename)
```
Virv12: This is too verbose

PR #1695, opened 2026-05-13
Fix controller hang when submission exits early

simonlindholm:
Having the `u_to_c_w` fd inherit to the controller means that when the submission exits, the controller will never read EOF, because it still itself has the write end of the pipe open. Easy fix is just to make that not inherit: nothing actually needs the inheritance, in fact it's rather a bit of a security issue that each spawned submission get access to extra fd's.

Small additional related fixes:
- if the interactive keeper hits walltime limit (claude pointed out some obscure circumstances in which this can happen, e.g. if the submission causes lots of isolate warning spam which clogs an stderr pipe which is never read from), we currently crash and leave the child un-waited. Better to generate a proper error.
- the change to make `c_to_u_r` not inherit means that now if the submission exits, controller writing will cause SIGPIPE unless the controller ignores that signal. We don't particularly want that to happen (it makes it impossible to detect which process exited first, which we really ought to do to give proper verdicts); better to leave the pipe unclosed.

### DIFF
--- a/cms/grading/tasktypes/interactive_keeper.py
+++ b/cms/grading/tasktypes/interactive_keeper.py
@@ -87,10 +87,8 @@ def main():
     for i in range(process_limit):
         c_to_u_r, c_to_u_w = os.pipe()
         u_to_c_r, u_to_c_w = os.pipe()
-        os.set_inheritable(c_to_u_r, True)
         os.set_inheritable(c_to_u_w, True)
         os.set_inheritable(u_to_c_r, True)
-        os.set_inheritable(u_to_c_w, True)
         pipes.append({"c_to_u": (c_to_u_r, c_to_u_w), "u_to_c": (u_to_c_r, u_to_c_w)})
 
     controller_sandbox = Sandbox(0, shard, name="controller", temp_dir=temp_dir)
@@ -165,7 +163,8 @@ def main():
                 wait=False,
             )
 
-            os.close(p["c_to_u"][0])
+            # We do not close p["c_to_u"][0] -- it only risks crashing the
+            # controller with SIGPIPE if the submission exits early.
             os.close(p["u_to_c"][1])
 
             try:
--- a/cms/grading/tasktypes/interactive_keeper.py
+++ b/cms/grading/tasktypes/interactive_keeper.py
@@ -163,8 +163,7 @@ def main():
                 wait=False,
             )
 
-            # We do not close p["c_to_u"][0] -- it only risks crashing the
-            # controller with SIGPIPE if the submission exits early.
+            os.close(p["c_to_u"][0])
             os.close(p["u_to_c"][1])
 
             try:
--- a/cms/grading/tasktypes/Interactive.py
+++ b/cms/grading/tasktypes/Interactive.py
@@ -252,7 +252,11 @@ class Interactive(TaskType):
                 stdin=subprocess.DEVNULL,
                 stdout=subprocess.PIPE,
             )
-            stdout, _ = p.communicate(timeout=self.controller_wall_limit * 2)
+            try:
+                stdout, _ = p.communicate(timeout=self.controller_wall_limit * 2)
+            except subprocess.TimeoutExpired:
+                p.kill()
+                stdout, _ = p.communicate()
 
             KEEPER_ERROR_MESSAGE = "Internal error in interactive keeper"
 


cms/grading/tasktypes/interactive_keeper.py
```diff
@@ -165,7 +163,8 @@ def main():
                 wait=False,
             )
 
-            os.close(p["c_to_u"][0])
+            # We do not close p["c_to_u"][0] -- it only risks crashing the
+            # controller with SIGPIPE if the submission exits early.
```
prandla: i thought "manager gets sigpipe" is an issue that can happen anyways and that properly written managers always need to signal(SIGPIPE, SIG_IGN) anyways? (at least, looking at managers from old communication tasks, most of them do that.)

i admit this isn't very well documented, but imo leaving file descriptors open in two processes is not the right way to solve this.

i suppose this is at least not a resource leak, given that interactive_keeper should also exit shortly after the solution.

simonlindholm: SIGPIPE cannot happen if we leave the pipe open, so this fix should remove the need for `signal(SIGPIPE, SIG_IGN)`. Is there any downside to leaving the pipe unclosed? IMO pipe closures are a big footgun which are worth eliminating on the judge side. Apart from the need for `signal(SIGPIPE, SIG_IGN)` which is very easy to forget, it's also easy for a judge to accidentally introduce excessive (well-intentioned) error checking, e.g.:
```c
void write_message(const char* format, ...) {
    va_list args;
    va_start(args, format);
    if (vfprintf(fifo_out, format, args) <= 0) wrong_answer("Pipe to contestant closed unexpectedly.");
    if (fflush(fifo_out) == EOF) wrong_answer("Pipe to contestant closed unexpectedly.");
}
```

In a case where a submission prints a query and then exits without reading the response, this results in a race condition where depending on timing the submission may either get accepted or wrong answer. I'd argue strongly that the correct behavior here is to ignore errors, so such a submission would always get accepted. This is similar to how it's acceptable for non-interactive submission to not read the entire input.

(It would be good to also fix this for regular interactive problems, not of the Interactive task type, and it would also be good to properly look at the order in which manager/controller vs submission exits with various verdicts. See also discussion in https://github.com/Kattis/problemtools/pull/77. I don't think I have the capacity to work on this, however.)

veluca93 (approved):

gollux:
Maybe I am missing something, but not closing the pipe can potentially make the problem much worse. If the solution sends multiple queries and then it exits, the controller will try to print answers for all the queries, filling up the pipe buffer and blocking indefinitely.

I would argue that it is much better to ignore both `SIGPIPE` and write errors inside the controller.

simonlindholm:
That issue can occur regardless. A solution which sends a large number of queries without reading responses will cause the checker->solution pipe to fill up, which causes the checker to stop processing queries, and then after a while the solution->checker pipe fills up too and we get a deadlock. In Kattis, this has been mitigated somewhat by increasing pipe buffer size to the Linux maximum of 1 MiB, and it is eventually planned to be fixed for real by introducing a proxy process (see https://github.com/Kattis/problemtools/issues/113; additional upsides of that solution is being able to log the interaction and show it for sample testcases, and being able to impose communication traffic limits to eliminate judge errors from resource exhaustion caused by unbounded-size `cin >> str;` in checkers; downsides are complexity and increased judging walltime).

I suppose a small difference compared to that scenario is that with the pipe left open the checker could be the one left to take the blame, if the solution exits early and the judging system treats checker walltime limit timeouts as judge errors.

Anyway, this not being a hill I want to die on, I just pushed a commit that closes the pipe.

prandla:
i can think of a way to avoid the deadlock, though it requires a bit of extra complexity on the manager side.

* set the manager->solution pipe as O_NONBLOCK.
* when writing to the solution, attempt a write(); if it fails with EAGAIN, stick the data into a queue.
* when reading, if there's any queued data then instead of reading directly, do a poll() on both pipes; if the writing pipe has space now, write some data from the queue into it before proceeding with the read.

while this is somewhat complicated, i think it can be abstracted away into a header-only library a la testlib.h.

i think this should have minimal performance impact if the solution is well-behaved and reads its end of the pipe normally (in which case the write() should rarely fail at all, at least when assuming that the pipe buffer is large enough for one query's response). in any case, it avoids the overhead of another proxy process and all the associated context switches.

simonlindholm:
Sure, that also works. I'm not the biggest fan of putting logic like this into the manager itself, admittedly, it means that old problems need be updated in order to behave properly, and it's hard to ensure task authors use the latest testlib instead of copy-pasting from an old problem.

gollux:
I'm fine with the current version of the PR.

Regarding deadlocks: I do not think that reporting deadlocks in communication is worth the complexity (and time overhead) it requires. Just let the evaluation hit the wall time limit in the rare cases when this happens.

PR #1706, opened 2026-08-15
implement features needed for caching in ranking

akmohtashami:


gollux:
I would really appreciate some description of the changes. Just saying "features needed for caching" for such a huge patch does not make it reviewable.

PR #1710, opened 2026-08-19
Remove cmsMake.

veluca93:
It is not used and broken.


prandla (approved):
cmscommon/terminal.py `directions` and `move_cursor` are unused now, but i guess that doesn't matter really (the rest of terminal.py is still used by the logging stuff)

PR #1711, opened 2026-08-19
Rename stub.%l to grader.%l in Communication tasks.

### LINKED ISSUE #1705: Rename stub.%l to grader.%l in communication tasks
(see title)

veluca93:
Fixes #1705.


docs/Task types.rst
```diff
@@ -161,7 +161,7 @@ TwoSteps has one parameter, similar to Batch's third, that specifies whether to
 
 TwoSteps supports user tests; contestants must provide the manager in addition to the input and their sources.
 
-**How to migrate from TwoSteps to Communication.** Any TwoSteps task can be implemented as a Communication task with two processes. The functionalities in the stub should be migrated to Communication's manager, which also must enforce any restriction in the computed data.
+**How to migrate from TwoSteps to Communication.** Any TwoSteps task can be implemented as a Communication task with two processes. The functionalities in the grader should be migrated to Communication's manager, which also must enforce any restriction in the computed data.
```
prandla: the "stub" here is referring to twosteps's thing, not communication's. reading the previous paragraph though, it looks like the thing is actually called "manager" instead, so this should probably be changed to that.

docs/Task types.rst
prandla, line 1: the documentation of Interactive says "The stub serves the same purpose as in the Communication task type.", but this is not true any more (and might be confusing now that the rest of the file doesn't mention stubs at all).

personally, i think we maybe should add a footnote somewhere that "the thing called "grader" is not really grading anything; it's called that for historical reasons; a more accurate name would be "stub"". but i might be a bit biased :)

cmscontrib/updaters/update_from_1.5.sql
```diff
@@ -105,4 +105,31 @@ ALTER TABLE contests DROP COLUMN analysis_stop;
 -- https://github.com/cms-dev/cms/pull/1672
 ALTER TABLE contests DROP COLUMN per_user_time;
 
+-- Rename Communication task type compilation parameter from 'stub' to 'grader'
```
prandla: should have the PR number in a comment too

prandla (approved):
Interactive is now the only task type to still call it "stub". presumably we should do the same thing there for consistency?

docs/Task types.rst
veluca93: Now it says grader for Interactive too.

veluca93:
> Interactive is now the only task type to still call it "stub". presumably we should do the same thing there for consistency?

Completely forgot about it :-) done in a second commit (still in the same PR and modifying the same updater)

PR #1712, opened 2026-08-22
Add JSON APIs for submission details and use them in SolutionChecker.py

veluca93:


cmscontrib/SolutionChecker.py
```diff
@@ -34,7 +44,7 @@
 import re
 import sys
 import time
-from typing import Optional, Dict, Any
+from typing import Optional, Dict, Any, List
```
prandla: we depend on python 3.11, so Optional, Dict and List are unnecessary, you can use `| None`, `dict` and `list` respectively instead

cmscontrib/SolutionChecker.py
```diff
@@ -52,43 +62,103 @@ class RedAlertFormatter(logging.Formatter):
...
+            logger.info("Successfully logged in with admin token as %s.", self.username)
             return
 
+        if self.username is None and self.password is None:
+            logger.info("No credentials provided, assuming IP autologin.")
+            return
+
+        if self.username is None and self.password is not None:
+            raise ValueError(
+                "Password provided without username. Please specify " "--username."
```
prandla: what's up with the `" "`?

cmscontrib/SolutionChecker.py
```diff
@@ -135,40 +207,346 @@ def get_time_limit(self, task_name: str) -> float:
...
+                    f"{statuses_list}"
+                )
+
+        else:
+            return f"subtask {st_idx}: unknown check type '{check}'"
+
+        return None
+
+    def check_subtasks(
+        self, details: Optional[List[Dict[str, Any]]], checks: list
```
prandla: this could be typed as `checks: list[str | None]`

docs/API.rst
```diff
@@ -102,6 +102,47 @@ Additional details on the submission's results can be retrieved by making an
...
+If the task uses a non-subtask score type such as ``Sum``, ``details`` will
+contain a list of testcase objects directly.
+
+Unfiltered submission details (JSON)
+====================================
+
+An authenticated ``GET`` request to
+``/api/{taskname}/submissions/{id}/full_details`` will return a JSON object with
+the complete, unfiltered submission details (including all testcase execution
+times and memory usage).
```
prandla: i think we should have at least one example here of what the json looks like with the times and memory included

cms/server/contest/handlers/api.py
```diff
@@ -182,4 +208,79 @@ def get(self, task_name: str):
...
+            is_analysis_mode = self.r_params["actual_phase"] == 3
+            if submission.tokened() or is_analysis_mode:
+                raw_details = sr.score_details
+            else:
+                raw_details = sr.public_score_details
+
+            if is_analysis_mode:
+                feedback_level = FEEDBACK_LEVEL_FULL
+            else:
+                feedback_level = task.feedback_level
```
prandla, line 248: this logic is duplicated with `tasksubmission.py`, maybe it should be factored out somewhere (to ensure they stay in sync mainly)

cmscontrib/SolutionChecker.py
veluca93: For some reason when `black` reformats a multi-line string onto a single line it doesn't join the strings...

cms/server/contest/handlers/api.py
veluca93: I added a commit on top that refactors that logic, PTAL - I am not fully convinced, but if you like it I'll fold it into the relevant commit :-)

prandla: yeah that's not very clean.. feel free to revert it

prandla (approved):

cms/server/contest/handlers/api.py
veluca93: Did so.

PR #1716, opened 2026-09-15
Require auth for seeing/modifying teams.

veluca93:


prandla (approved):

prandla:
(minor comment: you cannot actually see teams without auth, as the jinja template throws a 500 error trying to access the nonexistent admin. but you can still blindly modify them)

PR #1717, opened 2026-09-15
Do not allow disabled logged-in administrators to make RPC calls via AWS.

veluca93:


prandla (approved):
i would clarify the title as "do not allow disabled administrators to make RPC calls via AWS". they already cannot use the rest of AWS.

PR #1720, opened 2026-09-23
Carmel/generator min timeout

Carmel-Ab1:
The existing time limit for input generators is about the original task's. This commit enforces a 30s minimum for generator's execution - which may be necessary for heavy IO or to the creation of strong testcases.


prandla:
github's amazing ux strikes again... i think you meant to target this at ioi-isr/cms

